PR #14, opened 2022-12-20
Support for journal sync flush

hkadayam:
This journal sync flush feature is required to integrate with HomeReplication to ensure faster flush from NuRAFT (as a sync call)

### DIFF
--- a/src/lib/logstore/log_dev.cpp
+++ b/src/lib/logstore/log_dev.cpp
@@ -173,8 +173,8 @@ void LogDev::do_load(const off_t device_cursor) {
         const auto flush_ld_key =
             logdev_key{header->start_idx() + header->nrecords() - 1, group_dev_offset + header->total_size()};
         while (i < header->nrecords()) {
-            const auto* const rec{header->nth_record(i)};
-            const uint32_t data_offset{(rec->offset + (rec->get_inlined() ? 0 : header->oob_data_offset))};
+            const auto* rec = header->nth_record(i);
+            const uint32_t data_offset = (rec->offset + (rec->get_inlined() ? 0 : header->oob_data_offset));
 
             // Do a callback on the found log entry
             sisl::byte_view b = buf;
@@ -266,7 +266,6 @@ log_buffer LogDev::read(const logdev_key& key, serialized_log_record& return_rec
     } else {
         // Round them data offset to dma boundary in-order to make sure pread on direct io succeed. We need to skip
         // the rounded portion while copying to user buffer
-        // TO DO: Might need to differentiate based on data or fast type
         auto const rounded_data_offset = sisl::round_down(data_offset, m_vdev->align_size());
         auto const rounded_size = sisl::round_up(b.size() + data_offset - rounded_data_offset, m_vdev->align_size());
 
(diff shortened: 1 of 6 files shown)


src/lib/logstore/log_store.cpp
```diff
@@ -375,6 +380,35 @@ logstore_seq_num_t HomeLogStore::get_contiguous_completed_seq_num(logstore_seq_n
...
+    {
+        std::unique_lock lk(m_sync_flush_mtx);
+        while (true) {
+            // Step 1: Mark the waiter lsn to the seqnum we wanted to wait for. The completion of every lsn checks
+            // for this and if this lsn is completed, will make a callback which signals the cv.
+            m_sync_flush_waiter_lsn.store(upto_seq_num);
+
+            // Step 2: After marking this lsn, we again do a check, to avoid a race where completion checked for no lsn
+            // and the lsn is stored in step 1 above.
+            if (m_records.completed_upto() >= upto_seq_num) { break; }
```
yamingk, line 398: if the while is for the purpose of this `break`, maybe we can `return` out of the function here.

src/lib/logstore/log_dev.cpp
```diff
@@ -173,8 +173,8 @@ void LogDev::do_load(const off_t device_cursor) {
         const auto flush_ld_key =
             logdev_key{header->start_idx() + header->nrecords() - 1, group_dev_offset + header->total_size()};
         while (i < header->nrecords()) {
-            const auto* const rec{header->nth_record(i)};
-            const uint32_t data_offset{(rec->offset + (rec->get_inlined() ? 0 : header->oob_data_offset))};
+            const auto* rec = header->nth_record(i);
```
szmyd, line 176: Why are we converting all these initializer list based constructors to assignment style expressions? The former is the preferred style I believe.

yamingk: I believe it is for readability purpose (there is one discussion about this we had in one of the homestore meeting). `const auto` should be as good to cover the benefits we get from initializer list.

hkadayam: We are not replacing initialized based constructors from {} to =, but we are replacing initialization of POD types to assignment as it doesn't provide any meaningful value and take a huge hit in readability in my opinion. It also doesn't clearly differentiate with the cases where `{}` is meaningful, for example in `std::unique_lock l{mtx}` which is a classic case where it is helpful.

src/lib/logstore/log_store.cpp
yamingk: If I understand this race correctly, the previous flush is still progressing and a new sync flush arrives with some upto_seq_num smaller than the max seq number that is pending on its completion. 
If so, should we introduce a new `m_prev_flush_upto_seq_num` so that we don't need this check and replace line: 387 with `if (m_prev_flush_upto_seq_num >= upto_seq_num)`? 

Just a minor suggestion and if you want to keep it this way it is also good to me.

src/lib/logstore/log_store.cpp
```diff
@@ -375,6 +380,35 @@ logstore_seq_num_t HomeLogStore::get_contiguous_completed_seq_num(logstore_seq_n
...
 
+void HomeLogStore::flush_sync(logstore_seq_num_t upto_seq_num) {
+    if (upto_seq_num == invalid_lsn()) { upto_seq_num = m_records.active_upto(); }
+
+    // if we have flushed already, we are done
+    if (m_records.completed_upto() >= upto_seq_num) { return; }
+
+    {
+        std::unique_lock lk(m_sync_flush_mtx);
+        while (true) {
```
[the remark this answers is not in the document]
hkadayam: I agree that there is no need of the while loop here. I will remove that and update this PR.

src/lib/logstore/log_store.cpp
```diff
@@ -375,6 +380,34 @@ logstore_seq_num_t HomeLogStore::get_contiguous_completed_seq_num(logstore_seq_n
...
+    {
+        std::unique_lock lk(m_sync_flush_mtx);
+
+        // Step 1: Mark the waiter lsn to the seqnum we wanted to wait for. The completion of every lsn checks
+        // for this and if this lsn is completed, will make a callback which signals the cv.
+        m_sync_flush_waiter_lsn.store(upto_seq_num);
+
+        // Step 2: After marking this lsn, we again do a check, to avoid a race where completion checked for no lsn
+        // and the lsn is stored in step 1 above.
+        if (m_records.completed_upto() >= upto_seq_num) { return; }
```
yamingk, line 398: Thinking it more, we might not even need this check. 
Line 404 would just unblock wait since `m_records.completed_upto() >= upto_seq_num;`.
Or we are doing so to save calling `flush_if_needed`?

hkadayam: Yes it is to save flush_if_needed call.

PR #16, opened 2023-01-10
Rollback of journal/logstore support

hkadayam:
Rollback of logstore requirement is to make sure that every entry after the given seq_num will be invalidated. This needs to be recovery and truncation proof (in the sense after rollback - after restart of logstore we should still not see those entries). It is implemented at layer below logstore - which is log_dev, to make sure the above requirements are met. Everytime a log store is rolledback, log_dev writes the logstore_id and log_id (which is the key for log_dev) into a separate persistent rollback record in its superblock. This marks any log_id corresponding to the given logstore is invalid going forward and can be safely ignored during recovery. Upon truncation of the given log_id, this rollback record can be truncated.
NOTE
1. log_id is common id across log_stores. 
2. lsn (logstore_seq_num) is unique within log_store
3. logstore_id: ID indicating the logstore.

### DIFF
--- a/src/include/homestore/logstore/log_store.hpp
+++ b/src/include/homestore/logstore/log_store.hpp
@@ -40,6 +40,7 @@ class LogDev;
 class LogStoreServiceMetrics;
 
 static constexpr logstore_seq_num_t invalid_lsn() { return std::numeric_limits< logstore_seq_num_t >::min(); }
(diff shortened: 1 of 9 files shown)


szmyd (approved):
LGTM

yamingk (approved):

PR #17, opened 2023-01-10
 SDSTOR-8770 :  RangeRemove Unit tests and fix

shosseinimotlagh:
link: https://jirap.corp.ebay.com/browse/SDSTOR-8770


src/include/homestore/btree/btree_kv.hpp
```diff
@@ -285,6 +285,7 @@ class BtreeTraversalState {
 
     const BtreeKeyRange< K >& input_range() const { return m_input_range; }
     const BtreeKeyRange< K >& working_range() const { return m_working_range; }
+    const BtreeKeyRange< K >& current_range() const { return m_working_range; }
```
hkadayam, line 288: Why do we need duplicate method. Why can't the current_range be changed to working_range in the caller?

src/include/homestore/btree/btree_req.hpp
```diff
@@ -85,6 +85,7 @@ struct BtreeRangeRequest : public BtreeRequest {
     const BtreeKeyRange< K >& input_range() const { return m_search_state.input_range(); }
     const BtreeKeyRange< K >& next_range() { return m_search_state.next_range(); }
     const BtreeKeyRange< K >& working_range() const { return m_search_state.working_range(); }
+    const BtreeKeyRange< K >& current_range() const { return m_search_state.current_range(); }
```
hkadayam, line 88: Same comment as above

src/include/homestore/btree/btree_kv.hpp
shosseinimotlagh: @hkadayam you are right. I haven't cleaned it yet. I need to add some more changes, too.

src/include/homestore/btree/detail/btree_remove_impl.ipp
```diff
@@ -76,15 +82,20 @@ retry:
...
+        //        BT_NODE_REL_ASSERT_NE(count, 0, my_node, "get_all returns 0 entries for interior node is not valid
+        //        pattern");
 
     } else if constexpr (std::is_same_v< ReqT, BtreeRemoveAnyRequest< K > >) {
         const auto count = my_node->template get_all< V >(req.m_range, UINT32_MAX, start_idx, end_idx);
         BT_NODE_REL_ASSERT_NE(count, 0, my_node, "get_all returns 0 entries for interior node is not valid pattern");
         end_idx = start_idx = (end_idx - start_idx) / 2; // Pick the middle, TODO: Ideally we need to pick random
     }
-
+    if (go_to_out) { goto out_return; }
```
hkadayam, line 98: why do we need the go_to_out variable, why can't line 88, be just `goto out_return` ?

shosseinimotlagh: Technically your suggestion is legit. The way that is implemented is due to compilation warning. If I implemented as you suggest, a complation warning would be thrown out that not the label out_retrun has not reached in some code path. 

error: label ‘out_return’ defined but not used [-Werror=unused-label
cc1plus: all warnings being treated as errors

src/include/homestore/btree/detail/varlen_node.hpp
```diff
@@ -510,6 +510,56 @@ class VariableNode : public BtreeNode {
         return str;
     }
 
+    std::string to_string_keys(bool print_friendly = false) const override {
```
hkadayam, line 513: I don't believe we need this method and we can do it as part of to_string() itself. Having said that since the PR has been pending for long, I think its better to merge this and see if we can fold this into to_string() method itself.

hkadayam (approved):

hkadayam (approved):

PR #18, opened 2023-01-21
SDSTOR-8781 Read Blk Tracker

yamingk:
BlkReadTracker implementation.

BlkId will be aligned before we merge them into hash map, this simplifies the implementation by without need to handle the sub ranges of blkids;

Unit test all passed.

### DIFF
--- a/src/lib/blkdata_svc/blk_read_tracker.cpp
+++ b/src/lib/blkdata_svc/blk_read_tracker.cpp
@@ -1,7 +1,6 @@
 /*********************************************************************************
  * Modifications Copyright 2017-2019 eBay Inc.
  *
- *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


src/lib/blkdata_svc/blk_read_tracker.cpp
```diff
@@ -18,69 +17,97 @@
...
-                inserted = true;
-            }
-        }
-        ++it;
+void BlkReadTracker::merge(const BlkId& blkid, int64_t new_ref_count,
+                           const std::shared_ptr< blk_track_waiter >& waiters) {
+    if (new_ref_count == 0) {
+        assert(waiters);
+    } else {
+        assert(!waiters);
```
hkadayam, line 27: Cosmetic suggestion, we can use HS_ASSERT. Also we can brief this as

HS_ASSERT(new_ref_count ? !waiters : waiter, "Invalid waiters");

src/lib/blkdata_svc/blk_read_tracker.hpp
```diff
@@ -18,71 +17,163 @@
...
+    }
 
-struct blk_sub_range_record {
-    blk_sub_range_record(const BlkId& bid, int64_t ref) : sub_range{bid}, ref_count{ref} {}
+    after_remove_cb_t m_cb;
 
-    BlkId sub_range;
-    int64_t ref_count{0};
-    std::shared_ptr< blk_track_waiter > waiter;
+#ifdef _PRERELEASEA
```
hkadayam, line 46: I think it should be
```#ifdef _PRERELEASE```

hkadayam (approved):
Looks good to me

PR #19, opened 2023-01-21
Read before flush and device truncation fix

hkadayam:
2 changes in this PR:
1. There are use cases where read on logstore for a seq_num to return valid value even if those entries are not flushed. This is done by flushing the data if seq_num is not already flushed and then read from the persistent store. Alternative to that would be to read from cache, which would be more performant. However, this could cause a scenario where read could give different result before and after flush. (Thanks @yamingk for pointing that out)
2. There is an issue where if a logstore is created, appended with logs but it has not truncated ever, logdevice truncate (which truncates on least log_id of all logstore) doesn't truncate until the said logstore actually calls truncation. While this is not a huge issue as it is expected that all logstores should call truncate at some point, one can see lots of corner cases. This is fixed by setting the logdev upon first insertion.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -5,7 +5,7 @@ from conans import CMake
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "4.0.2"
+    version = "4.0.3"
     homepage = "https://github.com/eBay/Homestore"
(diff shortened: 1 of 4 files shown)


yamingk (approved):

yamingk:
LGTM

PR #20, opened 2023-01-30
SDSTOR-8781 Data service implementation and tests

yamingk:
Data service layer implementation:
===========================
1. Write request will multiple iovecs will be handled by data service layer and callback will be sent back only all the writes completed with corresponding iovecs; It requires corresponding changes in vdev layer;
3. Read api only write on single blkid. 
4. Read integrates with read blk tracker.
5. Async free blk cb will be called if no read is pending on the blkid. And if any read is pending on the same blkid, its calback will be triggered after read completes;
6. Other misc apis such as streams, etc.
7. Test binary of data service layer;
8. Add a few more debug assert in test read blk tracker to make sure free_blk cb only called once (should never call more than once);
9. All uint test passed, including existing test cases;

### DIFF
--- a/src/include/homestore/blk.h
+++ b/src/include/homestore/blk.h
@@ -141,6 +141,30 @@ std::basic_ostream< charT, traits >& operator<<(std::basic_ostream< charT, trait
 
     return outStream;
 }
+
+/* Hints for various allocators */
+struct blk_alloc_hints {
+    blk_alloc_hints() :
+            desired_temp{0},
+            dev_id_hint{INVALID_DEV_ID},
+            can_look_for_other_chunk{true},
+            is_contiguous{false},
+            multiplier{1},
+            max_blks_per_entry{BlkId::max_blks_in_op()},
+            stream_info{(uintptr_t) nullptr} {}
+
+    blk_temp_t desired_temp;       // Temperature hint for the device
+    uint32_t dev_id_hint;          // which physical device to pick (hint if any) -1 for don't care
+    bool can_look_for_other_chunk; // If alloc on device not available can I pick other device
+    bool is_contiguous;
+    uint32_t multiplier;         // blks allocated in a blkid should be a multiple of multiplier
+    uint32_t max_blks_per_entry; // Number of blks on every entry
+    uintptr_t stream_info;
+#ifdef _PRERELEASE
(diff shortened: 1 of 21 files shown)


src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -13,95 +13,123 @@
...
-    hdr.type = blkstore_type::DATA_STORE;
-
-    m_vdev = std::make_unique< VirtualDev >(hs()->device_manager(), "DataVDev", PhysicalDevGroup::DATA, blkalloc_type,
-                                            sizeof(blkstore_blob), 0, true /* is_stripe */, page_size,
-                                            bind_this(BlkDataService::vdev_io_completion, 1), hdr, size,
-                                            true /* auto_recovery */);
-    m_page_size = page_size;
+
+BlkDataService::BlkDataService() { m_blk_read_tracker = std::make_unique< BlkReadTracker >(); }
+BlkDataService::~BlkDataService() {}
```
hkadayam, line 27: You could declare it .h file as `~BlkDataService() = default`

yamingk: fixed

src/include/homestore/blk.h
```diff
@@ -141,6 +141,30 @@ std::basic_ostream< charT, traits >& operator<<(std::basic_ostream< charT, trait
...
+    blk_temp_t desired_temp;       // Temperature hint for the device
+    uint32_t dev_id_hint;          // which physical device to pick (hint if any) -1 for don't care
+    bool can_look_for_other_chunk; // If alloc on device not available can I pick other device
+    bool is_contiguous;
+    uint32_t multiplier;         // blks allocated in a blkid should be a multiple of multiplier
+    uint32_t max_blks_per_entry; // Number of blks on every entry
+    uintptr_t stream_info;
+#ifdef _PRERELEASE
+    bool error_simulate = false; // can error simulate happen
+#endif
```
hkadayam, line 165: Justa nit pick, we could set default values in the declaration itself instead of calling them in constructor.

hkadayam: Curious why we had to move this here from blk_allocator.h?

hkadayam (approved):
Looks good

PR #24, opened 2023-02-02
SDSTOR-9070 fix the crash loop during

yamingk:
The root cause of the 2nd crash:
1. it happens when disk drives start from a clean state to a very fragmented state, in which the bitmap initially is compressed and upto some point, it becomes uncompressed (when compression ration doesn’t meet limit).
2. the size we set it blk header will still be pointing to compressed size, but real data written is the original one.
3. when we boot up, all the data is read out from disk is intact, but header size doesn’t match and trigger the crash.

### DIFF
--- a/.jenkins/Dockerfile
+++ b/.jenkins/Dockerfile
@@ -5,6 +5,7 @@ LABEL description="Automated HomeStore compilation"
 WORKDIR /output
 
 RUN set -eux; \
+    sed -i 's/master/latest/g' /etc/apt/sources.list \
     apt-get update; \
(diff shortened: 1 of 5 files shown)


hkadayam (commented):
Do we want to update the unit test for the same with this PR, unless it is urgently needed within a day right? Whats the plan for that?

yamingk:
@hkadayam this is a urgent fix, yes, I plan to update the test case with another PR which will come into 1.3 branch either master or stable/v3.x

hkadayam (approved):

PR #28, opened 2023-02-03
Merge fix from master to develop branch

yamingk:


.jenkins/Dockerfile
```diff
@@ -5,6 +5,7 @@ LABEL description="Automated HomeStore compilation"
 WORKDIR /output
 
 RUN set -eux; \
+    sed -i 's/master/latest/g' /etc/apt/sources.list; \
```
hkadayam, line 8: Is this change needed in develop?

src/tests/test_meta_blk_mgr.cpp
```diff
@@ -134,7 +137,7 @@ class VMetaBlkMgrTest : public ::testing::Test {
         return sec.count();
     }
 
-    [[nodiscard]] bool keep_running() {
+        [[nodiscard]] bool keep_running() {
```
hkadayam, line 140: I think there is some misalignment in spaces. Existing change is correct. Perhaps format_on_save not enabled?

.jenkins/Dockerfile
d2chau: I think so, otherwise when building you can still run into the issue when running apt-get update

src/tests/test_meta_blk_mgr.cpp
```diff
@@ -322,12 +327,12 @@ class VMetaBlkMgrTest : public ::testing::Test {
...
     }
 
-    void do_sb_update(const bool aligned_buf_size) {
+    void do_sb_update(const bool aligned_buf_size, uint64_t size_to_update = 0) {
         ++m_update_cnt;
         uint8_t* buf{nullptr};
         auto overflow = do_overflow();
         if (!aligned_buf_size) { overflow = true; } // for unaligned buf size, let's generate overflow buf size;
-        auto sz_to_wrt = rand_size(overflow, aligned_buf_size);
+        auto sz_to_wrt{size_to_update ? size_to_update : rand_size(overflow, aligned_buf_size)};
```
hkadayam, line 335: I would prefer consistency with develop branch and do `=` for assignments and use `{}` only for object building.

src/tests/test_meta_blk_mgr.cpp
```diff
@@ -415,6 +420,31 @@ class VMetaBlkMgrTest : public ::testing::Test {
...
+    //
+    void write_compression_backoff() {
+        HS_SETTINGS_FACTORY().modifiable_settings([](auto& s) {
+            s.metablk.compress_ratio_limit = 100; // this will allow every compression attempt;
+            HS_SETTINGS_FACTORY().save();
+        });
+
+        LOGINFO("compression ratio limit changed to: {}", m_mbm->get_compress_ratio_limit());
+
+        [[maybe_unused]] const auto write_result = do_sb_write(true /* do_overflow */, 15 * Mi);
```
hkadayam, line 435: we removed the nodiscard from all apis, so this annoying maybe_unused assignment can be skipped and just do ```do_sb_write(...)```

src/tests/test_meta_blk_mgr.cpp
yamingk: my clang format plugin seems to be always get fooled with this keyword "[[nodiscard]]", let me fix it.

src/tests/test_meta_blk_mgr.cpp
yamingk: will do

src/tests/test_meta_blk_mgr.cpp
yamingk: hmm... it yields build failure for me in master, maybe it is not the case in develop branch.

.jenkins/Dockerfile
yamingk: every branch is failing without this line in image state including AM

src/tests/test_meta_blk_mgr.cpp
yamingk: /tmp/source/yk_develop/src/tests/test_meta_blk_mgr.cpp:435:20: error: ignoring return value of âuint64_t VMetaBlkMgrTest::do_sb_write(bool, size_t)â, declared with attribute ânodiscardâ [-Werror=unused-result]
  435 |         do_sb_write(true /* do_overflow */, 15 * Mi);
      |         ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

szmyd (dismissed):

d2chau (approved):

PR #31, opened 2023-02-07
SDSTOR-9070 fixing mblk size during upgrade

yamingk:


shosseinimotlagh (approved):
LGTM

hkadayam (commented):
Looks good, but whats the strategy to revert this change?

yamingk:
> Looks good, but whats the strategy to revert this change?

We will have to make another image to remove the hot fix (also changing the default value of the dynamic config).

PR #32, opened 2023-02-07
SDSTOR-xxxx fixing compressed filed as false during upgrade to fix ba…

yamingk:
…d data


shosseinimotlagh (approved):
LGTM

src/engine/meta/meta_blks_mgr.cpp
```diff
@@ -312,7 +312,11 @@ bool MetaBlkMgr::scan_and_load_meta_blks(meta_blk_map_t& meta_blks, ovf_hdr_map_
         if (read_sz != static_cast< uint64_t >(mblk->hdr.h.context_sz)) {
             LOGERROR("[type={}], total size read: {} mismatch from meta blk context_sz: {}", mblk->hdr.h.type, read_sz,
                      mblk->hdr.h.context_sz);
```
shosseinimotlagh, line 314: shouldn't be change to INFO ?

yamingk: It is real error, we are just recovering from an error.

PR #33, opened 2023-02-08
SDSTOR-xxxx add new test case to fix bad data during upgrade

yamingk:
1. added a new test case to simulate prod issue that bad data is written to disk and upgrade to a new image with the fix and verify bad data can be recovered;
2. added a flip point to turn on/off the fix, so that we can generate bad data without the fix
3. Testing steps can be found in test case
4. add to nightly test

Test passed by running the meta blkstore nightly script locally.

Update: 2/8: Jenkinds PR build passed.

### DIFF
--- a/src/engine/meta/meta_blks_mgr.cpp
+++ b/src/engine/meta/meta_blks_mgr.cpp
@@ -337,7 +337,7 @@ bool MetaBlkMgr::scan_and_load_meta_blks(meta_blk_map_t& meta_blks, ovf_hdr_map_
             // we are here because we write uncompressed data, but left compressed field as true and context_sz still
             // setting to compressed size;
 
-            if (HS_DYNAMIC_CONFIG(metablk.skip_header_size_check) == false) {
+            if (!(HS_DYNAMIC_CONFIG(metablk.skip_header_size_check))) {
                 HS_REL_ASSERT_EQ(read_sz, static_cast< uint64_t >(mblk->hdr.h.context_sz),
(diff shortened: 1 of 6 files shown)


src/engine/meta/meta_blks_mgr.cpp
```diff
@@ -331,9 +331,27 @@ bool MetaBlkMgr::scan_and_load_meta_blks(meta_blk_map_t& meta_blks, ovf_hdr_map_
...
-        HS_REL_ASSERT_EQ(read_sz, static_cast< uint64_t >(mblk->hdr.h.context_sz),
-                         "[type={}], total size read: {} mismatch from meta blk context_sz: {}", mblk->hdr.h.type,
-                         read_sz, mblk->hdr.h.context_sz);
+        if (read_sz != static_cast< uint64_t >(mblk->hdr.h.context_sz)) {
+            LOGERROR("[type={}], total size read: {} mismatch from meta blk context_sz: {}", mblk->hdr.h.type, read_sz,
+                     mblk->hdr.h.context_sz);
+            // we are here because we write uncompressed data, but left compressed field as true and context_sz still
+            // setting to compressed size;
+
+            if (HS_DYNAMIC_CONFIG(metablk.skip_header_size_check) == false) {
```
shosseinimotlagh, line 340: NIT: if (!HS_DYNAMIC_CONFIG(metablk.skip_header_size_check))

yamingk: done

shosseinimotlagh (approved):
LGTM

PR #34, opened 2023-02-09
CPManager test cases

hkadayam:
Following changes are performed as part of this PR
1. Added new CPManager test case to test cp flow
2. Change the CPCallbacks from bunch of function pointers to virtual abstract class to force user to define certain mandatory methods. In general it is little bit more cleaner
3. Given the start_homestore is being called by lots of tests, added test_common::HSTestHelper class with some common start/stop homestore methods.


yamingk (approved):

yamingk:
Looks good to me.

PR #35, opened 2023-02-10
SDSTOR-9150 add new homestore data service apis

yamingk:
1. Add a new API for data service that will take block ids as input for async_write;
2. add doxygen comment for public APIs
3. new api was tested by home replication storage test binary;

### DIFF
--- a/src/include/homestore/blkdata_service.hpp
+++ b/src/include/homestore/blkdata_service.hpp
@@ -72,8 +72,8 @@ public:
      * @param cb : callback that will be triggered after write completes;
      * @param part_of_batch : is this write part of a batch;
      */
-    void async_write(const sisl::sg_list& sgs, const blk_alloc_hints& hints, std::vector< BlkId >& out_blkids,
-                     const io_completion_cb_t& cb, bool part_of_batch = false);
+    void async_alloc_write(const sisl::sg_list& sgs, const blk_alloc_hints& hints, std::vector< BlkId >& out_blkids,
+                           const io_completion_cb_t& cb, bool part_of_batch = false);
 
     /**
      * @brief : asynchronous write with input block ids;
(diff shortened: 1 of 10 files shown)


src/include/homestore/blkdata_service.hpp
```diff
@@ -52,59 +52,120 @@ class BlkDataService {
...
+
+    /**
+     * @brief : asynchronous read
+     *
+     * @param bid : block id to read
+     * @param sgs : the read buffer stored
+     * @param size : size to read
+     * @param cb : callback that will be triggered after read completes
+     * @param part_of_batch : is this read part of batch;
+     */
```
hkadayam, line 99: My suggestion is to rename `async_write_ahead()` to `async_write()` and current `async_write()` to `async_alloc_write()`. Write ahead in general has different meaning right (similar to read_ahead)?

yamingk: I will take your suggestion, let me update the PR.

hkadayam (approved):
Looks good

src/include/homestore/homestore.hpp
```diff
@@ -132,14 +132,12 @@ class HomeStore {
...
-    HomeStore& with_data_service(float size_pct);
-    HomeStore& with_log_service(float data_size_pct, float ctrl_size_pct);
-    HomeStore& with_index_service(float index_size_pct, std::unique_ptr< IndexServiceCallbacks > cbs);
     HomeStore& after_init_done(hs_init_done_cb_t init_done_cb);
     HomeStore& before_init_devices(hs_init_starting_cb_t init_start_cb);
 
-    void init(bool wait_for_init = false);
+    void init_internal(bool wait_for_init);
+    void init(bool wait_for_init = false, float meta_size_pct = 0, float log_data_size_pct = 0,
+              float log_ctrl_size_pct = 0, float data_svc_size_pct = 0);
```
hkadayam, line 140: I didn't notice this change, can we please revert this. The idea of `.with` is to be explicit. If you don't agree with it, we can discuss and finalize it?

hkadayam (commented):
Can we discuss and finalize the .init() API part. I liked the way explicitly setting up the service. In future we can add more service seamlessly.

hkadayam (approved):
Looks good to me

PR #37, opened 2023-02-16
CPGuard and also made cp critical section re-entrant

hkadayam:


src/lib/checkpoint/cp_mgr.cpp
```diff
@@ -69,24 +71,29 @@ void CPManager::register_consumer(cp_consumer_t consumer_id, std::unique_ptr< CP
     }
 }
 
+[[nodiscard]] CPGuard CPManager::cp_guard() { return CPGuard{}; }
```
szmyd, line 74: Why isn't this just defined in the header where CPGuard(); is also declared public? Also, can this be static or at least `const`? It seems to use no instance fields.

src/lib/checkpoint/cp_mgr.cpp
```diff
@@ -69,24 +71,29 @@ void CPManager::register_consumer(cp_consumer_t consumer_id, std::unique_ptr< CP
...
     }
+    cp_ref(cp);
+    rcu_read_unlock();
+
+    return cp;
+}
 
-    auto cnt = cp->m_enter_cnt.fetch_add(1);
+void CPManager::cp_ref(CP* cp) {
+    cp->m_enter_cnt.fetch_add(1);
```
szmyd, line 90: Should this have `memory_order_relaxed` ordering since it's not protecting anything (that I see)?

szmyd (commented):
Questions...

src/lib/checkpoint/cp_mgr.cpp
```diff
@@ -108,45 +115,45 @@ void CPManager::trigger_cp_flush(cp_done_cb_t&& cb, bool force) {
         // They are already an existing CP on-going, but if force is set, we create a back-to-back CP.
```
yamingk, line 115: nit: there is

src/include/homestore/checkpoint/cp_mgr.hpp
```diff
@@ -134,10 +112,42 @@ struct cp_mgr_super_block {
...
+    CP* m_cp{nullptr};
+    bool m_pushed{false};
+
+    // Why we need this thread_local variable and that too of type stack?
+    // thread_local variable is needed because we wanted to make cp critical section re-entrant. So when a thread enters
+    // into a critical section and crosses methods and other code within the stack needs to enter to critical section,
+    // having this facility make sure that it uses already entered critical section within the stack.
+    //
+    // Why do we need a stack instead of only one CP* to track current critical section?
+    // It is because CPGuard can be moved from one thread to other. The thread which it is moved to can be accessed
```
yamingk, line 126: here in this example, cp1 and cp2 is the same cp or different cp?

src/lib/checkpoint/cp_mgr.cpp
```diff
@@ -205,6 +214,54 @@ void CPManager::cleanup_cp(CP* cp) {
...
+CPGuard::CPGuard() {
+    if (t_cp_stack.empty()) {
+        // First CP in this thread stack.
+        m_cp = hs()->cp_mgr().cp_io_enter();
+    } else {
+        // Nested CP sections
+        m_cp = t_cp_stack.top();
+        hs()->cp_mgr().cp_ref(m_cp);
+    }
+    t_cp_stack.push(m_cp);
```
yamingk, line 227: for nested cp case, here we are pushing same cp to stack, right? There is no problem with that, just trying to make sure I understand correctly.

src/lib/checkpoint/cp_mgr.cpp
```diff
@@ -69,24 +71,29 @@ void CPManager::register_consumer(cp_consumer_t consumer_id, std::unique_ptr< CP
     }
 }
 
+[[nodiscard]] CPGuard CPManager::cp_guard() { return CPGuard{}; }
+
 CP* CPManager::cp_io_enter() {
     rcu_read_lock();
     auto cp = get_cur_cp();
     if (!cp) {
         rcu_read_unlock();
```
yamingk, line 80: When we fall into this if, we return nullptr to caller, which is in this PR will be `CPGuard::CPGuard`, is it expected that we will be pushing nullptr to the stack?

src/lib/checkpoint/cp_mgr.cpp
```diff
@@ -205,6 +214,54 @@ void CPManager::cleanup_cp(CP* cp) {
     }
 }
 
+//////////////////////////////////////// CP Guard class ////////////////////////////////////////////
+CPGuard::CPGuard() {
+    if (t_cp_stack.empty()) {
+        // First CP in this thread stack.
+        m_cp = hs()->cp_mgr().cp_io_enter();
```
yamingk, line 221: m_cp can be nullptr in some case, is it expected?

src/lib/checkpoint/cp_mgr.cpp
hkadayam: This was original code as written in 1.2/1.3, I didn't modify it. However, I don't think it can be `memory_order_relaxed` because the enter_cnt once become zero, we need a barrier. This is same pattern we have in many part of our code and we generalized in sisl::atomic_counter (which uses acq/release pairs). I will replace that with atomic_counter, so we have one place to modify.

src/lib/checkpoint/cp_mgr.cpp
hkadayam: CPManager all other methods are singleton. I can understand it seems weird. CPManager is a singleton and we need to call the current instance of cp_manager() cp_io_enter(). So I have 2 options
1) Make this `const`, but it is technically not a const, because CPGuard calls cp_io_enter() which is not a const
2) CPGuard constructor will pass CPManager non const pointer as well and it maintains inside it, (so copy constructor can use that CPManager instance).

src/include/homestore/checkpoint/cp_mgr.hpp
hkadayam: cp1 and cp2 are different cps.

src/lib/checkpoint/cp_mgr.cpp
hkadayam: yes, either we need to push or maintain same pointer and count pair. This seems less penalty and also code looks much cleaner this way (I had a version like you mentioned - but it seems not clean). Also this case is generally rare to be more than 2 in this stack.

src/lib/checkpoint/cp_mgr.cpp
hkadayam: It is ok for us to push nullptr, because its destructor checks if cp is nullptr (exactly to cover this case).

src/lib/checkpoint/cp_mgr.cpp
hkadayam: yes you are right. I can remove the assert there. But in reality cp_io_enter() returns nullptr can happen only after CPManager::shutdown has been called and so that should not happen, but if it happens is it better to fire off the assert or not?

yamingk: shutdown should not be triggered if any cp is in progress, right? If so, we probably should keep the assert.

yamingk (approved):

src/lib/checkpoint/cp_mgr.cpp
szmyd: I don't think I follow this pattern...how can it matter where `CPGuard()` is called from (inside method of CPManager or not)? There is nothing passed to it to link it back to the instance that called it that I see?

hkadayam: I did the option 2) above, so now it always pass (this) pointer from CPManager.

szmyd (approved):
lgtm

PR #39, opened 2023-02-28
Disable realtime bitmap in release

shosseinimotlagh:


src/engine/blkalloc/blk_allocator.cpp
```diff
@@ -28,6 +28,13 @@ BlkAllocator::BlkAllocator(const BlkAllocConfig& cfg, const chunk_num_t id) : m_
     const auto bitmap_id{id};
     m_disk_bm = std::make_unique< sisl::Bitset >(m_cfg.get_total_blks(), bitmap_id, align_size);
 
+#if defined _PRERELEASE || defined DEBUG
```
yamingk, line 31: I think we should move this logic inside BlkAllocatorConfig::BlkAllocatorConfig so that there is only one place that sets the realtime bitmap. For 1.2 we can just leave as it is.

shosseinimotlagh: Yes. I believe your suggestion makes more sense than this one which is identical to 1.2!

src/engine/blkalloc/blk_allocator.h
```diff
@@ -69,20 +69,19 @@ class BlkAllocConfig {
...
-#else
-        // for release build, take it from dynamic config which is defaulted to false
-        m_realtime_bm_on = HS_DYNAMIC_CONFIG(blkallocator.realtime_bitmap_on);
+
+#if defined _PRERELEASE || defined DEBUG
+        if (!HS_DYNAMIC_CONFIG(blkallocator.realtime_bitmap_on))
 #endif
-    }
+        {
+            m_realtime_bm_on = false;
```
yamingk, line 83: in pre-release or debug build, how are you able to turn this feature on if blkallocator.realtime_bitmap_on is set to true?

yamingk (approved):

PR #51, opened 2023-04-26
migrate to Pistache for http library

raakella1:


src/homeblks/home_blks.cpp
```diff
@@ -177,7 +176,11 @@ HomeBlks::HomeBlks(const init_params& cfg) :
     m_recovery_stats = std::make_unique< HomeBlksRecoveryStats >();
     m_recovery_stats->start();
 
-    if (HB_DYNAMIC_CONFIG(general_config->boot_safe_mode)) { LOGINFO("HomeBlks booting into safe_mode"); }
+    if (HB_DYNAMIC_CONFIG(general_config->boot_safe_mode)) {
+        LOGINFO("HomeBlks booting into safe_mode");
+        // start http server
+        m_hb_http_server->start();
```
yamingk, line 182: if we are not in safe_mode, we will not start http server, shouldn't we also start it in non-safe-mode?

yamingk (approved):

PR #62, opened 2023-06-08
SDSTOR-8770 Fix range remove and add UTs

shosseinimotlagh:


src/include/homestore/btree/detail/btree_internal.hpp
```diff
@@ -242,7 +242,8 @@ struct BtreeConfig {
     uint8_t m_suggested_min_pct{30};
     uint8_t m_split_pct{50};
     uint32_t m_max_merge_nodes{3};
-    uint32_t m_rebalance_turned_on{false};
+    bool m_rebalance_turned_on{false};
```
sanebay, line 245: Why are both false by default ?

src/include/homestore/btree/detail/btree_remove_impl.ipp
```diff
@@ -36,16 +37,19 @@ btree_status_t Btree< K, V >::do_remove(const BtreeNodePtr& my_node, locktype_t
...
-                    modified = true;
+                if (my_node->total_entries()) {
+                    auto& subrange = req.working_range();
+                    auto const [start_found, start_idx] = my_node->find(subrange.start_key(), nullptr, false);
+                    auto [end_found, end_idx] = my_node->find(subrange.end_key(), nullptr, false);
+                    if (end_found) { ++end_idx; }
+
+                    removed_count = end_idx - start_idx;
+                    for (auto idx{start_idx}; idx < end_idx; ++idx) {
+                        call_on_remove_kv_cb(my_node, start_idx, req);
```
sanebay, line 48: Please add a comment that use start indx again because of memmove.

src/tests/test_mem_btree.cpp
```diff
@@ -349,6 +407,92 @@ TYPED_TEST(BtreeTest, RangeUpdate) {
...
+    this->range_remove(0, 2);
+    //    this->print_keys(); // EXPECT size = 11 : [0 1 2] 3 4 11 12 13 14 15 16 17 18 19
+    this->range_remove(18, 19);
+    //    this->print_keys(); // EXPECT size = 9 : 3 4 11 12 13 14 15 16 17 [18 19]
+    this->range_remove(17, 17);
+    //    this->print_keys(); // EXPECT size = 8 : 3 4 11 12 13 14 15 16 [17]
+    this->range_remove(1, 5);
+    //    this->print_keys(); // EXPECT size = 6 : [3 4] 11 12 13 14 15 16
+    this->range_remove(1, 20);
+    //    this->print_keys(); // EXPECT size = 0 : [11 12 13 14 15 16]
```
sanebay, line 430: Please add one more range_remove on empty tree. It should be no op.

src/tests/test_mem_btree.cpp
```diff
@@ -349,6 +407,92 @@ TYPED_TEST(BtreeTest, RangeUpdate) {
...
+TYPED_TEST(BtreeTest, SimpleRemoveRange) {
+    // Forward sequential insert
+    const auto num_entries = 20;
+    LOGINFO("Step 1: Do Forward sequential insert for {} entries", num_entries);
+    for (uint32_t i{0}; i < num_entries; ++i) {
+        this->put(i, btree_put_type::INSERT_ONLY_IF_NOT_EXISTS);
+    }
+
+    //    this->print_keys(); // EXPECT size = 20 : 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19
+    this->range_remove(5, 10);
```
sanebay, line 419: Could we validate the values which were removed, not only the size.

src/tests/test_mem_btree.cpp
```diff
@@ -349,6 +407,92 @@ TYPED_TEST(BtreeTest, RangeUpdate) {
...
+    static std::uniform_int_distribution< uint32_t > s_rand_key_generator{0, 2 * num_entries};
+    //    this->print_keys();
+    for (uint32_t i{0}; i < SISL_OPTIONS["num_iters"].as< uint32_t >() && this->m_shadow_map.size() > 0; ++i) {
+        uint32_t key1 = s_rand_key_generator(g_re);
+        uint32_t key2 = s_rand_key_generator(g_re);
+        uint32_t start_key = std::min(key1, key2);
+        uint32_t end_key = std::max(key1, key2);
+
+        LOGINFO("Step 2 - {}: Do Range Remove of maximum [{},{}] keys ", i, start_key, end_key);
+        this->range_remove(std::min(key1, key2), std::max(key1, key2));
```
sanebay, line 490: Validate values removed

src/include/homestore/btree/detail/btree_internal.hpp
shosseinimotlagh: I set `m_merge_turned_on` to be true. If the number of calculated new nodes is greater than if  number of current node, the value of rebalance determines that merge needed or not, while m_merge_turned_on is to enable the merge. Later we probably need to introduce a flag to enable/disable the merge during runtime if it is costly.

src/include/homestore/btree/detail/varlen_node.hpp
```diff
@@ -62,6 +62,10 @@ class VariableNode : public BtreeNode {
 
     virtual ~VariableNode() = default;
 
+    uint32_t occupied_size(const BtreeConfig& cfg) const {
```
szmyd, line 65: `override`?

szmyd (changes requested):

src/include/homestore/btree/detail/varlen_node.hpp
shosseinimotlagh: fixed

szmyd (approved):
lgtm

PR #121, opened 2023-07-05
Fix UT failures and fetch log store records batch size.

sanebay:


src/homelogstore/log_store.cpp
```diff
@@ -352,7 +356,8 @@ nlohmann::json HomeLogStore::dump_log_store(const log_dump_req& dump_req) {
             }
             json_records.emplace_back(std::move(json_val));
             decltype(idx) end_idx{std::min(max_idx, dump_req.end_seq_num)};
-            end_iterate = (cur_idx < end_idx) ? true : false;
+            end_iterate = (cur_idx < end_idx && --batch_size > 0) ? true : false;
```
yamingk, line 359: shouldn't it be --batch_size == 0 ?

yamingk (approved):

shosseinimotlagh (dismissed):
LGTM

src/homelogstore/log_store.cpp
sanebay: I modified the variable name to proceed to make it clear.

src/homelogstore/log_store.cpp
```diff
@@ -352,8 +356,9 @@ nlohmann::json HomeLogStore::dump_log_store(const log_dump_req& dump_req) {
             }
             json_records.emplace_back(std::move(json_val));
             decltype(idx) end_idx{std::min(max_idx, dump_req.end_seq_num)};
-            end_iterate = (cur_idx < end_idx) ? true : false;
-            return end_iterate;
+            proceed = (cur_idx < end_idx && --batch_size == 0) ? false : true;
+            if (cur_idx < max_idx && batch_size == 0) { json_dump["next_cursor"] = std::to_string(cur_idx + 1); }
```
yamingk, line 360: if (proceed) { ... }

src/homelogstore/log_store.cpp
```diff
@@ -352,8 +356,10 @@ nlohmann::json HomeLogStore::dump_log_store(const log_dump_req& dump_req) {
             }
             json_records.emplace_back(std::move(json_val));
             decltype(idx) end_idx{std::min(max_idx, dump_req.end_seq_num)};
-            end_iterate = (cur_idx < end_idx) ? true : false;
-            return end_iterate;
+            proceed = (cur_idx < end_idx && --batch_size > 0) ? true : false;
+            // User can provide either the end_seq_num or batch_size in the request.
+            if (cur_idx < end_idx && batch_size == 0) { json_dump["next_cursor"] = std::to_string(cur_idx + 1); }
```
yamingk, line 361: if batch_size hasn't reached to zero yet, but cur_idx has advanced to be as equal to end_idx, we should also set "next_cursor", right?

yamingk: I think what we need is just:

if (!proceed) {...}

yamingk (approved):

PR #123, opened 2023-07-05
Convert dump content in logstore/Metablk to base64 encoded format and add Metablk object view

shosseinimotlagh:
Sample:
python  hs_http_cli_tool.py get_object --name=MetaBlkMgr
python  hs_http_cli_tool.py cli_tool get_object --type=MetaBlk
python  hs_http_cli_tool.py get_object --type=MetaBlk --batch_size=3  --next 
python  hs_http_cli_tool.py get_object --name=MetaBlk_LogStoreFamily0  --verbosity=2
 python  hs_http_cli_tool.py get_object  --name=MetaBlk_LogStoreFamily0 --verbosity=3

### DIFF
--- a/src/api/meta_interface.hpp
+++ b/src/api/meta_interface.hpp
@@ -43,6 +43,10 @@ class BlkBuffer;
 template < typename Buffer >
 class BlkStore;
 
+// The offset of vol_name is needed for metablk service to translate the vol_name from buffer of metablk_VOLUME for
+// example in get_status
+#define VOL_NAME_OFFSET 48
+
 typedef homestore::BlkStore< BlkBuffer > blk_store_t;
 
 // each subsystem could receive callbacks multiple times
@@ -89,7 +93,6 @@ private:
     sisl::blob m_compress_info;
     MetablkMetrics m_metrics;
     bool m_inited{false};
-    sisl::sobject_ptr m_sobject;
 
 public:
     MetaBlkMgr(const char* const name = "MetaBlkStore");
@@ -221,8 +224,8 @@ public:
     [[nodiscard]] uint32_t get_align_size() const;
 
     [[nodiscard]] sisl::status_response get_status(const sisl::status_request& request);
-
-    sisl::sobject_ptr sobject() { return m_sobject; }
+    [[nodiscard]] sisl::status_response get_status_metablk(const sisl::status_request& request, std::string type = "");
+    void create_sobject(meta_sub_type type);
 
 public:
     /*********************** static public function **********************/
@@ -345,9 +348,11 @@ private:
     [[nodiscard]] uint64_t get_min_compress_size() const;
     [[nodiscard]] uint64_t get_max_compress_memory_size() const;
     [[nodiscard]] uint64_t get_init_compress_memory_size() const;
+
 public:
     [[nodiscard]] uint32_t get_compress_ratio_limit() const;
     [[nodiscard]] bool get_skip_hdr_check() const;
+
 private:
     [[nodiscard]] bool compress_feature_on() const;
 
--- a/src/engine/common/homestore_utils.cpp
+++ b/src/engine/common/homestore_utils.cpp
@@ -15,6 +15,9 @@
  *********************************************************************************/
 #include "homestore_utils.hpp"
 #include "homestore_assert.hpp"
+#include <boost/archive/iterators/base64_from_binary.hpp>
+#include <boost/archive/iterators/binary_from_base64.hpp>
+#include <boost/archive/iterators/transform_width.hpp>
 
 namespace homestore {
 uint8_t* hs_utils::iobuf_alloc(const size_t size, const sisl::buftag tag, const size_t alignment) {
@@ -72,6 +75,50 @@ sisl::byte_array hs_utils::extract_byte_array(const sisl::byte_view& b, const bo
                                               const size_t alignment) {
     return (is_aligned_needed) ? b.extract(alignment) : b.extract(0);
 };
+
+std::string hs_utils::encodeBase64(const uint8_t* first, std::size_t size) {
+    using Base64FromBinary = boost::archive::iterators::base64_from_binary<
+        boost::archive::iterators::transform_width< const char*, // sequence of chars
+                                                    6,           // get view of 6 bit
+                                                    8            // from sequence of 8 bit
+                                                    > >;
+    std::vector< unsigned char > bytes{first, first + size};
+    std::size_t bytes_to_pad = (3 - size % 3) % 3;
+    if (bytes_to_pad > 0) { bytes.resize(bytes.size() + bytes_to_pad, 0); }
+    std::string encoded{Base64FromBinary{bytes.data()}, Base64FromBinary{bytes.data() + (bytes.size() - bytes_to_pad)}};
+
+    return encoded.append(bytes_to_pad, '=');
+}
+
+std::string hs_utils::encodeBase64(const sisl::byte_view& b){
+    return encodeBase64(b.bytes(), b.size());
+}
+
+template <typename T>
+void hs_utils::decodeBase64(const std::string &encoded_data, T out)
+{
+    using BinaryFromBase64 = boost::archive::iterators::transform_width<
+        boost::archive::iterators::binary_from_base64<std::string::const_iterator>,
+        8, // get a view of 8 bit
+        6  // from a sequence of 6 bit
+        >;
+    auto unpadded_data = encoded_data;
+    const auto bytes_to_pad = std::count(begin(encoded_data), end(encoded_data), '=');
+    std::replace(begin(unpadded_data), end(unpadded_data), '=', 'A'); // A_64 == \0
+
+    std::string decoded_data{BinaryFromBase64{begin(unpadded_data)},
+                        BinaryFromBase64{begin(unpadded_data) + unpadded_data.length()}};
+
+    decoded_data.erase(end(decoded_data) - bytes_to_pad, end(decoded_data));
+    std::copy(begin(decoded_data), end(decoded_data), out);
+}
+
+std::string hs_utils::decodeBase64(const std::string &encoded_data)
+{
+    std::string rv;
+    decodeBase64(encoded_data, std::back_inserter(rv));
+    return rv;
+}
 
 size_t hs_utils::m_btree_mempool_size;
 } // namespace homestore
--- a/src/engine/meta/meta_blks_mgr.cpp
+++ b/src/engine/meta/meta_blks_mgr.cpp
@@ -90,10 +90,9 @@ void MetaBlkMgr::start(blk_store_t* sb_blk_store, const sb_blkstore_blob* blob, 
 
     HS_REL_ASSERT_GT(get_page_size(), META_BLK_HDR_MAX_SZ);
     HS_REL_ASSERT_GT(get_page_size(), MAX_BLK_OVF_HDR_MAX_SZ);
-
-    auto hs = HomeStoreBase::safe_instance();
-    m_sobject = hs->sobject_mgr()->create_object("module", "MetaBlkMgr",
-                                                   std::bind(&MetaBlkMgr::get_status, this, std::placeholders::_1));
+    auto sm{HomeStoreBase::safe_instance()->sobject_mgr()};
+    sm->create_object("module", "MetaBlkMgr",
+                                                 std::bind(&MetaBlkMgr::get_status, this, std::placeholders::_1));
 
     reset_self_recover();
     alloc_compress_buf(get_init_compress_memory_size());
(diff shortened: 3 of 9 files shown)


yamingk:
Can we add a new test case to cover this functionality?

src/engine/common/homestore_utils.cpp
```diff
@@ -73,5 +76,49 @@ sisl::byte_array hs_utils::extract_byte_array(const sisl::byte_view& b, const bo
...
+    return encoded.append(bytes_to_pad, '=');
+}
+
+std::string hs_utils::encodeBase64(const sisl::byte_view& b){
+    return encodeBase64(b.bytes(), b.size());
+}
+
+template <typename T>
+void hs_utils::decodeBase64(const std::string &encoded_data, T out)
+{
```
yamingk, line 99: Have you tested the decode function?

src/engine/common/homestore_utils.cpp
```diff
@@ -73,5 +76,49 @@ sisl::byte_array hs_utils::extract_byte_array(const sisl::byte_view& b, const bo
...
+    using BinaryFromBase64 = boost::archive::iterators::transform_width<
+        boost::archive::iterators::binary_from_base64<std::string::const_iterator>,
+        8, // get a view of 8 bit
```
yamingk, line 102: Is it expected that it is reverse order of view and sequence in encode function?

src/engine/common/homestore_utils.cpp
shosseinimotlagh: I did as a part of test_meta_blk_mgr

src/engine/common/homestore_utils.cpp
shosseinimotlagh: There is no reverse

src/engine/common/homestore_utils.cpp
```diff
@@ -73,5 +76,49 @@ sisl::byte_array hs_utils::extract_byte_array(const sisl::byte_view& b, const bo
     return (is_aligned_needed) ? b.extract(alignment) : b.extract(0);
 };
 
+std::string hs_utils::encodeBase64(const uint8_t* first, std::size_t size) {
+    using Base64FromBinary = boost::archive::iterators::base64_from_binary<
+        boost::archive::iterators::transform_width< const char*, // sequence of chars
+                                                    6,           // get view of 6 bit
```
yamingk, line 82: view: 6 bit
Sequence: 8 bit

Is it expected in decode function below:
View is set to 8 bit
Seq is set to 6 bit

src/engine/meta/meta_blks_mgr.cpp
```diff
@@ -235,6 +237,36 @@ void MetaBlkMgr::scan_meta_blks() {
...
+        auto meta_blk_obj =
+            sm->create_object(metablk_type, metablk_name,
+                              std::bind(&MetaBlkMgr::get_status_metablk, this, std::placeholders::_1, metablk_name));
+        m_sobject->add_child(meta_blk_obj);
+    }
+    // create child for each type and add them to child list
+    auto meta_blk_obj{m_sobject->get_child(metablk_name)};
+    auto metablk_meta_bids_name{metablk_name + "_" + std::to_string(meta_bid)};
+    auto metablk_meta_bids_type{metablk_name};
+    auto metablk_meta_bids_obj{sm->create_object(
```
yamingk, line 255: I didn't get why we need this level of object and child, shouldn't it be done through `MetaBlkMgr::get_status_metablk`?

src/engine/meta/meta_blks_mgr.cpp
```diff
@@ -1524,7 +1664,9 @@ nlohmann::json MetaBlkMgr::populate_json(const int log_level, meta_blk_map_t& me
                         const std::string file_path{fmt::format("{}/{}_{}", dump_dir, x.first, bid_cnt)};
                         std::ofstream f{file_path};
                         f.write(reinterpret_cast< const char* >(buf->bytes), buf->size);
-                        j[x.first]["meta_bids"][std::to_string(bid_cnt)] = file_path;
+                        j[x.first]["meta_bids"][std::to_string(bid_cnt)] =
```
yamingk, line 1667: this is not meta_bids, but content.

shosseinimotlagh: I ll change it in the revision

src/engine/common/homestore_utils.cpp
shosseinimotlagh: correct. During encoding, every sequence of 8 bits are converted to the view of 6 bits. In contrast, in decode every 6 bits are converted to 8 bits.
in base X64 encode BitsOut = 6, BitsIn = 8

src/engine/meta/meta_blks_mgr.cpp
shosseinimotlagh: I think one of main objectives of sobject is to make the hierarchical points of view. For this class for example, the meta block of each volume is a child of metablk VOLUME.

yamingk: Shouldn't `--type MetaBlk --name MetaBlk_Volume` (which you've already created the object at line:248) already served the purpose to get all the vol meta blks?

shosseinimotlagh: All registered volumes have the same type `MetaBlk_Volume`

src/engine/meta/meta_blks_mgr.cpp
```diff
@@ -32,6 +32,7 @@
 #include "engine/common/homestore_flip.hpp"
 #include "engine/homestore_base.hpp"
 #include "meta_sb.hpp"
+#include "homeblks/volume/volume.hpp"
```
szmyd, line 35: Should not be including `homeblks` into the MetaBlkMgr; this will not work in 4.x and possibly make circular dependency here.

szmyd (changes requested):

src/engine/meta/meta_blks_mgr.cpp
```diff
@@ -1435,9 +1447,90 @@ bool MetaBlkMgr::sanity_check(const bool check_ovf_chain) {
...
+                    if (bid_cnt < start_sub_type || bid_cnt > end_sub_type) {
+                        bid_cnt++;
+                        continue;
+                    }
+                    BlkId bid(y);
+                    auto it = m_meta_blks.find(y);
+                    std::string jname = "content";
+                    sisl::byte_array buf;
+                    if (client == "VOLUME" || log_level == 3) {
+                        HS_DBG_ASSERT_EQ(it != m_meta_blks.end(), true,
```
yamingk, line 1505: Remove all ASSERT and convert them to error response to http request. We can't crash in produce by running http request.

src/engine/meta/meta_blks_mgr.cpp
yamingk: Forward declaration should be good enough. Please do not include any volume specific header;

src/engine/meta/meta_blks_mgr.cpp
```diff
@@ -116,6 +117,7 @@ void MetaBlkMgr::stop() {
         {
             std::lock_guard< decltype(m_meta_mtx) > lg{m_meta_mtx};
             m_sub_info.clear();
+            // TODO remove sobject
```
yamingk, line 120: remove this line

src/api/meta_interface.hpp
```diff
@@ -221,6 +221,8 @@ class MetaBlkMgr {
     [[nodiscard]] uint32_t get_align_size() const;
 
     [[nodiscard]] sisl::status_response get_status(const sisl::status_request& request);
+    [[nodiscard]] sisl::status_response get_status_metablk(const sisl::status_request& request, std::string type = "");
+    void create_sobject(meta_sub_type type);
 
     sisl::sobject_ptr sobject() { return m_sobject; }
```
yamingk, line 227: we don't need m_sobject and this sobject(), right?

src/engine/meta/meta_blks_mgr.cpp
```diff
@@ -93,7 +94,7 @@ void MetaBlkMgr::start(blk_store_t* sb_blk_store, const sb_blkstore_blob* blob,
 
     auto hs = HomeStoreBase::safe_instance();
     m_sobject = hs->sobject_mgr()->create_object("module", "MetaBlkMgr",
```
yamingk, line 96: we don't need m_sobject right?

src/engine/meta/meta_blks_mgr.cpp
```diff
@@ -399,6 +403,13 @@ void MetaBlkMgr::register_handler(const meta_sub_type type, const meta_blk_found
     HS_LOG(INFO, metablk, "[type={}] registered with do_crc: {}", type, do_crc);
 }
 
+void MetaBlkMgr::create_sobject(meta_sub_type type) {
+    auto sm{HomeStoreBase::safe_instance()->sobject_mgr()};
+    auto meta_blk_obj =
```
yamingk, line 408: don't need to save to meta_blk_obj, right? 
Also we don't need a function here, i think, you can directly call HomeStoreBase::safe_instance()->sobject_mgr()->create_object(...); from line:266.

src/engine/meta/meta_blks_mgr.cpp
```diff
@@ -412,6 +423,7 @@ void MetaBlkMgr::add_sub_sb(const meta_sub_type type, const void* context_data,
 
     // add meta_bid to in-memory for reverse mapping;
     m_sub_info[type].meta_bids.insert(meta_bid.to_integer());
+    create_sobject(type);
```
yamingk, line 426: Have you verified that multiple call to create_object to with same type and name is fine in sisl sobject framework? 
Even though framework is tolerant to that, I don't think it is good to rely on that, reason being if in the future framework changed and won't support that, this code shouldn't need any change, agree?

src/engine/meta/meta_blks_mgr.cpp
yamingk: This is still not changed?

src/engine/meta/meta_blks_mgr.cpp
shosseinimotlagh: right

src/engine/meta/meta_blks_mgr.cpp
shosseinimotlagh: done

src/engine/meta/meta_blks_mgr.cpp
shosseinimotlagh: done

src/engine/meta/meta_blks_mgr.cpp
shosseinimotlagh: Agree. The check is added in `create_sobject`

src/engine/meta/meta_blks_mgr.cpp
```diff
@@ -399,6 +400,14 @@ void MetaBlkMgr::register_handler(const meta_sub_type type, const meta_blk_found
     HS_LOG(INFO, metablk, "[type={}] registered with do_crc: {}", type, do_crc);
 }
 
+void MetaBlkMgr::create_sobject(meta_sub_type type) {
+    if(m_sub_info[type].meta_bids.size()==1) {
+        auto sm{HomeStoreBase::safe_instance()->sobject_mgr()};
```
yamingk, line 405: nitpick: remove sm and do the call directly as `sm` doesn't seem used anywhere else.

src/engine/meta/meta_blks_mgr.cpp
```diff
@@ -399,6 +400,14 @@ void MetaBlkMgr::register_handler(const meta_sub_type type, const meta_blk_found
     HS_LOG(INFO, metablk, "[type={}] registered with do_crc: {}", type, do_crc);
 }
 
+void MetaBlkMgr::create_sobject(meta_sub_type type) {
+    if(m_sub_info[type].meta_bids.size()==1) {
```
yamingk, line 404: Have you run clang-format on your editer? It seems it is not doing format for you, normally it should be:
if (m_sub_info[type].meta_bids.size() = 1) // << with space in between

src/engine/meta/meta_blks_mgr.cpp
```diff
@@ -90,10 +90,9 @@ void MetaBlkMgr::start(blk_store_t* sb_blk_store, const sb_blkstore_blob* blob,
 
     HS_REL_ASSERT_GT(get_page_size(), META_BLK_HDR_MAX_SZ);
     HS_REL_ASSERT_GT(get_page_size(), MAX_BLK_OVF_HDR_MAX_SZ);
-
-    auto hs = HomeStoreBase::safe_instance();
-    m_sobject = hs->sobject_mgr()->create_object("module", "MetaBlkMgr",
-                                                   std::bind(&MetaBlkMgr::get_status, this, std::placeholders::_1));
+    auto sm{HomeStoreBase::safe_instance()->sobject_mgr()};
```
yamingk, line 93: similar comment here as below

yamingk (approved):

szmyd (approved):
lgtm

PR #127, opened 2023-07-13
SDSTOR-11060 : range_put bug due to lack of available space in a node

shosseinimotlagh:
Before update the leaf node, a check is added to make sure there is enough available size for new value

### DIFF
--- a/src/include/homestore/btree/detail/btree_mutate_impl.ipp
+++ b/src/include/homestore/btree/detail/btree_mutate_impl.ipp
@@ -203,7 +203,14 @@ btree_status_t Btree< K, V >::mutate_write_leaf_node(const BtreeNodePtr& my_node
                 BT_DBG_ASSERT(false, "For non-extent keys range-update should be really update and cannot insert");
                 ret = btree_status_t::not_supported;
             } else {
+                const auto new_val_size{(*req.m_newval).serialized_size()};
+                V tmp_v;
                 for (auto idx{start_idx}; idx <= end_idx; ++idx) {
+                    my_node->get_nth_value(idx, &tmp_v, false);
+                    if (my_node->available_size(m_bt_cfg) + tmp_v.serialized_size() <= new_val_size) {
+                        req.set_cursor_key(my_node->get_nth_key< K >(idx, false));
+                        return btree_status_t::has_more;
+                    }
                     my_node->update(idx, *req.m_newval);
                 }
             }
--- a/conanfile.py
+++ b/conanfile.py
@@ -5,7 +5,7 @@ from conans import CMake
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "4.1.1"
+    version = "4.1.2"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/.clang-format
+++ b/src/.clang-format
@@ -18,7 +18,6 @@ AlignOperands:   false
 AlignTrailingComments: true
 AllowShortBlocksOnASingleLine: true
 AllowShortIfStatementsOnASingleLine: true
-AllowShortBlocksOnASingleLine: true
 AllowShortCaseLabelsOnASingleLine: false
 # AllowShortFunctionsOnASingleLine: InlineOnly
 # AllowShortLoopsOnASingleLine: false
--- a/src/tests/test_mem_btree.cpp
+++ b/src/tests/test_mem_btree.cpp
@@ -28,7 +28,7 @@
 
 static constexpr uint32_t g_node_size{4096};
 using namespace homestore;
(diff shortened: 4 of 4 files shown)


yamingk:
Should we target the bug fix (unrelated to fiber) to master branch? The master branch need this fix to unblock other bug fixes (I have one pending)?

src/include/homestore/btree/detail/btree_mutate_impl.ipp
```diff
@@ -203,7 +203,12 @@ btree_status_t Btree< K, V >::mutate_write_leaf_node(const BtreeNodePtr& my_node
                 BT_DBG_ASSERT(false, "For non-extent keys range-update should be really update and cannot insert");
                 ret = btree_status_t::not_supported;
             } else {
+                const auto new_val_size{(*req.m_newval).serialized_size()};
                 for (auto idx{start_idx}; idx <= end_idx; ++idx) {
+                    if (my_node->available_size(m_bt_cfg) <= new_val_size) {
```
yamingk, line 208: I don't quite follow what is this trying to achieve, but result of this line: 208 doesn't seem to rely on the for loop. 
If it is false, it is always false, if it is true, it always do with the 1st idx. 

Also shouldn't the has_more be set to ret and fall through to below logic to handle it? I don't see any caller is handling has_more.

shosseinimotlagh: 1- `update()` changes the value of available_size. Let's assume the current val for entry _`idx`_  has size 1 and the the requested val size = 500. Therefore, it needs to check the requested val size is available or not. (There is a minor bug that I have fixed it to check  extra needed size which I will update this PR). 

2- For the second point this function should return `has_more` then in `put()` it checks the return value and when it is `has_more`, it calls `do_put()` again and in the next call of `do_put()` , `split_nodes()` will occur and the rest of request will perfom.

src/include/homestore/btree/detail/btree_mutate_impl.ipp
```diff
@@ -203,7 +203,12 @@ btree_status_t Btree< K, V >::mutate_write_leaf_node(const BtreeNodePtr& my_node
...
+                        req.set_cursor_key(my_node->get_nth_key< K >(idx, true));
```
yamingk, line 209: Is there a reason to copy the key?

src/include/homestore/btree/detail/btree_mutate_impl.ipp
```diff
@@ -203,7 +203,12 @@ btree_status_t Btree< K, V >::mutate_write_leaf_node(const BtreeNodePtr& my_node
                 BT_DBG_ASSERT(false, "For non-extent keys range-update should be really update and cannot insert");
                 ret = btree_status_t::not_supported;
             } else {
+                const auto new_val_size{(*req.m_newval).serialized_size()};
```
yamingk, line 206: What will happen if start_found or end_found is not true?

shosseinimotlagh: When we are in this point, we are sure that we are in the leaf. Before calling this function the request is getting trimmed `req.trim_working_range( my_node->min_of(` for the leaf. 
For the out of band case, if we point to same start and end without any match, it is hitting unavailable range. In this case, `count = my_node->template get_all< K, V > `returns 0 and `BT_NODE_LOG_ASSERT` would be raised accordingly.

src/include/homestore/btree/detail/btree_mutate_impl.ipp
shosseinimotlagh: confirmed. no need

src/include/homestore/btree/detail/btree_mutate_impl.ipp
yamingk: Can we add below test case with rangeput, the expected result should be RangePut should fail. 
Case-1:  start_key is invalid, end_key is valid.
Case-2: start_key is valid, end_key is invalid.
Case-3: both keys are invalid.

src/include/homestore/btree/detail/btree_mutate_impl.ipp
```diff
@@ -203,7 +203,14 @@ btree_status_t Btree< K, V >::mutate_write_leaf_node(const BtreeNodePtr& my_node
...
+                V tmp_v;
                 for (auto idx{start_idx}; idx <= end_idx; ++idx) {
+                    my_node->get_nth_value(idx, &tmp_v, false);
+                    if (my_node->available_size(m_bt_cfg) + tmp_v.serialized_size() <= new_val_size) {
+                        req.set_cursor_key(my_node->get_nth_key< K >(idx, false));
```
yamingk, line 211: What is the reason we don't update req.working_range here?

PR #129, opened 2023-07-17
Fix Upgrade in LogstoreFamily

shosseinimotlagh:
**Maintaining Compatibility for different versions**
In previous versions of AM image, i.e., AM20, `m_metablk_name` was populated with **LogDevFamily**. In the recent, m_metablk_name had been changed to `LogStoreFamily` which results in broken recovery procedure from older versions. In this PR I revert it back that upgrading from AM -20 to AM21 will be possible.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -2,7 +2,7 @@ from conans import ConanFile, CMake, tools
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "3.6.8"
+    version = "3.6.9"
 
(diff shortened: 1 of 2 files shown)


szmyd:
It may not be possible to point to a version of SISL that IOMgr does not point to in the GitHub CI since the sisl package is built via iomanager:stable/v8.x not a search in artifactory like in Jenkins.

raakella1 (approved):
LG! Please add a write-up explaining why this change is necessary

PR #132, opened 2023-08-08
SDSTOR-11409 : Fix sanitization issue in fiber btree UT

shosseinimotlagh:


src/tests/test_mem_btree.cpp
```diff
@@ -77,6 +77,8 @@ struct BtreeTest : public testing::Test {
     using K = typename TestType::KeyType;
     using V = typename TestType::ValueType;
 
+    virtual ~BtreeTest() override = default;
```
szmyd, line 80: Just `override`; we want there to be an error if the underlying removes the virtual attribute.

src/include/homestore/btree/mem_btree.hpp
```diff
@@ -43,7 +43,7 @@ class MemBtree : public Btree< K, V > {
     btree_status_t write_node_impl(const BtreeNodePtr& node, void* context) { return btree_status_t::success; }
 
     btree_status_t read_node_impl(bnodeid_t id, BtreeNodePtr& node) const override {
-        node = BtreeNodePtr{r_cast< BtreeNode* >(id)};
+        node.reset(r_cast< BtreeNode* >(id));
```
szmyd, line 46: It's not clear to me what issue this change fixed?

src/include/homestore/btree/detail/btree_node.hpp
```diff
@@ -619,8 +619,10 @@ class BtreeNode : public sisl::ObjLifeCounter< BtreeNode > {
 
     friend void intrusive_ptr_release(BtreeNode* node) {
         if (node->m_refcount.decrement_testz(1)) {
+            auto node_buffer = node->m_phys_node_buf;
             node->~BtreeNode();
             delete[] uintptr_cast(node);
+            delete[] node_buffer;
```
szmyd, line 625: It seems odd to deallocate in a different order than allocation. Originally the BtreeNode is allocated, then the node_buf is allocated, constructed and finally the BtreeNode is constructed. Should we not to the same in reverse? :
```
            auto node_buffer = node->m_phys_node_buf;
            node->~BtreeNode();
            delete[] node_buffer;
            delete[] uintptr_cast(node);
```

szmyd: Also, it's not clear that `node_buffer` was allocated with `new[]` anywhere. We should define a specific type for this so someone cannot allocate with `new` or `malloc` and expect this class to work right?

src/include/homestore/btree/mem_btree.hpp
shosseinimotlagh: the way that that has been implemented it causes the reference count for the intrusive pointer to bump up. As a result, in [post_order_traversal](https://raw.githubusercontent.com/eBay/HomeStore/fiber/src/include/homestore/btree/detail/btree_common.ipp#:~:text=read_and_lock_node(m_root_node_info.bnode_id()%2C%20root%2C%20ltype%2C%20ltype%2C%20nullptr)%3B), BtreeNodePtr root has not release the pointer.

src/tests/test_mem_btree.cpp
shosseinimotlagh: I removed it. I did as you suggested, destructor for each test would be needed( which is unnecessary).

src/include/homestore/btree/mem_btree.hpp
szmyd: Thanks for the explanation!

src/include/homestore/btree/detail/btree_node.hpp
shosseinimotlagh: The sequence is 
create_root_node() -> alloc_leaf_node() -> alloc_node() [[node_buffer is created here]](https://raw.githubusercontent.com/eBay/HomeStore/fiber/src/include/homestore/btree/mem_btree.hpp#:~:text=BtreeNodePtr%20alloc_node(bool%20is_leaf)%20override%20%7B), then new_node is created and the allocated [node_buffer is passed](https://raw.githubusercontent.com/eBay/HomeStore/fiber/src/include/homestore/btree/detail/btree_node_mgr.ipp#:~:text=static%20BtreeNode*%20create_node(uint32_t%20node_ctx_size%2C%20Args%26%26...%20args)%20%7B) .

szmyd (approved):

PR #134, opened 2023-08-10
SDSTOR-11060 : range_put bug due to lack of available space in a node

shosseinimotlagh:
Before update the leaf node, a check is required to make sure there is enough available size for new value

### DIFF
--- a/src/tests/test_mem_btree.cpp
+++ b/src/tests/test_mem_btree.cpp
@@ -115,6 +115,13 @@ struct BtreeTest : public testing::Test {
         }
     }
 
+    void range_put(uint32_t start_entry, uint32_t end_entry, bool expected) {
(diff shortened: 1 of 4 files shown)


src/tests/test_mem_btree.cpp
```diff
@@ -424,6 +431,152 @@ TYPED_TEST(BtreeTest, RangeUpdate) {
     this->query_validate(0, num_entries - 1, 75);
 }
 
+// Comment this for now after making the decision how to treat corner cases
```
szmyd, line 434: I'd prefer to simply leave this out and put it in a new pr. Having it like this could grow stale and not even build anymore when someone wants to uncomment it versus a pr that builds and would get attention eariler?

PR #136, opened 2023-08-14
Enable sanitize

shosseinimotlagh:


yamingk:
This PR can still be helpful for local builds to run sanitize by defualt, right?

PR #138, opened 2023-08-16
SDSTOR-11579 ChunkSelector and its default implementation 

JacksonYao287:
i am not quite sure whether it is appropriate to use `HomeStoreStaticConfig` for homeobject to pass the implementation of chunkselector to homeobject. this PR is used to finalize the interface of chunkselector between homeobject and homestore. the implementation of heapChunkSelector will be pushed to HomeObject repo after this pr is merged

### DIFF
--- a/src/include/homestore/homestore_decl.hpp
+++ b/src/include/homestore/homestore_decl.hpp
@@ -29,6 +29,8 @@
 #include <spdlog/fmt/fmt.h>
 #include <nlohmann/json.hpp>
 
+#include <homestore/chunk_selector.h>
+
 //
 // Note:
 // This file should only include stuffs that are needed by both service layer and internal homestore components;
@@ -138,11 +140,16 @@ public:
     io_flag data_open_flags{io_flag::DIRECT_IO}; // All data drives open flags
     io_flag fast_open_flags{io_flag::DIRECT_IO}; // All index drives open flags
 
+
     uint64_t app_mem_size{static_cast< uint64_t >(1024) * static_cast< uint64_t >(1024) *
                           static_cast< uint64_t >(1024)}; // memory available for the app (including cache)
     uint64_t hugepage_size{0};                            // memory available for the hugepage
     bool is_read_only{false};                             // Is read only
     bool auto_recovery{true};                             // Recovery of data is automatic or controlled by the caller
+    ChunkSelector* dataChunkSelector{nullptr};        // ChunkSelector for DataService
+    ChunkSelector* logChunkSelector{nullptr};          // ChunkSelector for LogService
+    ChunkSelector* IndexChunkSelector{nullptr};      // ChunkSelector for IndexService
+    ChunkSelector* metaChunkSelector{nullptr};        // ChunkSelector for MetaService
 
 #ifdef _PRERELEASE
     bool force_reinit{false};
--- a/src/lib/device/virtual_dev.cpp
+++ b/src/lib/device/virtual_dev.cpp
@@ -86,7 +86,7 @@ static std::shared_ptr< BlkAllocator > create_blk_allocator(blk_allocator_type_t
     }
 }
 
-void VirtualDev::init(DeviceManager* mgr, vdev_info_block* vb, uint32_t blk_size, bool auto_recovery,
+void VirtualDev::init(DeviceManager* mgr, ChunkSelector* chunkSelector, vdev_info_block* vb, uint32_t blk_size, bool auto_recovery,
                       vdev_high_watermark_cb_t hwm_cb) {
     m_mgr = mgr;
     m_vb = vb;
@@ -97,15 +97,16 @@ void VirtualDev::init(DeviceManager* mgr, vdev_info_block* vb, uint32_t blk_size
     m_recovery_init = false;
     m_auto_recovery = auto_recovery;
     m_hwm_cb = std::move(hwm_cb);
+    m_chunk_selector = chunkSelector;
 }
 
 /* Create a new virtual dev for these parameters */
-VirtualDev::VirtualDev(DeviceManager* mgr, const char* name, PhysicalDevGroup pdev_group,
+VirtualDev::VirtualDev(DeviceManager* mgr, ChunkSelector* chunkSelector, const char* name, PhysicalDevGroup pdev_group,
                        blk_allocator_type_t allocator_type, uint64_t size_in, uint32_t nmirror, bool is_stripe,
                        uint32_t blk_size, char* context, uint64_t context_size, bool auto_recovery,
                        vdev_high_watermark_cb_t hwm_cb) :
         m_name{name}, m_allocator_type{allocator_type}, m_metrics{name}, m_pdev_group{pdev_group} {
-    init(mgr, nullptr, blk_size, auto_recovery, std::move(hwm_cb));
+    init(mgr, chunkSelector, nullptr, blk_size, auto_recovery, std::move(hwm_cb));
 
     auto const pdev_list = m_mgr->get_devices(pdev_group);
     // Prepare primary chunks in a physical device for future inserts.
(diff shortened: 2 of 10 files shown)


src/include/homestore/homestore_decl.hpp
```diff
@@ -138,11 +140,16 @@ struct hs_input_params {
...
     io_flag fast_open_flags{io_flag::DIRECT_IO}; // All index drives open flags
 
+
     uint64_t app_mem_size{static_cast< uint64_t >(1024) * static_cast< uint64_t >(1024) *
                           static_cast< uint64_t >(1024)}; // memory available for the app (including cache)
     uint64_t hugepage_size{0};                            // memory available for the hugepage
     bool is_read_only{false};                             // Is read only
     bool auto_recovery{true};                             // Recovery of data is automatic or controlled by the caller
+    ChunkSelector* dataChunkSelector{nullptr};        // ChunkSelector for DataService
+    ChunkSelector* logChunkSelector{nullptr};          // ChunkSelector for LogService
```
zichanglai, line 150: I think ChunkSelector will be used by data service only, there is no need to provide a chunkselector for other services.

JacksonYao287: chunkSelector will be a parameter of the constructor of vDev, and every service will have a vDev, so we need to provide a chunkSelector for every service. for dataService , we can pass the actual chunkSelector(heapChunkSelector)  provided by homeObject. for other three service, we could just pass nullptr to ignore chunkSelector.

the default value of all the chunkSelectors here are nullptr.  the code here does not take much modifications while leaving a simple port for other service to implement their own chunkSelector if necessary in the future. it unifies the process of all the services to create a vDev

xiaoxichen (commented):
Instead of using nullptr,  can we have a SimpleChunkSelector , or just use `ChunkSelector` as base class (instead of virtual class) , implementing default behavior in add_chunk/select_chunk.

That makes the code logic more readable

JacksonYao287:
@xiaoxichen 
1 i believe we need a virtual class here, because the derived class will include the member of base class, which may not be useful for the new implementation.
for example, if the base class is a roundRobinChunkSelector, it may have a member of a vector, which will hold all the candidate chunks. but in HeapChunkSelector, a heap will be used to hold all the available chunks. the vector of base class is useless to the derived  HeapChunkSelector.

2 i also considered add a default chunkselector for this. well, will do that.

xiaoxichen:
> for example, if the base class is a roundRobinChunkSelector, it may have a member of a vector, which will hold all the candidate chunks. but in HeapChunkSelector, a heap will be used to hold all the available chunks. the vector of base class is useless to the derived HeapChunkSelector.
> 
Yes that is true, however it seems like we only need L442(https://github.com/eBay/HomeStore/pull/138/files#diff-c65d919e9f45d2ab0417666450cbc1105e3157f91dc7847ef6b6dd253306c71fR442) in the basic(base) implementation

src/lib/device/virtual_dev.cpp
```diff
@@ -434,7 +438,8 @@ BlkAllocStatus VirtualDev::do_alloc_blk(blk_count_t nblks, const blk_alloc_hints
         BlkAllocStatus status = BlkAllocStatus::FAILED;
 
         while (try_streams != 0) {
-            preferred_chunk = reinterpret_cast< PhysicalDevChunk* >(stream_info->chunk_list[stream_info->stream_cur]);
+            if(m_chunk_selector) preferred_chunk = m_chunk_selector->select_chunk();
```
xiaoxichen, line 441: maybe select_chunk need take some argument at least hints.

JacksonYao287:
> Yes that is true, however it seems like we only need L442(https://github.com/eBay/HomeStore/pull/138/files#diff-c65d919e9f45d2ab0417666450cbc1105e3157f91dc7847ef6b6dd253306c71fR442) in the basic(base) implementation

yea， that is the only place the default chunkSelector works. and that is why i did not implement the default chunkSelector in the first version of this PR. but , for a graceful code structure, we should implement a default chunkSelector here. let me take a look how to do it.

src/lib/device/virtual_dev.cpp
zichanglai: actually I do not think this is the usage pattern for HeapChunkSelector.  this function will be called for per alloc blkc request and if select_chunk() is called here, the chunk will be remove from the HeapChunkSelector and these is no chance for the blk alloc request from the shard to allocate blk from this chunk anymore. 

Instead I think select_chunk() will be called from honeobject layer and once chunk is slected, it will be used as a hint to alloc blk to tell it use this chunk as destination to alloc blk

JacksonYao287: > the chunk will be remove from the HeapChunkSelector and these is no chance for the blk alloc request from the shard to allocate blk from this chunk anymore.

when we close a shard, the chunk is not busy, so we should call add_chunk() to return it back to chunkselector . so that this chunk can be reused for the next blk allocate

zichanglai: yes, that is sure. but my point is before shard is sealed/closed,   put shard request will try to allocate a blk from honestore, but now the first put request of this shard, a chunk will be selected and chunk is removed from selector.      let us say that is chunkA. after that,  the next put request of this shard will try to allocate a new block with this function, but it will return another chunk again which is definitely not chunk A.

JacksonYao287: if i don`t missunderstand, put shard request is actually the create shard request, which will put a block as a shard header, right?

if yes , when a shard header is put to the chunk, this means the shard is open and the chunk is in busy state, we could not return the chunk to the selector until the shard is sealed. so next put request will also try to allocate a new block in the same chunk.

zichanglai: no,  put shard mean put blobs to the shard, both put shard and create shard will allocate blk from data service, but create shard will put shard header to the blk, but put shard will put blob content to the blk.  both of them will need to allocate blk.

JacksonYao287: yea, get your point here. you mean we should expose `select_chunk()` to homeobject, not in `allocate_block`, so that homeobject can manage the chunk itself. it will return the chunk to chunkselector until homeobject seal a shard. homeobject will manager blocks itself when allocating a block, not call the `allocateBlk` of homestore , right?

zichanglai: > yea, get your point here. you mean we should expose `select_chunk()` to homeobject, not in `allocate_block`, so that homeobject can manage the chunk itself. it will return the chunk to chunkselector until homeobject seal a shard. 

yes, that is my thought. 

> homeobject will manager blocks itself when allocating a block, not call the `allocateBlk` of homestore , right?
that is not decided yet. I think both work.  it can allocate a block from the selected chunk directly or call allocateBlk but with chunk as the hint params.   

let us hear more thoughts from other workmates.

xiaoxichen: I hope we can still leave the `alloc_blks` in homestore and homeobj still use that.

@zichanglai the shard->chunk affinitiy should be implemented in `HeapChunkSelector::select_chunk(const blk_alloc_hints& hints)` in `HeapChunkSelector'.

The most obvious way is add the `chunk` into alloc_hint , in this case `HeapChunkSelector::select_chunk` always return that chunk in the alloc_hint`.  And `alloc_blks` return chunk info to caller(homeobj).  But this approach is ugly.

Would prefer we add `shardID` , `PGID` into  `alloc_hint`,  `HeapChunkSelector::select_chunk` can maintain a mapping table  to ensure same `shardID` always get same chunk and same `PGID` always return chunks within same device.

zichanglai: >  The most obvious way is add the chunk into alloc_hint , in this case HeapChunkSelector::select_chunk always return that chunk in the alloc_hint.  And alloc_blks` return chunk info to caller(homeobj). But this approach is ugly

if we add chunk to the alloc_hint,  there is no need to call HeapChunkSelector any interface any more. it will just call chunk->blk_allocator_mutable()->alloc() interface to allocate a new blk.

> Would prefer we add shardID , PGID into alloc_hint, HeapChunkSelector::select_chunk can maintain a mapping table to ensure same shardID always get same chunk and same PGID always return chunks within same device.

I am ok with this, but something lilely "shardID" and "PGID" are something homestore doest not know at all, but blk_alloc_hints is a homestore struct, it will be better to called user data area or something else. and for homeobject, the user data area will store shard_id/pg_id.

xiaoxichen: yes that makes sense.

Can we move this discussion to slack to hear more from others?

JacksonYao287:
HeapChunkSelector will be pushed to HomeObject repo

PR #139, opened 2023-08-16
SDSTOR-11578 add user data(private area) support in homestore chunk

zichanglai:
homestore chunk will setup a user data area which allow upper layer(such as homeobject) to store some customize data, these data can be  homestore unawared.

### DIFF
--- a/src/lib/device/physical_dev.hpp
+++ b/src/lib/device/physical_dev.hpp
@@ -91,6 +91,8 @@ struct chunks_block {
 };
 #pragma pack()
 
+static constexpr size_t MAX_USER_DATA_SIZE_IN_CHUNK{32};
+
 #pragma pack(1)
 struct chunk_info_block {
     uint64_t chunk_start_offset{0}; // Start offset of the chunk within a pdev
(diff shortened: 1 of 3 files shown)


src/lib/device/physical_dev.hpp
```diff
@@ -382,6 +386,14 @@ class PhysicalDevChunk {
     void free_slot() { m_chunk_info->set_slot_allocated(false); }
     void update_end_of_chunk(uint64_t sz) { m_chunk_info->end_of_chunk_size = int64_cast(sz); }
     void update_start_offset(uint64_t start_offset) { m_chunk_info->update_start_offset(start_offset); }
+    void set_user_data(uint8_t* user_data, uint8_t user_data_size) {
+        HS_REL_ASSERT_LE(user_data_size, MAX_USER_DATA_SIZE_IN_CHUNK);
+        std::memcpy(m_chunk_info->user_data, user_data, user_data_size);
+        if (user_data_size < MAX_USER_DATA_SIZE_IN_CHUNK) {
```
zhiteng, line 392: Since the situation when 'user_data_size' is equal or larger than MAX_USER_DATA_SIZE_IN_CHUNK is ready asserted in line 389, do we still have to add the branch here?

zichanglai: asset just make sure user_data_size is **less or equal** to MAX_USER_DATA_SIZE_IN_CHUNK, so if user_data_size is less,  line 393 will fill the unfilled tail area to zero.

zichanglai:
chunk private area is already implemented in DeviceManager v2 pr.

zichanglai:
chunk private area is already implemented in DeviceManager v2 pr.

PR #144, opened 2023-08-30
SDSTOR-11237 SDSTOR-11255 AppendBlkAllocator and its checkpointing

yamingk:
Below APIs:
1. async_alloc_write (alloc_blk)
2. async_read (include read verification)
3. async_free (free_blk bookkeeping freeable space for each chunk)
4. cp_flush (two chunks created, with only one dirty chunk and during cp_flush only dirty chunk is flushed)

All existing test binaries passed.


src/lib/device/virtual_dev.cpp
```diff
@@ -502,6 +520,19 @@ void VirtualDev::update_vdev_private(const sisl::blob& private_data) {
...
+VDevCPContext::VDevCPContext(cp_id_t cp_id) : CPContext(cp_id) {}
+
+std::unique_ptr< CPContext > VirtualDev::create_cp_context(cp_id_t cp_id) {
+    return std::make_unique< VDevCPContext >(cp_id);
+}
+
+void VirtualDev::cp_flush(CP*) {
+    // pass down cp so that underlying componnents can get their customized CP context if needed;
+    m_chunk_selector->foreach_chunks([this](cshared< Chunk >& chunk) { chunk->cp_flush(); });
+}
```
zichanglai, line 534: I think it should use m_all_chunks which will contain all chunks of this virtualdev.
if chunkselector is HeapChunkSelector, once a chunk is selected, it will be removed and only contains part of chunks in this virtualdev.

src/lib/blkalloc/append_blk_allocator.h
```diff
@@ -0,0 +1,102 @@
...
+    AppendBlkAllocator& operator=(const AppendBlkAllocator&) = delete;
+    AppendBlkAllocator& operator=(AppendBlkAllocator&&) noexcept = delete;
+    virtual ~AppendBlkAllocator() = default;
+
+    BlkAllocStatus alloc(BlkId& bid) override;
+    BlkAllocStatus alloc(blk_count_t nblks, const blk_alloc_hints& hints, std::vector< BlkId >& out_blkid) override;
+
+    void free(const std::vector< BlkId >& blk_ids) override;
+    void free(const BlkId& b) override;
+
```
zichanglai, line 74: do we need another commit() interface for upper layer to replay the journal log and update last offset during the replay

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,162 @@
...
+BlkAllocStatus AppendBlkAllocator::alloc(blk_count_t nblks, const blk_alloc_hints& hint,
+                                         std::vector< BlkId >& out_bids) {
+    if (available_blks() < nblks) {
+        COUNTER_INCREMENT(m_metrics, num_alloc_failure, 1);
+        LOGERROR("No space left to serve request nblks: {}, available_blks: {}", nblks, available_blks());
+        return BlkAllocStatus::SPACE_FULL;
+    }
+
+    // Push 1 blk to the vector which has all the requested nblks;
+    out_bids.emplace_back(m_last_append_offset, nblks, m_chunk_id);
```
JacksonYao287, line 73: sorry, why here 1 blk has all the requested nblks, but not n blks in the vector?

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,162 @@
...
+void AppendBlkAllocator::clear_dirty_offset() { m_is_dirty = false; }
+
+//
+// free operation does:
+// 1. book keeping "total freeable" space
+// 2. if the blk being freed happens to be last block, move last_append_offset backwards accordingly;
+//
+void AppendBlkAllocator::free(const BlkId& bid) {
+    const auto n = bid.get_nblks();
+    m_freeable_nblks += n;
```
zichanglai, line 131: set_dirty_offset()

src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -31,8 +32,9 @@ BlkDataService::BlkDataService() { m_blk_read_tracker = std::make_unique< BlkRea
 BlkDataService::~BlkDataService() = default;
 
 // first-time boot path
-void BlkDataService::create_vdev(uint64_t size) {
-    const auto phys_page_size = hs()->device_mgr()->optimal_page_size(HSDevType::Data);
+void BlkDataService::create_vdev(uint64_t size, homestore::blk_allocator_type_t alloc_type,
+                                 homestore::chunk_selector_type_t chunk_sel_type) {
+    const auto phys_page_size = hs()->device_mgr()->optimal_page_size(HSDevType::Fast);
```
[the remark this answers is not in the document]
yamingk: interesting, it is carried over from parent branch and didn't seem to be merged correctly after parent branch has fixed this.

src/lib/blkalloc/append_blk_allocator.cpp
yamingk: because it is append only, nblks are always continuous and can be represented by 1 blkid.

src/lib/device/virtual_dev.cpp
yamingk: I thought about it, it depends on how chunkselector is maintaining its chunks. E.g. should block_allocator bypass chunkselector to assume that all chunks should participant checkpionting or always let chunkselector to decide. 
Your point is valid that when a chunk is selected, it is removed from the heap, but one design option is heapchunkselector can still have a view of all the chunks that it maintains. Or we can just remove it from chunkselector and let vdev to add it later, which is also a fine design.

hkadayam: I see there are 2 options broadly.
Option 1: m_all_chunks is maintained in vdev and for CP and other instrumentation, we iterate over m_all_chunks. chunk_selector exposes a method foreach_active_chunk() which = m_all_chunks for round robin selector, but only unallocated chunks for heap selector
This m_all_chunks needs to be thread safe (for vdev supporting dynamic chunk addition), which will do rcu based
Pros: The vdev doesn't need a lamba or additional things and can iterate and functionality sits in vdev.
Cons: A split responsibility between vdev and chunk selector
(OR)
Option 2: vdev does not maintain any list of chunks, but put the responsibility to chunk_selector. chunkselector base class provides rcu based protection (only for dynamic chunk addition cases). It will expose 2 methods foreach_chunk() and foreach_active_chunk()
Pros: Clean interface and everything is well clear and vdev will be handling off of to chunkselector
Cons: We need a lambda to access blk_allocation which is a very very critical code path. This problem is mitigated if we don't use std::function, but just lambda (which compiler optimizes)

src/lib/blkalloc/append_blk_allocator.cpp
hkadayam: Next features we will remove the return of vector of blkids and return a CompositeBlkId where a single instance refers to multiple blks and all vdev calls accept this id for IO which internally runs multiple ios and get that. At present the vector of blkids are given to upper layer which needs to do lots of dance to persist them or serialize them. For Object store (using append blk allocator) this is never a problem because we return a single `BlkId`

src/lib/device/virtual_dev.cpp
zichanglai: yes, I am ok with option 2 too which will make interface more clear and more make sense.   if we are going to do that, we should at least add a function description to foreach_chunk()/foreach_active_chunk() to make the derived class user to know its purpose and difference.

src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -43,14 +45,15 @@ void BlkDataService::create_vdev(uint64_t size) {
                                                         .num_chunks = 1,
                                                         .blk_size = phys_page_size,
                                                         .dev_type = HSDevType::Data,
+                                                        .alloc_type = alloc_type,
+                                                        .chunk_sel_type = chunk_sel_type,
```
hkadayam, line 49: I think blk_alloc_type is good, but suggest chunk selector instance needs to be driven by the callbacks as part of homestore::init(). 

So I was making changes something like this
```
bool HomeStore::start(const hs_input_params& input, hs_before_services_starting_cb_t svcs_starting_cb,
                      std::unique_ptr< IndexServiceCallbacks > index_cbs,
                      std::unique_ptr< ReplServiceCallbacks > repl_cbs)
```
so this callbacks will have a method create_chunk_selector() which will be invoked. Otherwise HomeObject will not have a chance to create its instance with its own constructor right?

src/lib/blkdata_svc/data_svc_cp.hpp
```diff
@@ -0,0 +1,37 @@
...
+    virtual ~DataSvcCPCallbacks() = default;
+
+public:
+    std::unique_ptr< CPContext > on_switchover_cp(CP* cur_cp, CP* new_cp) override;
+    folly::Future< bool > cp_flush(CP* cp) override;
+    void cp_cleanup(CP* cp) override;
+    int cp_progress_percent() override;
+
+private:
+    VirtualDev* m_vdev;
```
hkadayam, line 34: Can this be `shared< VirtualDev >`?

src/lib/blkdata_svc/data_svc_cp.cpp
```diff
@@ -0,0 +1,43 @@
...
+    // iomanager.run_on_forget(hs()->cp_mgr().pick_blocking_io_fiber(), [this, cp]() {
+    auto cp_ctx = s_cast< VDevCPContext* >(cp->context(cp_consumer_t::BLK_DATA_SVC));
+    m_vdev->cp_flush(cp); // this is a blocking io call
+    cp_ctx->complete(true);
+    //});
+
+    return folly::makeFuture< bool >(true);
+}
+
+void DataSvcCPCallbacks::cp_cleanup(CP* cp) {}
```
hkadayam, line 39: I think we may need to make this `no-op` decision at vdev layer and so let this service call vdev->cp_cleanup() which can be a `no-op`. What do you think?

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,180 @@
...
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ * *
+ * *********************************************************************************/
+#include <homestore/meta_service.hpp>
+#include "append_blk_allocator.h"
+
+namespace homestore {
+
+AppendBlkAllocator::AppendBlkAllocator(const BlkAllocConfig& cfg, bool first_time_boot, chunk_num_t id) :
```
hkadayam, line 21: I know we use the chunk_num_t as id, but may be we can change that to uint32_t? You can chose to ignore this comment as well.

src/lib/blkalloc/append_blk_allocator.h
```diff
@@ -0,0 +1,103 @@
...
+#include <sisl/logging/logging.h>
+#include <homestore/blk.h>
+#include "blk_allocator.h"
+#include "common/homestore_assert.hpp"
+#include "common/homestore_config.hpp"
+#include <homestore/superblk_handler.hpp>
+
+namespace homestore {
+static constexpr uint64_t append_blkalloc_sb_magic{0xd0d0d02b};
```
hkadayam, line 26: Some interesting magic here, does it convey anything :-)

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,180 @@
...
+        get_name(),
+        [this](meta_blk* mblk, sisl::byte_view buf, size_t size) { on_meta_blk_found(std::move(buf), (void*)mblk); },
+        nullptr);
+
+    if (first_time_boot) {
+        m_freeable_nblks = available_blks();
+        m_last_append_offset = 0;
+
+        m_sb.create(sizeof(append_blkalloc_sb));
+        m_sb->chunk_id = id;
```
hkadayam, line 34: chunk_id -> allocator_id ??

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+AppendBlkAllocator::AppendBlkAllocator(const BlkAllocConfig& cfg, bool first_time_boot, chunk_num_t id) :
+        BlkAllocator{cfg, id}, m_metrics{get_name().c_str()}, m_sb{get_name()} {
+    // all append_blk_allocator instances use same client type;
+    meta_service().register_handler(
+        get_name(),
+        [this](meta_blk* mblk, sisl::byte_view buf, size_t size) { on_meta_blk_found(std::move(buf), (void*)mblk); },
+        nullptr);
+
+    if (first_time_boot) {
```
hkadayam, line 29: nit: `first_time_boot` is probably not the right name? since chunks can be dynamically created unlike Homestore 3.x :-)

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,180 @@
...
+// VDevCPContext, which contains a list of (ThreadVector) dirty buffers from all the append_blk_allocator instances
+// (protected by RCU);
+//
+// And cp_flush flush every dirty buffer in that list to disk;
+//
+// Option-2. vdev cp_flush will call each individual chunk's blkallocator and each append_blk_allocator instance flush
+// its dirty buffer on demond;
+//
+// We are using option-2.
+//
```
hkadayam, line 113: While for object storage (immutable) data option-1 might be ok, but we need to think through a little bit if we wanted to support append_blk_allocator for other mutable storage use cases, where we might need to think about CPGuard. So suggest to put this as TODO here.

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,180 @@
...
+    if (bid.get_blk_num() + n == m_last_append_offset) {
+        // we are freeing the the last blk id, let's rewind.
+        m_last_append_offset -= n;
+    }
+    set_dirty_offset();
+}
+
+void AppendBlkAllocator::free(const std::vector< BlkId >& blk_ids) {
+    for (const auto b : blk_ids) {
+        free(b);
```
hkadayam, line 158: nit: Is it safe to call this one as this->free() to avoid name clash with malloc/free

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,180 @@
...
+    }
+}
+
+blk_cap_t AppendBlkAllocator::available_blks() const { return get_total_blks() - get_used_blks(); }
+
+blk_cap_t AppendBlkAllocator::get_used_blks() const { return m_last_append_offset; }
+
+bool AppendBlkAllocator::is_blk_alloced(const BlkId& in_bid, bool use_lock) const {
```
hkadayam, line 166: either [[maybe_unused]] for use_lock or better no declaration for use_lock like
```bool AppendBlkAllocator::is_blk_alloced(const BlkId& in_bid, bool) const {```

File: src/lib/blkalloc/append_blk_allocator.cpp (hunk 3 above)
Comment (JacksonYao287) [reply src/lib/blkalloc/append_blk_allocator.cpp:89]: > it is append only, nblks are always continuous and can be represented by 1 blkid.

got it , but it is a little wired here to return a vector but having only one blk in it.

Comment (hkadayam) [reply src/lib/blkalloc/append_blk_allocator.cpp:89]: @JacksonYao287 We will be removing the vector of BlkIds when we address this issue.
https://github.com/eBay/HomeStore/issues/151

File: src/lib/blkdata_svc/data_svc_cp.cpp (hunk 8 above)
Comment (yamingk) [reply src/lib/blkdata_svc/data_svc_cp.cpp:39]: yes, done.

File: src/lib/blkalloc/append_blk_allocator.h (hunk 10 above)
Comment (yamingk) [reply src/lib/blkalloc/append_blk_allocator.h:26]: just don't want to be same as other magic, I think I found similar pattern otherwhere in homestore, maybe logstore magic? :)

File: src/lib/blkalloc/append_blk_allocator.cpp (hunk 14 above)
Comment (yamingk) [reply src/lib/blkalloc/append_blk_allocator.cpp:158]: done.

File: src/lib/blkalloc/append_blk_allocator.cpp (hunk 11 above)
Comment (yamingk) [reply src/lib/blkalloc/append_blk_allocator.cpp:34]: done

File: src/lib/blkalloc/append_blk_allocator.cpp (hunk 3 above)
Comment (yamingk) [reply src/lib/blkalloc/append_blk_allocator.cpp:89]: @JacksonYao287 For NuBlox, when there are not enough continuous blks that can serve the request, blk_allocator needs to go through chunks to collectively get all the blocks and put them in a vector. When there are enough continuous blocks, we only return 1 block_id which result in consumer to only do one write with the returned block id. Consumer can not assume it can always get 1 blkid for nblks that was requested. Hope this explains the doubt.

File: src/lib/blkalloc/append_blk_allocator.cpp (hunk 15 above)
Comment (yamingk) [reply src/lib/blkalloc/append_blk_allocator.cpp:166]: done

File: src/lib/blkdata_svc/data_svc_cp.hpp (hunk 7 above)
Comment (yamingk) [reply src/lib/blkdata_svc/data_svc_cp.hpp]: done.

File: src/lib/blkalloc/append_blk_allocator.cpp (hunk 9 above)
Comment (yamingk) [reply src/lib/blkalloc/append_blk_allocator.cpp:21]: done

File: src/lib/blkalloc/append_blk_allocator.cpp [hunk 16]
```diff
@@ -0,0 +1,180 @@
...
+        nullptr);
+
+    if (first_time_boot) {
+        m_freeable_nblks = available_blks();
+        m_last_append_offset = 0;
+
+        m_sb.create(sizeof(append_blkalloc_sb));
+        m_sb->allocator_id = id;
+        m_sb->last_append_offset = 0;
+        m_sb->freeable_nblks = m_freeable_nblks;
```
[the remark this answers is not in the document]
yamingk: no, sb write only happens during cp flush.

src/lib/blkalloc/append_blk_allocator.cpp
yamingk: done.

src/lib/blkdata_svc/blkdata_service.cpp
yamingk: HO will initialize homestore with `HomeStore::instance()->format_and_start`, which will takes `hs_format_params` as input and in which HO can specify for which service it want which type of allocator/chunkselector, right?

src/lib/blkalloc/append_blk_allocator.h
yamingk: no, when logs are replayed, it will call alloc blk which advances the offset.

src/lib/blkalloc/append_blk_allocator.h
```diff
@@ -0,0 +1,103 @@
...
+static constexpr uint64_t append_blkalloc_sb_version{0x1};
+
+#pragma pack(1)
+struct append_blkalloc_sb {
+    uint64_t magic{append_blkalloc_sb_magic};
+    uint32_t version{append_blkalloc_sb_version};
+    uint64_t allocator_id;
```
JacksonYao287, line 33: is it necesssary to record a id for allocator? what is it used for

src/lib/blkalloc/append_blk_allocator.h
zichanglai: I revisit the code and find commit_blk is the interface of VirtualDev and not the BlkAllocator, yeah, I think use alloc() is enough.

src/lib/device/virtual_dev.cpp
xiaoxichen: @hkadayam  would you please explain more on `Cons: We need a lambda to access blk_allocation which is a very very critical code path. ` ?

JacksonYao287: @hkadayam IMO, the only difference between active chunk and inactive chunk is active chunk will receive allocblk request and inactive chunk not. but they both might have dirty offset, so there is not difference between them for taking checkpoint and i am not sure why we need `foreach_active_chunk()` here

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,180 @@
...
+    }
+}
+
+void AppendBlkAllocator::set_dirty_offset() { m_is_dirty = true; }
+
+void AppendBlkAllocator::clear_dirty_offset() { m_is_dirty = false; }
+
+//
+// free operation does:
+// 1. book keeping "total freeable" space
```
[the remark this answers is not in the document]
yamingk: free can happen on non-tail blob.

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,180 @@
...
+// We are using option-2.
+//
+
+//
+// cp_flush is happening in different thread that is running concurrently as AppendBlkAllocator::alloc;
+//
+// 1. if clear_dirty_offset happens before offset is updated by alloc, but before set_dirty_offset, we are flushing
+// offset twice, which is fine.
+// 2. If clear_dirty_offset happens after set_dirty_offset, we are flushing offset ahead of log
+// store, which is also fine during recovery, it is consistent after log store flush completes;
```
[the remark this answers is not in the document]
yamingk: it is fine for append offset to be ahead, in either case, crash recovery or next cp, we will achieve consistency between the appendblkallocator and logstore.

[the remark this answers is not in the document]
yamingk: next cp will achieve consistency between them. If cp didn't happen before we crash, log replay will alloc blocks and we don't open for ready until all log entries are replayed.

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,180 @@
...
+// 3. Any alloc happened after metablk is persisted will come to next cp;
+//
```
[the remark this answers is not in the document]
yamingk: I am not sure what you mean by partial update, no persistent is being done here until next_cp.

src/lib/blkalloc/append_blk_allocator.h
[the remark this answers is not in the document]
yamingk: allocator is chunk_id which we want to remove chunk concept to allocator as chunk is consumer of this class. It is used for recovery.

src/lib/blkdata_svc/blkdata_service.cpp
yamingk: @JacksonYao287 HO just indicate to HomeStore that dataservice want heapchunkselector (which is a ENUM value), and homestore will initialize its chunkselector type to HeapChunkSelector.

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,180 @@
...
+// Option-2. vdev cp_flush will call each individual chunk's blkallocator and each append_blk_allocator instance flush
+// its dirty buffer on demond;
+//
+// We are using option-2.
+//
+
+//
+// cp_flush is happening in different thread that is running concurrently as AppendBlkAllocator::alloc;
+//
```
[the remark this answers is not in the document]
yamingk: yes. it is nothing different as a bitmap we are doing for varsize blk allocator, only difference is instead of a bitmap, we are persisting a few counter/offsets.

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,180 @@
...
+ * *
+ * *********************************************************************************/
+#include <homestore/meta_service.hpp>
+#include "append_blk_allocator.h"
+
+namespace homestore {
+
+AppendBlkAllocator::AppendBlkAllocator(const BlkAllocConfig& cfg, bool first_time_boot, chunk_num_t id) :
+        BlkAllocator{cfg, id}, m_metrics{get_name().c_str()}, m_sb{get_name()} {
+    // all append_blk_allocator instances use same client type;
```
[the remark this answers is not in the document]
yamingk: I will change it to a TODO, I want to make all appendblkallocator to use the same type to reduce the in-memory footprint for metablk (metablk caches all its consumer's header per their unique type).

src/lib/device/virtual_dev.cpp
yamingk: @xiaoxichen I think the only thing is std::function has a runtime performance cost which foreach_chunk is taking as input parameter, but can be mitigated by lambda.

hkadayam: @JacksonYao287 You are right that `foreach_active_chunk()` is needed only in case we need to do something specific to active chunks and that there are no use case right now. The checkpoint will have a CPContext which will have list of dirtied chunks and persist them irrespective of whether they are active or not.

@xiaoxichen My point was std::function does dynamic memory allocation and make sense only for storing the function pointers in an async call. So we can use plain lambda with (auto) instead of std::function and then the cost is just another function call.

xiaoxichen: Yes but I am not very understanding why this is in the critical code path?   For IO we will not go through any of the for-each right?  ChunkSelector select a chunk and block allocation is made on that chunk only

src/lib/blkalloc/append_blk_allocator.cpp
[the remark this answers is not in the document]
hkadayam: @yamingk I think we can perhaps consider surrounding this with a CPGuard if it is not already done. That way CP will not proceed without exiting this block.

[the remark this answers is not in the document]
yamingk: @hkadayam I probably didn't get the question, it is during cp_flush state and it is a sync call, why do we want to hold a CPGuard? next_cp won't start until CPManager::on_start_flush completes, right?

src/lib/blkalloc/append_blk_allocator.h
```diff
@@ -0,0 +1,103 @@
...
+    bool is_blk_alloced(const BlkId& in_bid, bool use_lock = false) const override;
+    std::string to_string() const override;
+
+    void set_dirty_offset();
+
+    /// @brief : clear dirty is best effort;
+    /// offset flush is idempotent;
+    void clear_dirty_offset();
+
+    void cp_flush() override;
```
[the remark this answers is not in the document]
JacksonYao287: does the journal here mean the raft log from upper layer?

src/lib/device/virtual_dev.cpp
hkadayam: Yes I take that statement back that `alloc_blks()need to iterate the list of chunks`. It does iterate when the chunk selector returns a chunk which doesn't have much space or allocation failed, which is a rarer scenario.

My only concern is that we don't want to walk through all chunks on block allocation time. We do that for get_available_blks() and get_used_blks() which I am hoping is done rarer.

src/lib/blkalloc/append_blk_allocator.cpp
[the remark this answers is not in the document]
hkadayam: @yamingk Yes, but while cp is being flushed, the last_append_offset & freeable_nblks could be in a middle of the update and so lets say the io thread updated last_append_offset and not freeable_nblks and in parallel cp is flushing only last_append_offset, but not freeable_nblks there could be inconsistency right?

That is why I was suggesting to put this entire thing in a CPGuard which returns a CPContext pointer and can attach this dirty information into CPContext*. Now it should be safe from concurrent access.

One more benefit is this CPContext* can keep track of all dirty chunks and hence we don't need to rely on HO to persist this or look for active_chunks.

src/lib/blkalloc/append_blk_allocator.h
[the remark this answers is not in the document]
hkadayam: raft is one of the consumer of journal, but other use cases can use, which is the `LogStoreService`

src/lib/blkdata_svc/blkdata_service.cpp
JacksonYao287: @yamingk  the question is that we do not have  the implemention of HeapChunkSelector inside homestore, it is implement as a derived class of `ChunkSelector` in Homeobject, so Homestore can not get the defination of HeapChunkSelector for itself. that why i want to clarify how to pass HeapChunkSelector from homeobject to homestore

src/lib/device/virtual_dev.cpp
xiaoxichen: We have nblks passed in to chunk selector
`virtual Chunk* select(blk_count_t nblks, const blk_alloc_hints& hints)` which should gareentee/best effort?  the returned chunk has sufficient blks, or , at least the chunk with most allocable blks if none of the chunk has `avail() > nblks`
so yes we should very rare(maybe never) use `foreach_chunks` in allocation path.

yamingk: When allocation fails on a particular shard, HO should seal that shard and create another shard which goes to chunkselector and picks the top chunk in heap. If later blk allocation also fails, it repeat this step until we have picked all the chunks and emptied the heap, in this rare case, I would say GC would be kicked in sometime ealier by some watermark/watchdog. 

in either cases, I don't see a need to call foreach_chunk in either blk allocator or shard creation.

src/lib/blkalloc/append_blk_allocator.cpp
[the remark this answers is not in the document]
yamingk: @hkadayam Point taken, although I am thinking the other way around: every time a counter/offset are being dirtied (alloc/free), push the dirty buffer to my CPContext. During cp_flush, we just flush whatever was there in my dirty list and switch my next_cp pointing to my next dirty list (RCU). 

My original thought is try not to do the Option-1 since at that time we only have one atomic counter and we don't have race issue with IO thread.  Now that we have freeable_nblks, it turns into a buffer now which gives us no choice but to switch to use Option-1.

src/lib/device/virtual_dev.cpp
xiaoxichen: Yes we dont need to call foreach for sure.
Through a few nit on the previous statement:
>When allocation fails on a particular shard, HO should seal that shard and create another shard which goes to chunkselector and picks the top chunk in heap. 
HO will report IO error and DM decide to seal,  or HO seal and notify DM . In both case HO should not be the one to create new shard as the `putBlob(shardID,  blobData)` has shardID in the signature. 


Anyway this discussion is clear enough,  as far as it is not in the critical path,  do we see we need to change it to function ptr (lambda) instead of std::function?

src/lib/blkalloc/append_blk_allocator.cpp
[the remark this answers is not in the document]
xiaoxichen: not sure if I understand correctly , when someone call the alloc(),   we alloc a blk , update these two counters to (N+1, K -1) , and before we even returning the blk to caller,  the CP comes and flush the counter. 

Questsions:
1.  who generate the log and when the log was generated?  I was thinking the caller will log after it gets the blk and finish some business logic?
2.  If we crash at this point and during recovery,  we got updated counter (N+1, K-1) , even we have the log and replay , arent we make it (N+2, K-2)?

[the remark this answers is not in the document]
yamingk: We are going to use option-1 since now it is not just an dirty atomic counter, but a buffer (can't garunteen atomicity while racing with io thread). But if it were still an automic counter during recovery/reply, we will need to do something to make sure the reply achieves final consistency, e.g. compare the alloc requests with existing persisted last_append_offset and if head, no-op, etc. But this is based on we were on option-2. 

Yes, your are right Log was generated by HO and converted by ReplDeva as was indicated by RaftBasedDesign Doc.

src/lib/blkalloc/append_blk_allocator.cpp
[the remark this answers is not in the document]
hkadayam: We don't need to create ThreadVector for this right? CP already provides a safe mechanism for you to protect for race against flushing vs updating (with CPContext).

Now race between concurrent updates of `last_append_offset` and `freeable_nblks` I don't think you need anything other than just atomics like what you have. Nobody is going to read last_append_offset and freeable_nblks together to make any decision yet.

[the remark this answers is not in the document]
yamingk: We need to discuss whether we can tolerant the race. I am thinking about recovery (log replay) we need to make sure the consistency of these two.

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,180 @@
...
+    m_last_append_offset = m_sb->last_append_offset;
+    m_freeable_nblks = m_sb->freeable_nblks;
+
+    HS_REL_ASSERT_EQ(m_sb->magic, append_blkalloc_sb_magic, "Invalid AppendBlkAlloc metablk, magic mismatch");
+    HS_REL_ASSERT_EQ(m_sb->version, append_blkalloc_sb_version, "Invalid version of AppendBlkAllocator metablk");
+}
+
+// alloc a single block;
+BlkAllocStatus AppendBlkAllocator::alloc(BlkId& bid) {
+    if (available_blks() < 1) {
```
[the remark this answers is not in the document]
zichanglai: it is possible for upper layer to call AppendBlkAllocator::alloc() at the same time? if yes, it seems it is not safely enough.  for example: there is only one last blk is left on this chunk and both two requests try to allocate one blk from this chunk and seems both requests will be successfully at last, but that is not expected.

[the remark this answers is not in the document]
yamingk: The assumption for append blk allocator is only one writer per each shard/chunk. 
We can choose to support multiple writers for appendblkallocator (if so, we should not be using append allocator, IMO), we can revisit this design and make required changes. Or do you see a need from your side we need to support multiple writers?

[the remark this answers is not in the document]
zichanglai: there will be the case that s3 GW will send multi put blobs(from different objects ) request in one shard to homeobject, then homeobject will use call homestore data service alloc_blk() in parallel to alloc blk for writing. I think the case may happen, but we can avoid calling homestore data service alloc_blk() in parallel for a chunk by using lock in homeobject layer.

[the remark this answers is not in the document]
yamingk: one chunk one writer, I don't think HO will violate this rule.

src/include/homestore/homestore_decl.hpp
```diff
@@ -89,9 +89,11 @@ ENUM(io_flag, uint8_t,
      DIRECT_IO,   // recommended mode
      READ_ONLY    // Read-only mode for post-mortem checks
 );
-ENUM(blk_allocator_type_t, uint8_t, none, fixed, varsize);
+ENUM(blk_allocator_type_t, uint8_t, none, fixed, varsize, append);
 ENUM(chunk_selector_type_t, uint8_t, // What are the options to select chunk to allocate a block
+     NONE,                           // Caller want nothing to be set
      ROUND_ROBIN,                    // Pick round robin
+     HEAP,                           // Heap chunk selector
```
hkadayam, line 96: FYI: In next PR, we will remove this HEAP, since Homestore is not aware of how upper layer is building their chunk selector.

src/lib/blkalloc/append_blk_allocator.cpp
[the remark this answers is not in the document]
hkadayam: @zichanglai @yamingk Just wanted to make sure we are in same page: HO will not call alloc_blk or write directly. It is done by ReplDev. 

With that we have 2 choices:
1. Let ReplDev take a lock before every alloc_blk and surrounding async write
2. Handle this in AppendBlkAllocator and let it return safe blkids to allow concurrent allocation. But the subsequent writes are not locked and if it so happens that the writes come in out-of-order, so be it. Moreover, we might be doing buffered IO in HDD case, so its really not a problem.

I prefer Option: 2, because 
a) ReplDev is a common layer and not append_blk_allocator specific and we don't want to introduce logic.
b) We never take lock surrounding IO operations. This is all through the homestore and design goal/requirement.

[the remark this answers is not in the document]
hkadayam: BTW: I am not suggesting to make this change as part of this PR, but please open up an issue to track this.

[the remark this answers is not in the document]
yamingk: ReplDev doesn't need to take lock surrounding async_write, concurrent write can still happen if they are not on same shard. 
If write arrives on same shard with multiple blobs, ReplDev only needs to make sure that those async_write on *same* shard happens in same thread (next async_write doesn't need to wait for previous async-write to complete). Again, no lock is needed.

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -59,9 +65,10 @@ BlkAllocStatus AppendBlkAllocator::alloc(BlkId& bid) {
     }
 
     bid.set(m_last_append_offset, 1, m_chunk_id);
+
+    [[maybe_unused]] auto cur_cp = hs()->cp_mgr().cp_guard();
```
hkadayam, line 69: You may need to be move the m_last_append_offset and m_freeable_blks to the current CPContext and change the CPContext when you do `on_switchover_cp`. Current change might not protect from the race between cp modification and flush.

src/lib/blkalloc/append_blk_allocator.cpp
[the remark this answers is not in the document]
hkadayam: @yamingk ReplDev will not know anything about the shard, it will not ensure any per shard ordering. With the current flow after all the discussions, where HO doesn't participate in any write, ReplDev will call alloc_blks() to append_blk_allocator. Unless it takes a global lock, it is impossible to make it concurrent on each shards.

So the requirement to AppendBlkAllocator or any BlkAllocator for that matter is its allocation needs to be thread safe. I believe we can still achieve this atomics, but that is implementation details. 

I understand that this requirement has changed, since we discussed AppendBlkAllocator, so we may need to open an issue and resolve it in next PR.

src/lib/blkalloc/append_blk_allocator.cpp
yamingk: Changing CPContext is already done `on_swtichover_cp`. 
That (moving offset/freeable_nblks to CPContext) is something I was saying yesterday, if we move to CPContext, we will need ThreadVector/RCU for different blk_allocator.  
But why, isn't this good enough for us make sure either all of them are in current cp or next cp. Nothing will be in between?

src/lib/blkalloc/append_blk_allocator.cpp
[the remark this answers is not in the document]
yamingk: it is okay to change the requirement. 
Maybe I am not very clear on the IO threading model between HO and ReplDev. I thought HO will handle all the incoming I/Os and make sure I/Os sending to ReplDev in each thread won't clash (e.g. write on same shard will arrive on same thread), but I could be wrong.

src/lib/blkalloc/append_blk_allocator.cpp
hkadayam: We don't need threadvector/RCU because CPContext* on CP is already an RCU. You need to assign all your members (need a containing structure) (offset/freeable_nblks) into the CPContext* inside the CP. Now you assign this structure to CPContext*. When you switchover CP, you copy these containing structure and attach to next CP.

I put in my suggestion here:

```
class AppendBlkAllocator {

struct AllocContext {
   offset;
   freeable_nblks;
};

on_switchover_cp(old_cp, new_cp) {
      new_cp->cp_ctx->alloc_context = make_unique< AllocContext >();
      // copy offset, freeable_nblks from old_cp
      *new_cp->cp_ctx->alloc_context  = *old_cp->context->alloc_context;
}

alloc_blk() {
    cp = CPGuard();
    cp->cp_ctx->alloc_context->...(all current logic)
}

flush_cp() {
   // Take data from cp->cp_ctx and flush its details.
}
};
```

src/lib/blkalloc/append_blk_allocator.cpp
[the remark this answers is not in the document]
zichanglai: @hkadayam yes, I take the statement of HO will call alloc_blk() back, HO will not call alloc_blk() directly and ReplDev will do that. ReplDev is pg level and not aware of shard at all, unless holding a global lock, it is possible that concurrent writes will happen on the same shard, it will also happens on different shards too.
@yamingk I think we can try use CAS to achieve thread-safe alloc, for example:
std::atomic<uint64_t> last_offset;

BlkID alloc_blk(hints) {
       uint64_t now_offset = last_offset.load();
       do  {
            if ((get_total_blks() - now_offset) < 1) {
                  xxxxx 
                  return failed;
            }
            
            if (!last_offset.compare_exchange_strong(now_offset, now_offset + 1)) {
                  //race condition happens and now_offset is already be updated to the actually stored value in last_offset
                 continue;
            }

            //allocation is succeeded now;
            out_bids.emplace_back(now_offset, nblks, m_chunk_id);
           m_freeable_nblks -= nblks;
           set_dirty_flag();
           return. xxxx;

      } while (true);
}

but yes, even allocation is thread-safe and there is no guarantee that it will always be sequence IO while writing on disk. but I think it is already impossible for us to achieve that because many shards  on same disk are being writing concurrent.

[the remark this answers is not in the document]
yamingk: thread-safe and sequential I/O are likely exclusive, you need to give up one for the other. 
Thread-safe makes perfect sense for appendblkallocator to work with nvme drive. 
For HDD, I am trying to keep that option open (sequencial I/O will achieve better performance) open and leave consumer to decide. It is our implementation that choose to do concurrent writes on same shard/chunk, which is fine.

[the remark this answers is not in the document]
yamingk: closing this conversion with thread_safe mode added.

src/include/homestore/blkdata_service.hpp
```diff
@@ -131,6 +132,11 @@ class BlkDataService {
      */
     BlkReadTracker* read_blk_tracker() { return m_blk_read_tracker.get(); }
 
+    /**
+     * @brief : start data service;
+     */
+    void start();
```
szmyd, line 138: Is it intentionally meant for something like HomeObject to call this? If not, let's hide it somewhere internally yeah?

src/lib/blkalloc/append_blk_allocator.h
```diff
@@ -97,8 +102,8 @@ class AppendBlkAllocator : public BlkAllocator {
     void on_meta_blk_found(const sisl::byte_view& buf, void* meta_cookie);
 
 private:
+    std::mutex m_mtx; // thread_safe, TODO: open option for consumer to choose to go lockless;
```
xiaoxichen: folly::RWSpinLock?

xiaoxichen:
LGTM,  @yamingk could you please do a rebase and cleanup/squash those temporary commits?

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -0,0 +1,192 @@
...
+        // consumer(vdev) already handles this case.
+        COUNTER_INCREMENT(m_metrics, num_alloc_failure, 1);
+        LOGERROR("Can't serve request nblks: {} larger than max_blks_in_op: {}", nblks, BlkId::max_blks_in_op());
+        return BlkAllocStatus::FAILED;
+    }
+
+    // Push 1 blk to the vector which has all the requested nblks;
+    out_bids.emplace_back(m_last_append_offset, nblks, m_chunk_id);
+
+    [[maybe_unused]] auto cur_cp = hs()->cp_mgr().cp_guard();
```
hkadayam, line 110: Nit. You can remove the [[maybe_unused]] since we are surely using that now.

hkadayam (approved):

src/include/homestore/blkdata_service.hpp
yamingk: right, it can be a private member function, will move it with next PR.

src/lib/blkalloc/append_blk_allocator.h
yamingk: should be fine for either of them since they are all in-mem ops.

yamingk:
> | LGTM, @yamingk could you please do a rebase and cleanup/squash those temporary commits? 
@xiaoxichen yes, all homestore PR merges are defaulted (un-squashed merge I think is disabled IIRC) to squash merge.

src/lib/blkalloc/append_blk_allocator.h
yamingk: I would say to keep with mutex to give other threads some cycles to run to achieve better throughput. 
It is only a few cpu cycles anyway within the scope of this mutex. We are concurrently writing the same chunks so I/O latency would be the overwhelming factor of the total latency.

PR #157, opened 2023-09-05
SDSTOR-11579. expose chunk and chunkSelector Header

JacksonYao287:


xiaoxichen (commented):
lgtm

src/include/homestore/chunk.h
```diff
@@ -13,10 +13,14 @@
...
+
+#include<homestore/homestore_decl.hpp>
+#include<homestore/blk.h>
+
 
 namespace homestore {
 class BlkAllocator;
+class PhysicalDev;
 
 class Chunk {
```
szmyd, line 25: I think we should expose a different structure that hides all of this; we don't need HomeObject to depend on the `sizeof(Chunk);` with all these dependencies if possible.

Can we hide this using a PIMPL so we can also change the implementation without worrying about the upper layers?

src/include/homestore/homestore_decl.hpp
```diff
@@ -211,6 +212,36 @@ struct cap_attrs {
...
+    bool is_allocated() const;
+    void set_allocated();
+    void set_free();
+
+    void set_selector_private(const sisl::blob& data);
+    void set_user_private(const sisl::blob& data);
+    void compute_checksum();
+};
+#pragma pack()
+
```
szmyd, line 244: Again, no layer above HomeStore needs to see these fields and structure (I hope); is this all needed by the HeapChunkSelector? If not can we prune it similarly or remove it entirely from the API?

szmyd (dismissed):

src/include/homestore/chunk.h
JacksonYao287: yes, initially , i have considered using a pimpl and it is not graceful to expose all details and useless functions, but there are three interface might be used by upper layer

`available_blks()` :  this might be used by heapChunkSelector to sort the heap.

`user_private()` and `set_user_private()` : this might be used by upper layer to write and read private data, which make no sense to homestore.

another point is HeapChunkSelector should make sure all the shards of a PG should reside in the same pdev, so we might keep a heap for each of pdev, which might need `PhysicalDev*`

src/include/homestore/homestore_decl.hpp
JacksonYao287: yea, i do not want is here too. as I mentioned above , i have to expose "class Chunk", and  it has "chunk_info m_chunk_info", so it is necessary to sizeof(chunk_info) and we have to expose it.

a solution is to use `shared_ptr<chunk_info>` instead of `chunk_info`, and this need more modification in homestore to change all the reference to this. if you agree with this idea, i will do this change

src/include/homestore/chunk.h
xiaoxichen: Yes it had back and forth for several times.

The challenge is the caller of ChunkSelector need `chunk` to proceed , so we either need to return a `chunk` from Selector, or do some static_cast in the caller.
The former one expose chunk to external and as the `const chunk_info& info() const;` expose chunk_info then we need to expose `chunk_info` .
The latter one seems more worse IMHO.

src/include/homestore/homestore_decl.hpp
hkadayam: We should certainly not put in homestore_decl.hpp. I don't think we need to change `chunk_info` to `std::shared_ptr< chunk_info >`, if you can define Chunk itself as a abstract class (say AbstractChunk) and then Chunk could be derived and can have chunk_info() native to Chunk and inside device/ folder

src/lib/device/chunk_info.cpp
```diff
@@ -0,0 +1,36 @@
+/*********************************************************************************
```
hkadayam, line 1: Is there a reason we didn't do this inside chunk_info.h itself and needed to define another .cpp file?

src/include/homestore/chunk.h
hkadayam: Why `available_blks()`, `user_private()` cannot be hidden in Pimpl ?

src/include/homestore/homestore_decl.hpp
xiaoxichen: Yes then the `ChunkSelector` can only return `AbstractChunk`,  the caller need to static cast that to `Chunk` as it is from base to derived class

src/include/homestore/chunk.h
JacksonYao287: we need to expose this to upper layer , so we can not hide in pimpl

src/include/homestore/homestore_decl.hpp
JacksonYao287:
```
class PhysicalDev;
class baseChunk {
public:
    baseChunk() = defalut;
    virtual ~baseChunk() = default;

    virtual void set_user_private(const sisl::blob& data);
    virtual uint8_t* get_user_private();
    virtual blk_cap_t available_blks();
    virtual PhysicalDev* getPhysicalDev();
}
```

```
class ChunkSelector {
public:
    ChunkSelector() = default;
    virtual void add_chunk(cshared< baseChunk >& chunk) = 0;
    virtual void foreach_chunks(std::function< void(cshared< baseChunk >&) >&& cb) = 0;
    virtual baseChunk* select(blk_count_t nblks, const blk_alloc_hints& hints) = 0;

    virtual ~ChunkSelector() = default;
};
```

i have considered exposing the two classes in the header file earlier. the only problem is that , as @xiaoxichen said , we have to do static_cast from a base class to derived class when calling `select` in vdev, which seems not graceful.

but seems this will make the exposed header more simple. 

@hkadayam @xiaoxichen are you ok to expose the two classes in the header file , and do the static_cast inside vdev after calling `chunkselector#selector`

src/include/homestore/chunk.h
xiaoxichen: @szmyd @hkadayam  see if the current implementation with pimpl looks good.
`getInternalChunk` is used to expose Chunk for HS internal

src/include/homestore/vchunk.h
```diff
@@ -0,0 +1,38 @@
...
+#include <sisl/fds/buffer.hpp>
+#include <homestore/blk.h>
+#include <homestore/homestore_decl.hpp>
+
+namespace homestore {
+class Chunk;
+
+class VChunk {
+public:
+    VChunk(cshared< Chunk >&);
```
szmyd, line 26: We probably aren't going to mutate the pointer passed in to the caller on their behalf correct? So this should be either:
`cshared<Chunk> const&` or `cshared<Chunk>&&`
that way someone can do something like:
```
... VChunk(std::make_shared<Chunk>(...)) ... ;
```
requiring the pointer to be mutable means it needs to be a non-temporary on the callers stack (for no reason that I can see).

src/include/homestore/chunk_selector.h
```diff
@@ -0,0 +1,30 @@
+/*********************************************************************************
+ * Modifications Copyright 2017-2019 eBay Inc.
```
szmyd, line 2: New files can have a new copyright date (2023)

src/include/homestore/vchunk.h
szmyd: alson, this should be `explicit` right?

xiaoxichen: yes you are right, `const &` probabaly better than rvalue reference as it can still accept left value.

src/include/homestore/chunk_selector.h
JacksonYao287: got it!

src/include/homestore/vchunk.h
JacksonYao287: > also, this should be `explicit` right?

it think it is better to keep this implicity for convenience , see https://github.com/eBay/HomeStore/blob/8065a05407c0715609a6405d36715d9c65a61c82/src/lib/device/virtual_dev.cpp#L102


other comments makes sense, fixed

xiaoxichen (approved):
lgtm

src/include/homestore/vchunk.h
```diff
@@ -0,0 +1,39 @@
...
+    ~VChunk() = default;
+
+    void set_user_private(const sisl::blob& data);
+    uint8_t* get_user_private();
+    blk_cap_t available_blks();
+    uint32_t getPdevID();
+    cshared< Chunk > getInternalChunk();
+
+private:
+    cshared< Chunk > internalChunk;
```
hkadayam, line 37: Why this class can't be a abstract class instead of container class, where we store internalChunk? With this change, it would mean everytime there is a public interface change to Chunk, we need to implement and do forwarding. Also IMO it would look much more intutive to have them as abstract and impl class.

src/lib/device/round_robin_chunk_selector.cpp
```diff
@@ -12,23 +12,22 @@
...
                       "Dynamically adding chunk to chunkselector is not supported, need RCU to make it thread safe");
 }
 
-void RoundRobinChunkSelector::add_chunk(cshared< Chunk >& chunk) { m_chunks.push_back(chunk); }
+void RoundRobinChunkSelector::add_chunk(VChunk chunk) { m_chunks.push_back(chunk); }
 
 Chunk* RoundRobinChunkSelector::select(blk_count_t, const blk_alloc_hints&) {
     if (*m_next_chunk_index >= m_chunks.size()) { *m_next_chunk_index = 0; }
-    return m_chunks[(*m_next_chunk_index)++].get();
+    return m_chunks[(*m_next_chunk_index)++].getInternalChunk().get();
```
hkadayam, line 27: Nit: But within homestore, can we do existing case style, like get_internal_chunk() to be consistent. On new repo (say HomeObject) it doesn't matter.

src/lib/device/vchunk.cpp
```diff
@@ -0,0 +1,39 @@
...
+ * CONDITIONS OF ANY KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations under the License.
+ *
+ *********************************************************************************/
+#include <homestore/vchunk.h>
+#include "blkalloc/blk_allocator.h"
+#include "device/chunk.h"
+
+namespace homestore {
+    VChunk::VChunk(cshared< Chunk >& chunk) : internalChunk(chunk){}
```
hkadayam, line 20: Like my comment above, this entire class can be an abstract class and let `Chunk` derive from this "VChunk" and implement them?

src/include/homestore/chunk_selector.h
```diff
@@ -0,0 +1,30 @@
...
+#include <homestore/vchunk.h>
+
+namespace homestore {
+class Chunk;
+class ChunkSelector {
+public:
+    ChunkSelector() = default;
+    virtual void add_chunk(VChunk chunk) = 0;
+    virtual void foreach_chunks(std::function< void(VChunk&) >&& cb) = 0;
+    virtual Chunk* select(blk_count_t nblks, const blk_alloc_hints& hints) = 0;
```
yamingk, line 26: Is there a reason to return a raw pointer here? Can we make it a unique ptr? From API signature, it means the caller of select now owns this chunk and when adding back the chunk it also can be a unique ptr, saying, hey, the caller is now passing ownership back to ChunkSelector.

yamingk, line 26: I probably didn't get the reason why add_chunk is takign VChunk, but select is returning Chunk*. 
Can `select` return VChunk by moving `Chunk` derived from `VChunk`?

xiaoxichen: it could be shared_ptr but cannot be unique_ptr,  the selected chunk will not be removed from ChunkSelector as the `for_each` still expect to touch every chunk , regarless if they are in-use (remember the dicsussion to have for_each_active ).

src/lib/device/round_robin_chunk_selector.cpp
```diff
@@ -12,23 +12,22 @@
...
 }
 
-void RoundRobinChunkSelector::foreach_chunks(std::function< void(cshared< Chunk >&) >&& cb) {
+void RoundRobinChunkSelector::foreach_chunks(std::function< void(VChunk&) >&& cb) {
```
yamingk, line 30: is there a reason why it can not be a shared_ptr?

src/lib/device/round_robin_chunk_selector.h
```diff
@@ -43,14 +35,12 @@ class RoundRobinChunkSelector : public ChunkSelector {
...
-
+    void add_chunk(VChunk chunk) override;
     Chunk* select(blk_count_t nblks, const blk_alloc_hints& hints) override;
-
-    void foreach_chunks(std::function< void(cshared< Chunk >&) >&& cb) override;
+    void foreach_chunks(std::function< void(VChunk&) >&& cb) override;
 
 private:
-    std::vector< shared< Chunk > > m_chunks;
+    std::vector< VChunk > m_chunks;
```
yamingk, line 43: If the decision is to make VChunk a base/abstract class, it needs to be vector of shared_ptr of VChunk (you probably are already aware of this).

szmyd: A `shared_ptr<VChunk>` does not make any sense; it's simply a shared_ptr itself; it's move/copy friendly so it should never be wrapped in a pointer and passed by reference or move semantics.

yamingk: @szmyd you misunderstood my comment, which is based on VChunk itself doesn't contain a cshared <Chunk>, but have Chunk derived from VChunk. VChunk will become an abstract class, and m_chunks will be pointing to the base class and use dynamic_cast in run_time to get its real type. 

I am not suggesting we should do this way, that why the comment started with an "if"

src/include/homestore/vchunk.h
```diff
@@ -0,0 +1,39 @@
...
+class VChunk {
+public:
+    VChunk(cshared< Chunk > const&);
+    
+    ~VChunk() = default;
+
+    void set_user_private(const sisl::blob& data);
+    uint8_t* get_user_private();
+    blk_cap_t available_blks();
+    uint32_t getPdevID();
```
hkadayam, line 33: Both these functions  can be const right?

src/lib/device/round_robin_chunk_selector.cpp
JacksonYao287: sure, will change the function name style

src/include/homestore/chunk_selector.h
JacksonYao287: i just copy the current definition of chunk_seletcor, it is defined in the current.
actually, we can not return a unique_prt, since the since chunk_selector will also keep the pointer to it.

we can return  share_ptr if necessary

src/include/homestore/vchunk.h
JacksonYao287: yea, we can add a abstract here, but the result is that we have to do dynamic cast inside homestore, that is my point

src/include/homestore/chunk_selector.h
```diff
@@ -0,0 +1,30 @@
...
+#pragma once
+
+#include <homestore/vchunk.h>
+
+namespace homestore {
+class Chunk;
+class ChunkSelector {
+public:
+    ChunkSelector() = default;
+    virtual void add_chunk(VChunk chunk) = 0;
```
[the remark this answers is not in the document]
JacksonYao287: well, it is defined in the current homestore. but yes, we can keep them the same style

src/include/homestore/vchunk.h
JacksonYao287: yea, will change them to const

src/lib/device/round_robin_chunk_selector.cpp
JacksonYao287: Vchunk has a shared_ptr in itself, so we do not need a shard_ptr of shared_ptr

src/include/homestore/chunk_selector.h
JacksonYao287: the draw back of deriving is we have to do dynamic_cast inside homestore, that is not recommended.
Vchunk is actually a wrapper class of cshared<Chunk>.

src/lib/device/vchunk.cpp
```diff
@@ -19,21 +19,21 @@
 namespace homestore {
     VChunk::VChunk(cshared< Chunk >& chunk) : internalChunk(chunk){}
 
-    void VChunk::set_user_private(const sisl::blob& data){
+    void VChunk::set_user_private(const sisl::blob& data) const{
```
xiaoxichen, line 22: It is too bad to mark this function as const...

src/lib/device/vchunk.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+namespace homestore {
+    VChunk::VChunk(cshared< Chunk >& chunk) : internalChunk(chunk){}
+
+    void VChunk::set_user_private(const sisl::blob& data) const{
+        internalChunk->set_user_private(data);
+    }
+
+    uint8_t* VChunk::get_user_private() const {
+        return internalChunk->user_private();
```
xiaoxichen, line 27: uint8_t* user_private()

user_private was used to get raw ptr, which is not "const",  suggesting `const uint8_t*`
perfer update from chunk.h

xiaoxichen (changes requested):

src/lib/device/vchunk.cpp
JacksonYao287: a const function of a class means it will never modify the member of this class. the only member in VChunk is `cshared internalChunk` (`const shared_ptr`).

mark this function as const means we should not let `internalChunk` point to another chunk or be modified, that is it. but we could modify the `chunk` which it points to, which does not disobey the Semantics of const.

well, since we have already has `const shared_ptr` instead of `shared_ptr` for `internalChunk`, so actually it is unnecessary to add `const` to this function。is this your point here?

src/lib/device/vchunk.cpp
JacksonYao287: agree

src/lib/device/vchunk.cpp
xiaoxichen: it is very misleading that a "set" call could be "const",  which is weird.

JacksonYao287: sure，let`s remove the const.

src/lib/device/round_robin_chunk_selector.cpp
```diff
@@ -12,23 +12,22 @@
...
+#include "round_robin_chunk_selector.h"
 
 namespace homestore {
 RoundRobinChunkSelector::RoundRobinChunkSelector(bool dynamic_chunk_add) : m_dynamic_chunk_add{dynamic_chunk_add} {
     RELEASE_ASSERT_EQ(dynamic_chunk_add, false,
                       "Dynamically adding chunk to chunkselector is not supported, need RCU to make it thread safe");
 }
 
-void RoundRobinChunkSelector::add_chunk(cshared< Chunk >& chunk) { m_chunks.push_back(chunk); }
+void RoundRobinChunkSelector::add_chunk(VChunk chunk) { m_chunks.push_back(chunk); }
```
hkadayam, line 23: Can you do std::move(chunk) instead of copy VChunk, which does increment and decrement of atomic.

src/lib/device/round_robin_chunk_selector.h
```diff
@@ -43,14 +35,12 @@ class RoundRobinChunkSelector : public ChunkSelector {
...
     ~RoundRobinChunkSelector() = default;
 
-    void add_chunk(cshared< Chunk >& chunk) override;
-
-    Chunk* select(blk_count_t nblks, const blk_alloc_hints& hints) override;
-
-    void foreach_chunks(std::function< void(cshared< Chunk >&) >&& cb) override;
+    void add_chunk(VChunk chunk) override;
+    Chunk* select_chunk(blk_count_t nblks, const blk_alloc_hints& hints) override;
+    void foreach_chunks(std::function< void(VChunk&) >&& cb) override;
```
hkadayam, line 40: Is there a way to avoid std::function, to auto& and make use of lambda. If not, we don't want to call foreach_chunks on very critical path. I know it is a micro optimization, but one that accumulates over time.

src/lib/device/round_robin_chunk_selector.cpp
yamingk: My original comment is again based on VChunk is an abstract class, which doesn't hold shared_ptr of Chunk. Nevermind if you want to keep it this way.

src/include/homestore/chunk_selector.h
```diff
@@ -0,0 +1,30 @@
...
+    virtual void foreach_chunks(std::function< void(VChunk&) >&& cb) = 0;
+    virtual Chunk* select_chunk(blk_count_t nblks, const blk_alloc_hints& hints) = 0;
```
yamingk, line 26: Is there any challenge or reason that we need to expose two different types VChunk/Chunk* to consumer with add/select? 
Can we either return VChunk or Chunk for both add/select apis?

src/lib/device/round_robin_chunk_selector.cpp
xiaoxichen: it is back to the https://ebay-eng.slack.com/archives/CAVQXEAKS/p1694203171225199 discussion, vchunk only has single member <cshared_ptr>

src/include/homestore/chunk_selector.h
xiaoxichen: customer code can use
```
ChunkSelector::add_chunk(chunk *)
chunk *p_chunk = ChunkSelector:select()
```

JacksonYao287: for now, add_chunk(cshared< Chunk >& chunk) is actually used in homestore. if we pass raw pointer directly, this might case a pointer leaker, since a raw pointer is manager by different shared_ptr.

if we want to use `add_chunk(chunk *)` , we need to change the use style in homestore

src/lib/device/round_robin_chunk_selector.cpp
JacksonYao287: sure

src/lib/device/round_robin_chunk_selector.h
JacksonYao287: it tried , but the compiler throws an error that we can not use `auto&` or `auto&&` in a virtual function declaration

src/include/homestore/chunk_selector.h
JacksonYao287: @yamingk i change it to return VChunk in both add/select apis to keep the style consistent, and do the transition inside homestore

xiaoxichen (approved):

src/lib/device/virtual_dev.cpp
```diff
@@ -197,7 +197,7 @@ BlkAllocStatus VirtualDev::do_alloc_blk(blk_count_t nblks, const blk_alloc_hints
         size_t attempt{0};
 
         do {
-            chunk = m_chunk_selector->select(nblks, hints);
+            chunk = m_chunk_selector->select_chunk(nblks, hints).get();
```
xiaoxichen, line 200: @hkadayam This is dangerous as raw ptr/shared ptr are mixed used.

The cshare<chunk>  was born in `PhysicalDev::create_chunks` and all the way pass to chunkSelector, however the consumer part(allocator, etc) use raw ptr get from the cshare<chunk>.

It is functioning fine as a reference to every chunk was kept at `m_chunks_map` but it is a bad practice. Maybe lets have a ticket to move those *chunk to cshare<chunk> ?

hkadayam: I don't agree with this change of doing a .get() and select_chunk() returning a shared_ptr. Why incrementing atomic reference when all you are doing is accessing an in-memory member. I am not asking to change now, but will change in next PR.

I would rather return a chunk_id and look it up on the all chunks in vdev. If vdev doesn't maintain all chunks, then I would rather ask it to return a reference to chunks. This is a critical path for homestore, don't want unneccessary atomic reference increment and decrements.

xiaoxichen: the upstream (PhysicalDev::create_chunks ) use shared_ptr<chunk>, if we dont want ref counting we can change the interface of add/select to ref/raw_ptr.  It could work as pdev holds the shared_ptr that ensuring the chunks will not be de-construct,  just not safe.

PR #166, opened 2023-09-13
MultiBlkId for all vdev and services to avoid caller to use std::vector

hkadayam:


src/include/homestore/blk.h
```diff
@@ -24,122 +24,162 @@
...
 
+public:
+    BlkId() = default;
     explicit BlkId(uint64_t id_int);
-    BlkId(blk_num_t blk_num, blk_count_t nblks, chunk_num_t chunk_num = 0);
-    BlkId() { invalidate(); }
-    BlkId(const BlkId&) = default;
-    BlkId& operator=(const BlkId&) = default;
+    BlkId(blk_num_t blk_num, blk_count_t nblks, chunk_num_t chunk_num);
+    BlkId(BlkId const&) = default;
```
yamingk, line 64: I take it as readability preference, for `T const&` over `const T&`?

src/include/homestore/blk.h
```diff
@@ -24,122 +24,162 @@
...
-// hash function definitions
+///////////////////// hash function definitions /////////////////////
 namespace std {
 template <>
 struct hash< homestore::BlkId > {
-    typedef homestore::BlkId argument_type;
-    typedef size_t result_type;
-    result_type operator()(const argument_type& bid) const noexcept {
-        return std::hash< uint64_t >()(bid.to_integer());
+    size_t operator()(const homestore::BlkId& bid) const noexcept { return std::hash< uint64_t >()(bid.to_integer()); }
```
yamingk, line 165: nit: I wasn't aware of this utility. Won't `bid.to_integer()` served the purpose as hash key since it is unique between all BlkIds? Or there might be some cases that we want to store BlkIds from different SMs?

src/include/homestore/blk.h
```diff
@@ -24,122 +24,162 @@
...
-    bool operator<(const BlkId& other) const { return (compare(*this, other) < 0); }
+
+    bool operator==(BlkId const& other) const { return (compare(*this, other) == 0); }
+    bool operator>(BlkId const& other) const { return (compare(*this, other) > 0); }
+    bool operator<(BlkId const& other) const { return (compare(*this, other) < 0); }
+
+    blk_num_t blk_num() const { return s.m_blk_num; }
+    blk_count_t blk_count() const { return s.m_nblks; }
+    chunk_num_t chunk_num() const { return s.m_chunk_num; }
+    bool is_multi() const { return s.m_is_multi; }
```
yamingk, line 76: can this be moved to MultiBlkId?

src/include/homestore/blk.h
```diff
@@ -24,122 +24,162 @@
...
-/* This structure represents the application wide unique block number. It also encomposses the number of blks. */
+#pragma pack(1)
 struct BlkId {
-private:
-    static constexpr uint64_t s_blk_num_mask{(static_cast< uint64_t >(1) << BLK_NUM_BITS) - 1};
-    static constexpr uint64_t s_nblks_mask{(static_cast< uint64_t >(1) << NBLKS_BITS) - 1};
-    static constexpr uint64_t s_chunk_num_mask{(static_cast< uint64_t >(1) << CHUNK_NUM_BITS) - 1};
+protected:
+    struct serialized {
+        blk_num_t m_is_multi : 1; // Is it a part of multi blkid or not
```
yamingk, line 47: I probably didn't get the reason BlkId has to understand this `m_is_multi`? 
What will be the problem if BlkId doesn't know about `m_is_multi`, I think MultiBlkId should just work as it is, no?

src/lib/blkalloc/varsize_blk_allocator.cpp
```diff
@@ -405,251 +384,380 @@ void VarsizeBlkAllocator::fill_cache_in_portion(blk_num_t portion_num, blk_cache
...
-                total_allocated += num_allocated;
-                BLKALLOC_LOG(TRACE, "Alloced additional blk_num={} directly", out_blkids.back().to_string());
-            } else {
-                // failure to get more is really partial if we have some
-                BLKALLOC_LOG(TRACE, "Failed to alloc additional blks directly with code {}", status2);
-                if (status == BlkAllocStatus::PARTIAL) status2 = BlkAllocStatus::PARTIAL;
-            }
-            status = status2;
+    MultiBlkId tmp_blkid;
+    MultiBlkId& out_mbid = out_blkid.is_multi() ? r_cast< MultiBlkId& >(out_blkid) : tmp_blkid;
```
xiaoxichen, line 413: can we fallback to alloc_contiguous if `is_multi() = false` ?

Or maybe instead of having `alloc_contiguous` and `alloc`,  we have overload `alloc` 
```
  blk_count_t nblks, blk_alloc_hints const& hints, MultiBlkId& out_blkid)  ->original alloc
  blk_count_t nblks, blk_alloc_hints const& hints, BlkId& out_blkid).  --> alloc_contiguous
 ```

File: src/lib/blkalloc/varsize_blk_allocator.cpp [hunk 6]
```diff
@@ -405,251 +384,380 @@ void VarsizeBlkAllocator::fill_cache_in_portion(blk_num_t portion_num, blk_cache
...
+    return n_freed;
+}
 
-void VarsizeBlkAllocator::free_on_bitmap(const BlkId& b) {
-    BlkAllocPortion& portion = blknum_to_portion(b.get_blk_num());
-    {
-        auto const start_blk_id = portion.get_portion_num() * get_blks_per_portion();
-        auto const end_blk_id = start_blk_id + get_blks_per_portion() - 1;
+bool VarsizeBlkAllocator::is_blk_alloced(BlkId const& b, bool use_lock) const {
+    if (!m_inited) { return true; }
```
xiaoxichen, line 707: do we need to handle MultiBlkId here?

src/lib/blkalloc/varsize_blk_allocator.cpp
```diff
@@ -405,251 +384,380 @@ void VarsizeBlkAllocator::fill_cache_in_portion(blk_num_t portion_num, blk_cache
...
 }
 
+blk_num_t VarsizeBlkAllocator::available_blks() const { return get_total_blks() - get_used_blks(); }
+blk_num_t VarsizeBlkAllocator::get_used_blks() const { return get_alloced_blk_count(); }
+
 #ifdef _PRERELEASE
-bool VarsizeBlkAllocator::is_set_on_bitmap(const BlkId& b) const {
-    const BlkAllocPortion& portion = blknum_to_portion_const(b.get_blk_num());
+bool VarsizeBlkAllocator::is_set_on_bitmap(BlkId const& b) const {
+    BlkAllocPortion const& portion = blknum_to_portion_const(b.blk_num());
```
xiaoxichen, line 732: similar here,  MultiBlkId support/check

xiaoxichen (commented):
Only reviewed
```
blk.h/blk.cpp
varsize_blk_allocator.cpp/.hpp (Not used by NuObject)
```
rest still wip

src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -53,87 +53,155 @@ void BlkDataService::create_vdev(uint64_t size, blk_allocator_type_t alloc_type,
...
+
+    if (blkid.num_pieces() == 1) {
+        return do_read(blkid.to_single_blkid(), buf, size, part_of_batch);
+    } else {
+        static thread_local std::vector< folly::Future< std::error_code > > s_futs;
+        s_futs.clear();
+
+        auto it = blkid.iterate();
+        while (auto const bid = it.next()) {
+            s_futs.emplace_back(do_read(*bid, buf, size, part_of_batch));
```
xiaoxichen, line 92: those N reads overwritten the same buffer with each other,  why we need this?

src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -53,87 +53,155 @@ void BlkDataService::create_vdev(uint64_t size, blk_allocator_type_t alloc_type,
...
     } else {
-        static thread_local std::vector< folly::Future< bool > > s_futs;
+        static thread_local std::vector< folly::Future< std::error_code > > s_futs;
         s_futs.clear();
-        sisl::sg_iterator sg_it{sgs.iovs};
-        for (const auto& bid : blkids) {
-            const auto iovs = sg_it.next_iovs(bid.get_nblks() * m_page_size);
-            s_futs.emplace_back(m_vdev->async_writev(iovs.data(), iovs.size(), bid, part_of_batch));
+
+        auto it = blkid.iterate();
```
xiaoxichen, line 118: add
```
auto sg_iter = sisl::sg_iterator(sgs)
```

src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -53,87 +53,155 @@ void BlkDataService::create_vdev(uint64_t size, blk_allocator_type_t alloc_type,
...
+        while (auto const bid = it.next()) {
+            s_futs.emplace_back(do_read(*bid, sgs, size, part_of_batch));
```
xiaoxichen, line 120:
```
this_iov = sg_iter.next_iovs(this_read_size)
s_futs.emplace_back(do_read(*bid, this_iov, size, part_of_batch));
```

src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -53,87 +53,155 @@ void BlkDataService::create_vdev(uint64_t size, blk_allocator_type_t alloc_type,
...
+        return collect_all_futures(s_futs);
+    }
 }
 
-folly::Future< bool > BlkDataService::async_write(const sisl::sg_list& sgs, const blk_alloc_hints& hints,
-                                                  const std::vector< BlkId >& blkids, bool part_of_batch) {
-    if (blkids.size() == 1) {
-        // Shortcut to most common case
-        return m_vdev->async_writev(sgs.iovs.data(), sgs.iovs.size(), blkids[0], part_of_batch);
+folly::Future< std::error_code > BlkDataService::async_read(MultiBlkId const& blkid, sisl::sg_list& sgs, uint32_t size,
```
xiaoxichen, line 99: in general this function is not implemented correctly as it doesnt seperate sgs::iovs for each BlkId

src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -53,87 +53,155 @@ void BlkDataService::create_vdev(uint64_t size, blk_allocator_type_t alloc_type,
...
+    } else {
+        auto it = blkid.iterate();
+        while (auto const bid = it.next()) {
+            m_vdev->commit_blk(*bid);
+        }
+    }
+}
+
+folly::Future< std::error_code > BlkDataService::async_free_blk(MultiBlkId const& bids) {
+    // Shortcut to most common case
```
xiaoxichen, line 196: remove this comment as vdev::free_blk() did the heavy lift

src/include/homestore/blk.h
yamingk: Continous blkId is also one blkid which is still unique.

xiaoxichen: sorry I mean
blkId{0, 0, 1} and blkid{0, 1, 1}   i.e chunk0, first blk and second blk,  in to_interger() they are continues which is 0/1.
hashing re-distribute them to the uint64 space, maybe 0 and 2048576, which re-distribute the continues blkID across whole uint64 space to avoid hot-spot.

Though honestly I dont see a use case in code right now

src/include/homestore/blk.h
yamingk: @xiaoxichen, we don't rely on `m_is_multi` to know how to deserialize the BlkID/MultiBlkId. It is a compare of blob.size vs sizeof(BlkId) to know whether this is a MultiBlkId of a BlkId. 

I think the reason is a lot of places is using `bid.is_multi()` to determine the need to access multi blkids. 
But it also be achieved by `dynamic_cast<MultiBlkId*>` with BlkId knows nothing about how derived class is using this base class. I could missing some other use cases that forbid us from doing above and would like to learn what are they.

src/include/homestore/blk.h
hkadayam: I prefer `T const&` over `const T&`, but I am not going ahead and changing everywhere and it is upto individual preference, if one has strong inclination over other.

IMHO: `T const&` gives more meaning since it says it defines the data type and then tells whether its a reference or const reference or by value.

src/include/homestore/blk.h
hkadayam: std::hash of integer implementation is to just literal `no-op`, but we don't want to assume that. We have an integer and let the compiler/STL decides how to convert that to near unique. More over the to_integer() could return `uint32_t` and in that case, we need to make it `size_t`

hkadayam: @xiaoxichen This is needed for us to put MultiBlkId in STL containers std::unordered_map, concurrent_hashmap etc

src/include/homestore/blk.h
hkadayam: MultiBlkId is not a vtable based dervied class, it is a non-virtual derived class. This is done for backward compatibility where upper layer refers them as BlkId and we need to see if they are multi or not (from BlkId reference). It is used to assert that and we would also need to serialize the BlkId. Hence this should stay in BlkId.

src/include/homestore/blk.h
hkadayam: > bool type might be more readable

bool type size is not guaranteed across and it uses min 1 byte, whereas we need to serialize as 1 bit.

hkadayam: The MultiBlkId is not a virtual derived class of BlkId. It is present right now as a way for upper layer to continue to use BlkId in-place of MultiBlkId. For example, metablk uses the BlkId everywhere and I wanted to make sure their usage is not broken. Once we address in all the places, we can make MultiBlkId as non-derived class of BlkId.

src/include/homestore/blk.h
```diff
@@ -24,122 +24,162 @@
...
-
-    BlkId8_t& operator=(const BlkId& rhs) {
-        BlkId::operator=(rhs);
-        return *this;
-    }
+    bool operator==(MultiBlkId const& other) const { return (compare(*this, other) == 0); }
+    bool operator>(MultiBlkId const& other) const { return (compare(*this, other) > 0); }
+    bool operator<(MultiBlkId const& other) const { return (compare(*this, other) < 0); }
+
+    sisl::blob serialize();
```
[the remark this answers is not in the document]
hkadayam: I can do const for sure, can't do override I think since its not a virtual derived class.

src/include/homestore/blk.h
```diff
@@ -24,122 +24,162 @@
...
-    typedef size_t result_type;
-    result_type operator()(const argument_type& bid) const noexcept {
-        return std::hash< uint64_t >()(bid.to_integer());
+    size_t operator()(const homestore::BlkId& bid) const noexcept { return std::hash< uint64_t >()(bid.to_integer()); }
+};
+
+template <>
+struct hash< homestore::MultiBlkId > {
+    size_t operator()(const homestore::MultiBlkId& mbid) const noexcept {
+        static constexpr size_t s_start_seed = 0; // TODO: Pickup a better seed
```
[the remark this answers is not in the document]
hkadayam: Thanks, will use that.

src/include/homestore/blk.h
```diff
@@ -163,27 +218,40 @@ std::basic_ostream< charT, traits >& operator<<(std::basic_ostream< charT, trait
...
-    uint32_t multiplier;           // blks allocated in a blkid should be a multiple of multiplier
-    uint32_t max_blks_per_entry;   // Number of blks on every entry
-    uintptr_t stream_info;
-#ifdef _PRERELEASE
-    bool error_simulate = false; // can error simulate happen
-#endif
+    blk_temp_t desired_temp{0};                  // Temperature hint for the device
+    std::optional< uint32_t > pdev_id_hint;      // which physical device to pick (hint if any) -1 for don't care
+    std::optional< chunk_num_t > chunk_id_hint;  // any specific chunk id to pick for this allocation
+    std::optional< stream_id_t > stream_id_hint; // any specific stream to pick
```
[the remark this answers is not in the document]
hkadayam: we have pdev_id_hint for that (I am assuming you want one PG to be one pdev and not 1 stream right). Stream is a children of PDev.

src/lib/blkdata_svc/blkdata_service.cpp
hkadayam: Ok

src/lib/blkdata_svc/blk_read_tracker.cpp
```diff
@@ -98,8 +94,16 @@ void BlkReadTracker::merge(const BlkId& blkid, int64_t new_ref_count,
 void BlkReadTracker::insert(const BlkId& blkid) { merge(blkid, 1, nullptr); }
 void BlkReadTracker::remove(const BlkId& blkid) { merge(blkid, -1, nullptr); }
 
-void BlkReadTracker::wait_on(const BlkId& blkid, after_remove_cb_t&& after_remove_cb) {
-    merge(blkid, 0, std::make_shared< blk_track_waiter >(std::move(after_remove_cb)));
+void BlkReadTracker::wait_on(MultiBlkId const& blkids, after_remove_cb_t&& after_remove_cb) {
+    if (blkids.num_pieces() == 1) {
+        merge(blkids, 0, std::make_shared< blk_track_waiter >(std::move(after_remove_cb)));
+    } else {
+        auto waiter = std::make_shared< blk_track_waiter >(std::move(after_remove_cb));
```
yamingk, line 101: There will be race, right? When a read finishes on this specifc blkid but not yet for other blkids, the remove_cb will be called immaturely by freeing all the blkids in vdev?

src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -53,87 +53,155 @@ void BlkDataService::create_vdev(uint64_t size, blk_allocator_type_t alloc_type,
...
+
     // create blk read waiter instance;
-    folly::Promise< bool > promise;
+    folly::Promise< std::error_code > promise;
     auto f = promise.getFuture();
 
-    m_blk_read_tracker->wait_on(bid, [this, bid, p = std::move(promise)]() mutable {
-        m_vdev->free_blk(bid);
-        p.setValue(true);
+    m_blk_read_tracker->wait_on(bids, [this, bids, p = std::move(promise)]() mutable {
```
yamingk, line 202: I would like to suggest do the similar way done in read/write, that `m_blk_read_tracker` still handles BlkId, not MultiBlkId.

src/include/homestore/blk.h
yamingk: I understand it require a lot of dyamic_cast<BlkID> to be added for a virtual derived class. 
It seems a virtual derived class makes more sense, with BlkId takens care of single blkid, and MultiBlkId takens care of fragmented BlkId, and single blkid's business be taken care by its base class. 

Or non-derived version with BlkId as a container of fragmented blkids.

Totally fine if it needs to be done in two phases, if it is two much in one shot. You can decide how to approach.

src/include/homestore/blk.h
```diff
@@ -24,122 +24,162 @@
...
+};
+#pragma pack()
 
-    void set_blk_num(blk_num_t blk_num);
-    blk_num_t get_blk_num() const { return m_blk_num; }
-    // last blk num is the last blk num that belongs this blkid;
-    blk_num_t get_last_blk_num() const { return get_blk_num() + get_nblks() - 1; }
+#pragma pack(1)
+struct MultiBlkId : public BlkId {
+    static constexpr uint32_t max_addln_pieces{5};
```
yamingk, line 92: Is there a reason for max_addln_pieces to be such a small number?
Or is it because we don't want to dynamic extending of the pieces?

src/include/homestore/blk.h
yamingk: @hkadayam we return uint64_t for to_integer(), not uint32_t, right? Isn't that good enough for `std::unordered_map<uint64_t /* bid.to_integer() as key */, xxx>` ?

src/include/homestore/blk.h
```diff
@@ -163,27 +218,40 @@ std::basic_ostream< charT, traits >& operator<<(std::basic_ostream< charT, trait
...
-    bool is_contiguous;
-    uint32_t multiplier;           // blks allocated in a blkid should be a multiple of multiplier
-    uint32_t max_blks_per_entry;   // Number of blks on every entry
-    uintptr_t stream_info;
-#ifdef _PRERELEASE
-    bool error_simulate = false; // can error simulate happen
-#endif
+    blk_temp_t desired_temp{0};                  // Temperature hint for the device
+    std::optional< uint32_t > pdev_id_hint;      // which physical device to pick (hint if any) -1 for don't care
+    std::optional< chunk_num_t > chunk_id_hint;  // any specific chunk id to pick for this allocation
```
yamingk, line 248: I saw you are already adding chunk_id_hint do you want to also handle it in `VirtualDev::do_alloc_blk`, or I can do it with my PR too, just it might cause some merge conflict to your PR if I merge first, which is a small issue.

src/include/homestore/blk.h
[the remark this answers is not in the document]
xiaoxichen: One approach is maintaining the PG->Pdev mapping + Active Shard->Chunk mapping inside the heap chunk selector, so that those placement logic is centrolized.

If we take this part out to else where of HO,  HO::CreateShard needs to identify the Bid/ChunkID return by chunk_selector in `alloc_write` and saves the mapping , which seems a bit wired

src/include/homestore/blk.h
hkadayam: @yamingk Suggested practice for hash is to do std::hash(<integer>) even for integers. Let STL deal their evenness.

src/include/homestore/blk.h
hkadayam: We can't do vtable if we are going to directly serialize this, since sizeof(BlkId) will become 16 bytes instead of 8 bytes. I don't see much value in making BlkId being virtual in any case. In fact we should try to reverse, where I am thinking of making MultiBlkId independent of BlkId. Only reason it is done today is for backward compatibility (especially metablks directly persists BlkId) in persistent structures.

src/include/homestore/blk.h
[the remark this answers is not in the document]
hkadayam: Since it is directly returning uint8_t*, const will fail. Thats why I remember removing const from this. I have added a TODO to handle this, by doing something like sisl::const_blob for reasons like these.

src/lib/blkalloc/blk.cpp
```diff
@@ -17,94 +17,162 @@
...
+        n_addln_piece = 0;
+    } else {
+        n_addln_piece = other->n_addln_piece;
+        std::copy(other->addln_pieces.begin(), other->addln_pieces.begin() + other->n_addln_piece,
+                  addln_pieces.begin());
+    }
 }
 
-void BlkId::set(const BlkId& bid) { set(bid.get_blk_num(), bid.get_nblks(), bid.get_chunk_num()); }
+uint16_t MultiBlkId::num_pieces() const { return n_addln_piece + (BlkId::is_valid() ? 1 : 0); }
```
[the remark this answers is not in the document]
hkadayam: Excellent catch. Thanks

src/lib/blkalloc/varsize_blk_allocator.cpp
```diff
@@ -405,251 +384,380 @@ void VarsizeBlkAllocator::fill_cache_in_portion(blk_num_t portion_num, blk_cache
...
-        }
+        if (realtime_bm_on()) { alloc_on_realtime(out_mbid); }
 
 #ifdef _PRERELEASE
-        alloc_sanity_check(total_allocated, hints, out_blkids);
+        alloc_sanity_check(num_allocated, hints, out_mbid);
 #endif
     }
 
+    if (!out_blkid.is_multi()) { out_blkid = out_mbid.to_single_blkid(); }
```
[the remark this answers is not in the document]
hkadayam: out_mbid.num_pieces() is never > 1 for !is_multi() cases right?

src/lib/blkalloc/varsize_blk_allocator.cpp
hkadayam: I will add that code, but as is is_blk_alloced is not used or this code is to be modified not to assert here.

src/lib/device/virtual_dev.cpp
```diff
@@ -198,13 +186,19 @@ BlkAllocStatus VirtualDev::do_alloc_blk(blk_count_t nblks, const blk_alloc_hints
...
+            }
 
-            status = alloc_blk_from_chunk(nblks, hints, out_blkid, chunk);
-            if (status == BlkAllocStatus::SUCCESS || !hints.can_look_for_other_chunk) { break; }
+            status = alloc_blks_from_chunk(nblks, hints, out_blkid, chunk);
+            if ((status == BlkAllocStatus::SUCCESS) || !hints.can_look_for_other_chunk ||
+                (status == BlkAllocStatus::PARTIAL && hints.partial_alloc_ok)) {
+                break;
+            }
         } while (++attempt < m_all_chunks.size());
```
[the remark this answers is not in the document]
hkadayam: It is entirely upto chunk_selector right? Say if have 10 chunks and 9 doesn't have space, I don't want to return no_space since 10th might have it.

src/lib/device/virtual_dev.cpp
```diff
@@ -198,13 +186,19 @@ BlkAllocStatus VirtualDev::do_alloc_blk(blk_count_t nblks, const blk_alloc_hints
...
 
-        if (status != BlkAllocStatus::SUCCESS) {
+        if ((status != BlkAllocStatus::SUCCESS) || (status != BlkAllocStatus::PARTIAL)) {
             LOGERROR("nblks={} failed to alloc after trying to alloc on every chunks {} and devices {}.", nblks);
```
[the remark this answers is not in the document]
hkadayam: This message is correct for roundrobin and more importantly any non-custom chunk selector. What HomeObject use case for chunk selection is slightly an outlier.

src/lib/device/virtual_dev.cpp
```diff
@@ -217,41 +211,66 @@ BlkAllocStatus VirtualDev::do_alloc_blk(blk_count_t nblks, const blk_alloc_hints
...
     if (auto const fake_status =
             iomgr_flip::instance()->get_test_flip< uint32_t >("blk_allocation_flip", nblks, chunk->vdev_id())) {
         return static_cast< BlkAllocStatus >(fake_status.get());
     }
 #endif
-    static thread_local std::vector< BlkId > chunk_blkid{};
-    chunk_blkid.clear();
-    auto status = chunk->blk_allocator_mutable()->alloc(nblks, hints, chunk_blkid);
-    if (status == BlkAllocStatus::PARTIAL) {
+    auto status = chunk->blk_allocator_mutable()->alloc(nblks, hints, out_blkid);
```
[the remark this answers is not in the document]
hkadayam: BlkAllocator should take MultiBlkId because blk_allocator has functionality and much faster to allocate multiple blks. There is a big logic around varsize_blk_allocation to efficiently do that.

src/lib/blkdata_svc/blkdata_service.cpp
hkadayam: Yes there are issues in async_read/readv routines. Thanks for pointing out. I have fixed these. I have added few more test case surrounding that in next sets of PRs.

src/lib/blkdata_svc/blkdata_service.cpp
hkadayam: As mentioned above

src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -53,87 +53,155 @@ void BlkDataService::create_vdev(uint64_t size, blk_allocator_type_t alloc_type,
...
 }
 
-folly::Future< bool > BlkDataService::async_write(const sisl::sg_list& sgs, const blk_alloc_hints& hints,
-                                                  const std::vector< BlkId >& blkids, bool part_of_batch) {
-    if (blkids.size() == 1) {
-        // Shortcut to most common case
-        return m_vdev->async_writev(sgs.iovs.data(), sgs.iovs.size(), blkids[0], part_of_batch);
+folly::Future< std::error_code > BlkDataService::async_read(MultiBlkId const& blkid, sisl::sg_list& sgs, uint32_t size,
+                                                            bool part_of_batch) {
+    auto do_read = [this](BlkId const& bid, sisl::sg_list& sgs, uint32_t size, bool part_of_batch) {
```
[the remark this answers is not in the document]
hkadayam: As mentioned above

src/include/homestore/blk.h
hkadayam: Having higher number will increase the memory foot print, because it is freely passed around. As is they are 48 bytes. If a block is not available in 6 pieces, then its better to return the allocation failure than doing it otherwise IMO.

src/lib/blkdata_svc/blkdata_service.cpp
hkadayam: If we do that, caller has to create multiple shared_ptrs and caller has to wait for all remove_cb and handle vector of futures. So all parties are waiting for same event and its better to wait as a group, than individually set it up to wait and then wait for all waiters to finish right?

src/lib/blkdata_svc/blk_read_tracker.cpp
hkadayam: We are having a single waiter instance for all `blkids`, so if any read waits for a specific blkid, they will only call waiter after all refcount goes to zero.

In other words, isn't this same scenario as in existing implementation. If a blkid is bigger than what it can hold, it already splits into multiple smaller blkids and adding a reference. Now that is extended to multiple pieces of BlkIds right?

src/lib/blkalloc/varsize_blk_allocator.cpp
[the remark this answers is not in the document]
xiaoxichen: the logic here seems like if passed-in parameter out_blkid is sinle (`if (!out_blkid.is_multi())`,  then we force-cast the out_mbid to single regardless the is_multi of out_mbid

src/lib/device/virtual_dev.cpp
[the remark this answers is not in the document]
xiaoxichen: yes but there is no gareentee in the API contract that it will return N different chunks for N different select_chunk, so  to some point we need to believe chunk_selector does its best scheduling,  for the given allocation_hint. It says no then no (retry just to cover a few edge case that you have add_chunk/delete_blob or similar release space operation in-flight).

src/lib/device/virtual_dev.cpp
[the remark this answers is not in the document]
xiaoxichen: yes, just a bit misleading if we see this log from prod and forget the contxt, but that's fine

src/lib/device/virtual_dev.cpp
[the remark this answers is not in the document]
xiaoxichen: as discussed in other thread, if we treat HO's selector as an out lier then this makes sense.

src/lib/device/virtual_dev.cpp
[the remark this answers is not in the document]
xiaoxichen: sorry my bad, I forgot no overload as we dont do virtual.....

xiaoxichen (approved):

src/lib/blkdata_svc/blk_read_tracker.cpp
yamingk: Say we 2 separate blks addressed by this MultiBlkId. one blkid is merged and waiter waits on it. Before this thread got chance to merge the 2nd (line: 104), read completes on the 1st one and waiter got zero ref_cnt and waiter will be gone and trigger the callback immaturely. 

Existing implementation is a one-blk-id, one waiter pattern, here we have multiple blk-ids, one waiter, which breaks the assumption blkreader track is working. 

Now this assumption is not there and one pending blk read completes could affect (free) all the other blks in vdev.

src/include/homestore/blk.h
yamingk: I am fine doing non-derived version.

src/lib/device/virtual_dev.cpp
```diff
@@ -217,41 +211,66 @@ BlkAllocStatus VirtualDev::do_alloc_blk(blk_count_t nblks, const blk_alloc_hints
     }
 }
 
-BlkAllocStatus VirtualDev::alloc_blk_from_chunk(blk_count_t nblks, const blk_alloc_hints& hints,
-                                                std::vector< BlkId >& out_blkid, Chunk* chunk) {
+BlkAllocStatus VirtualDev::alloc_blks(blk_count_t nblks, blk_alloc_hints const& hints,
```
yamingk, line 214: It also needs to be exposed to data service layer for caller to switch the API if alloc_blks with MultiBlkId fails, right?

src/lib/blkdata_svc/blk_read_tracker.cpp
hkadayam: We will call the callback only on the destructor of `blk_track_waiter` right? The scenario you mentioned will not happen before all the blk_ids are added, because `blk_track_waiter` will still be in scope.

src/include/homestore/blk.h
hkadayam: I think we can handle the chunk_id hints for blk allocation as part of another PR.

src/lib/blkdata_svc/blk_read_tracker.cpp
yamingk: every merge operation will push the waiter into the BlkReadTrack record, and once one merge operation is done, there is nothing can prevent read completes routine to call m_read_blk_tracker->remove(blkid, xxx) which will trigger a deref to that record, in return destruct the waiter.

yamingk: I take back the race, we are holding the shared_ptr of waiter. 
But we are forcing all blk ids to be freed as a whole for each single blk to be able to be freed in vdev, which is un-necessary, right? In fact, each blk-id should be able to get freed once its pending blkid finished their reads.

yamingk (approved):

PR #167, opened 2023-09-14
SDSTOR-11755. Make VChunk CopyAssignable and CopyConstructible

JacksonYao287:
In HeapChunkSelector, we use a priority_queue as the heap, which will take std::vector or std:deque as the internal container to do the heap sort. well,  a type T in std::vector or std:deque must meet the requirements of [CopyAssignable](https://en.cppreference.com/w/cpp/named_req/CopyAssignable) and [CopyConstructible](https://en.cppreference.com/w/cpp/named_req/CopyConstructible). so we need to make VChunk meeting this requirements.

actually, the complier will generate default copy constructor and copy assign operator if necessary, the default action of the above two is just copy all the member of the class one by one. now , in Vchunk the only member is `cshard<chunk>`, which can not be assigned a new value, so we need to change this to a non-const `shard_ptr`

### DIFF
--- a/src/include/homestore/vchunk.h
+++ b/src/include/homestore/vchunk.h
@@ -34,6 +34,6 @@ public:
     cshared< Chunk > get_internal_chunk() const;
 
 private:
-    shared< Chunk > internalChunk;
+    shared< Chunk > m_internal_chunk;
 };
 } // namespace homestore
--- a/src/include/homestore/vchunk.h
+++ b/src/include/homestore/vchunk.h
@@ -24,7 +24,7 @@ class Chunk;
 class VChunk {
 public:
     VChunk(cshared< Chunk > const&);
-    
+
     ~VChunk() = default;
 
     void set_user_private(const sisl::blob& data);
@@ -34,6 +34,6 @@ public:
     cshared< Chunk > get_internal_chunk() const;
 
 private:
-    cshared< Chunk > internalChunk;
+    shared< Chunk > internalChunk;
 };
-}// namespace homestore
+} // namespace homestore
--- a/src/lib/device/vchunk.cpp
+++ b/src/lib/device/vchunk.cpp
@@ -17,23 +17,15 @@
 #include "device/chunk.h"
 
 namespace homestore {
-    VChunk::VChunk(cshared< Chunk >& chunk) : internalChunk(chunk){}
+VChunk::VChunk(cshared< Chunk >& chunk) : m_internal_chunk(chunk) {}
 
-    void VChunk::set_user_private(const sisl::blob& data){
-        internalChunk->set_user_private(data);
-    }
+void VChunk::set_user_private(const sisl::blob& data) { m_internal_chunk->set_user_private(data); }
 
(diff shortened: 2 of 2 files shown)


JacksonYao287:
If we implement the copy constructor and copy assign operator by ourself, the const member can be initialized in the initialization_list , right?

if copy constructor , yes. but copy assign operator seems not , since copy assign operator has not initialization_list，it is used to copy an existed object to another exist object

src/include/homestore/chunk_selector.h
```diff
@@ -21,8 +21,8 @@ class Chunk;
 class ChunkSelector {
 public:
     ChunkSelector() = default;
-    virtual void add_chunk(cshared< Chunk >&) = 0;
-    virtual void foreach_chunks(std::function< void(cshared< Chunk >&) >&& cb) = 0;
+    virtual void add_chunk(cshared< Chunk >) = 0;
+    virtual void foreach_chunks(std::function< void(cshared< Chunk >) >&& cb) = 0;
```
hkadayam, line 25: I am not finding the reason to change Chunk to Chunk&. It puts onus on the caller to move and if they are using it for only referencing we should pass Chunk&. Sending shared_ptr by value for referencing is not the right thing to do.

src/include/homestore/vchunk.h
```diff
@@ -34,6 +34,6 @@ class VChunk {
     cshared< Chunk > get_internal_chunk() const;
 
 private:
-    cshared< Chunk > internalChunk;
+    shared< Chunk > internalChunk;
```
hkadayam, line 37: I suggest to use consistent naming convention of the repo. Homestore doesn't use camel case
internalChunk -> m_internal_chunk

src/lib/device/round_robin_chunk_selector.cpp
```diff
@@ -20,14 +20,14 @@ RoundRobinChunkSelector::RoundRobinChunkSelector(bool dynamic_chunk_add) : m_dyn
                       "Dynamically adding chunk to chunkselector is not supported, need RCU to make it thread safe");
 }
 
-void RoundRobinChunkSelector::add_chunk(cshared< Chunk >& chunk) { m_chunks.emplace_back(std::move(chunk)); }
+void RoundRobinChunkSelector::add_chunk(cshared< Chunk > chunk) { m_chunks.emplace_back(std::move(chunk)); }
```
hkadayam, line 23: It is ok here, but I am not clear why we needed to do that.

src/include/homestore/vchunk.h
JacksonYao287: sure ， will do that

src/include/homestore/chunk_selector.h
JacksonYao287: sorry for this , i just push a PR w/o cancel the test modification, will change it back

xiaoxichen (approved):

xiaoxichen:
merge it as it is small and all review comments addressed

PR #181, opened 2023-09-22
Fix split, merge, varlen remove, (parent size estimation), simple remove

shosseinimotlagh:
The objective of this pull request is to address specific edge cases in btree implementations that are more likely to occur with lower node sizes (such as 512).

The proposed fixes are:

For split nodes: a) Add a capacity check before inserting entities into the new child. b) Modify configuration settings so that variable sized key/value pairs never exceed the maximum size and provide sufficient capacity to accommodate at least two entities in the worst-case scenarios.

For merge nodes: a) Ensure that the parent node always has enough space to accommodate updated keys before performing the merge for variable-sized node types. b) Correct the number of entries that can be accommodated to the leftmost node for simple node types. c) Address the calculation of available space when removals happen in variable-length node types d) unlocking/free up the nodes must in reverse order.

### DIFF
--- a/src/include/homestore/btree/detail/btree_remove_impl.ipp
+++ b/src/include/homestore/btree/detail/btree_remove_impl.ipp
@@ -288,6 +288,7 @@ bool Btree< K, V >::remove_extents_in_leaf(const BtreeNodePtr& node, BtreeRangeR
 template < typename K, typename V >
 template < typename ReqT >
 btree_status_t Btree< K, V >::check_collapse_root(ReqT& req) {
+    if (!m_bt_cfg.m_merge_turned_on) { return btree_status_t::merge_not_required; }
     BtreeNodePtr child;
     BtreeNodePtr root;
     btree_status_t ret = btree_status_t::success;
@@ -355,7 +356,7 @@ btree_status_t Btree< K, V >::merge_nodes(const BtreeNodePtr& parent_node, const
     for (auto indx = start_idx + 1; indx <= end_idx; ++indx) {
         if (indx == parent_node->total_entries()) {
             BT_NODE_LOG_ASSERT(parent_node->has_valid_edge(), parent_node,
-                               "Assertion failure, expected valid edge for parent_node: {}");
+                               "Assertion failure, expected valid edge for parent_node");
         }
 
         BtreeLinkInfo child_info;
@@ -463,28 +464,25 @@ btree_status_t Btree< K, V >::merge_nodes(const BtreeNodePtr& parent_node, const
 
     if (!K::is_fixed_size()) {
         // Lets see if we have enough room in parent node to accommodate changes. This is needed only if the key is not
-        // fixed length. For fixed length node merge will always result in lesser or equal size
-        int64_t post_merge_size{0};
-        auto& old_node = old_nodes[leftmost_src.ith_nodes.back()];
-        if (old_node->total_entries()) {
-            post_merge_size += old_node->get_nth_obj_size(
-                std::min(leftmost_src.last_node_upto, old_node->total_entries() - 1)); // New leftmost entry
-        }
-        post_merge_size -= parent_node->get_nth_obj_size(start_idx); // Previous left entry
+        // fixed length. For fixed length node merge will always result in less or equal size
 
-        for (auto& node : new_nodes) {
-            if (node->total_entries()) { post_merge_size += node->get_nth_obj_size(node->total_entries() - 1); }
-        }
+        // we first calculate the least amount of space being released after removing excess children. the key size
+        // cannot be taken account; so we know for sure that value (i.e., linkinfo) and also its record will be freed.
+        // If the end_idx is the parent's edge, the space is not released eventually.
+        auto excess_releasing_nodes =
+            old_nodes.size() - new_nodes.size() - parent_node->total_entries() == end_idx ? 1 : 0;
+        auto minimum_releasing_excess_size =
+            excess_releasing_nodes * (BtreeLinkInfo::get_fixed_size() + parent_node->get_record_size());
 
-        for (auto& node : old_nodes) {
-            if (node->total_entries()) { post_merge_size -= node->get_nth_obj_size(node->total_entries() - 1); }
-        }
-
-        if (post_merge_size > parent_node->available_size(m_bt_cfg)) {
+        // aside from releasing size due to excess node, K::get_estimate_max_size is needed for each updating element
+        // at worst case (linkinfo and record remain the same for old and new nodes). The number of updating elements
+        // are the size of the new nodes (the last key of the last new node is not getting updated; hence excluded) plus
+        // the leftmost node.
+        if (parent_node->available_size(m_bt_cfg) + minimum_releasing_excess_size <
+            (1 + new_nodes.size() ? new_nodes.size() - 1 : 0) * K::get_estimate_max_size()) {
             BT_NODE_LOG(DEBUG, parent_node,
-                        "Merge is needed, however after merge it will add {} bytes which is more than "
-                        "available_size={}, so not proceeding with merge",
-                        post_merge_size, parent_node->available_size(m_bt_cfg));
+                        "Merge is needed, however after merge, the parent MAY not have enough space to accommodate the "
+                        "new keys, so not proceeding with merge");
             ret = btree_status_t::merge_not_required;
             goto out;
         }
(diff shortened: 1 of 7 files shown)


src/include/homestore/btree/detail/btree_mutate_impl.ipp
```diff
@@ -552,6 +552,7 @@ btree_status_t Btree< K, V >::split_node(const BtreeNodePtr& parent_node, const
 template < typename K, typename V >
 template < typename ReqT >
 bool Btree< K, V >::is_split_needed(const BtreeNodePtr& node, const BtreeConfig& cfg, ReqT& req) const {
+    if (node->total_entries() <= 1) { return false; }
```
shosseinimotlagh, line 555: It can be possible that in some key/value size setting, the total used size of a  node is more than ideal size even with one entity ( `sizeof(key[0]) + sizeof (value[0]) + record_size > ideal_size`). In this scenario, the split cannot happen.  So a check is needed that if the total entities is 1, split can be performed under no circumstance. 

If the force_split_node is enabled in one thread and the do_put returns retry, it is possible that in another thread get the lock and remove all entities in this node. So a check is needed that if the node is empty no split is needed

src/include/homestore/btree/detail/btree_remove_impl.ipp
```diff
@@ -288,6 +288,7 @@ bool Btree< K, V >::remove_extents_in_leaf(const BtreeNodePtr& node, BtreeRangeR
 template < typename K, typename V >
 template < typename ReqT >
 btree_status_t Btree< K, V >::check_collapse_root(ReqT& req) {
+    if (!m_bt_cfg.m_merge_turned_on) { return btree_status_t::merge_not_required; }
```
shosseinimotlagh, line 291: This is for not collapsing root when `merge_disable `is triggered

src/include/homestore/btree/detail/simple_node.hpp
```diff
@@ -147,8 +147,7 @@ class SimpleNode : public BtreeNode {
     }
 
     uint32_t num_entries_by_size(uint32_t start_idx, uint32_t size) const override {
-        uint32_t possible_entries = (size == 0) ? 0 : (size - 1) / get_nth_obj_size(0) + 1;
-        return std::min(possible_entries, this->total_entries() - start_idx);
+        return std::min(size / get_nth_obj_size(0), this->total_entries() - start_idx);
```
shosseinimotlagh, line 150: for simple node type, when the size was not divisible, the round up value is returned which can lead to out of space happens in merge. 

For example:
```
size = 100
get_nth_obj_size = 30
```

The correct return value must be `100/30 = 3` and not 4.

src/include/homestore/btree/detail/varlen_node.hpp
```diff
@@ -166,7 +166,7 @@ class VariableNode : public BtreeNode {
             get_nth_value(ind_s - 1, &last_1_val, false);
             this->set_edge_value(last_1_val);
 
-            for (uint32_t i = ind_s; i < total_entries; i++) {
+            for (uint32_t i = ind_s - 1; i < total_entries; i++) {
```
shosseinimotlagh, line 169: wrong calculation of available size.
When removing the edge entity happens, the `entity[ind_s-1]` is set as a new edge and removed as the normal entity. So its size must be released because it is set as a new edge.

src/include/homestore/btree/detail/varlen_node.hpp
```diff
@@ -270,15 +270,15 @@ class VariableNode : public BtreeNode {
...
-            auto sz = other.insert(0, kb, vb); // Keep on inserting on the first index, thus moving everything to right
-            if (!sz) break;
-
-            --ind;
-            ++nmoved;
             if ((kb.size + vb.size + this->get_record_size()) > size_to_move) {
                 // We reached threshold of how much we could move
                 break;
             }
+
```
shosseinimotlagh, line 277: In the current implementation of insert, when there is no enough space for accommodating the new entity, an assert is triggered. To avoid hitting it, the check for available size needs to be performed in advance.

src/tests/btree_test_kvs.hpp
```diff
@@ -49,7 +49,8 @@ static std::string gen_random_string(size_t len, uint32_t preamble = std::numeri
     static thread_local std::random_device rd{};
     static thread_local std::default_random_engine re{rd()};
     std::uniform_int_distribution< size_t > rand_char{0, alphanum.size() - 1};
-    for (size_t i{0}; i < len; ++i) {
+    if (len < str.size()) { len = str.size(); }
```
shosseinimotlagh, line 52: `len + to_string(preamble).size()` must be always less than g_max_keysize

src/tests/btree_test_kvs.hpp
```diff
@@ -49,10 +49,10 @@ static std::string gen_random_string(size_t len, uint32_t preamble = std::numeri
...
+    for (size_t i{0}; i < len - str.size(); ++i) {
         str += alphanum[rand_char(re)];
     }
-    str += '\0';
```
shosseinimotlagh, line 55: not needed

src/tests/test_mem_btree.cpp
```diff
@@ -439,32 +439,6 @@ TYPED_TEST(BtreeTest, RangeUpdate) {
     this->query_validate(0, num_entries - 1, 75);
 }
 
-TYPED_TEST(BtreeTest, SimpleRemoveRange) {
```
shosseinimotlagh, line 442: a very primitive UT (not needed)

src/include/homestore/btree/detail/btree_remove_impl.ipp
```diff
@@ -465,26 +466,79 @@ btree_status_t Btree< K, V >::merge_nodes(const BtreeNodePtr& parent_node, const
         // Lets see if we have enough room in parent node to accommodate changes. This is needed only if the key is not
         // fixed length. For fixed length node merge will always result in lesser or equal size
         int64_t post_merge_size{0};
-        auto& old_node = old_nodes[leftmost_src.ith_nodes.back()];
-        if (old_node->total_entries()) {
-            post_merge_size += old_node->get_nth_obj_size(
-                std::min(leftmost_src.last_node_upto, old_node->total_entries() - 1)); // New leftmost entry
+        auto excess = old_nodes.size() - new_nodes.size();
```
shosseinimotlagh, line 469: It is important to note that the available size for the parent node may appear to be adequate at first, but can potentially lead to space shortages during the merging process itself. To prevent such shortages, it is necessary to estimate the available space in advance, just as it would occur during the actual merge. This would enable the calculation and avoidance of any potential violations that may arise during the merging process, such as those caused by insufficient space.
the step is a following:

> 1- release space  in the parent due to excess nodes
> 2- remove old last key and insert new last key (and check the available space for each iteration)
> 3- update the left most key

src/tests/test_mem_btree.cpp
hkadayam: I would suggest not to remove this UT. I made changes in the btree logic and this test is the only one which could find me the bug, rest passed, so we don't need to throw away this tests.

src/include/homestore/btree/detail/btree_remove_impl.ipp
shosseinimotlagh: To be more precise, it is the updated keys in the parent * max_key_size(). The number of updated keys are `new_nodes.size() + 1 `.  There is a possibility that the last key of all new nodes and also the last key of the leftmost node can be updated.

src/include/homestore/btree/detail/varlen_node.hpp
shosseinimotlagh: Screenshot 2023-10-10 at 10 57 12 AM

src/include/homestore/btree/detail/btree_remove_impl.ipp
shosseinimotlagh: Screenshot 2023-10-10 at 10 57 12 AM

src/include/homestore/btree/detail/btree_remove_impl.ipp
```diff
@@ -463,28 +464,25 @@ btree_status_t Btree< K, V >::merge_nodes(const BtreeNodePtr& parent_node, const
...
-        if (post_merge_size > parent_node->available_size(m_bt_cfg)) {
+        // fixed length. For fixed length node merge will always result in less or equal size
+
+        // we first calculate the least amount of space being released after removing excess children. the key size
+        // cannot be taken account; so we know for sure that value (i.e., linkinfo) and also its record will be freed.
+        // If the end_idx is the parent's edge, the space is not released eventually.
+        auto excess_releasing_nodes =
+            old_nodes.size() - new_nodes.size() - parent_node->total_entries() == end_idx ? 1 : 0;
+        auto minimum_releasing_excess_size =
+            excess_releasing_nodes * (BtreeLinkInfo::get_fixed_size() + parent_node->get_record_size());
```
hkadayam, line 475: This could be not be true, in case of different node types, we can't always assume node releases the entire record (say if the node has format where node are ref counted). My PR removes this assumption and ask the node to calculate that for us than btree apis to assume that. I think its ok to do it and modify as part of my PR.

hkadayam (dismissed):
Looks good to me, except the condition I mentioned in merge logic. As long as we keep that in my mind to do following my PR, I am ok with these changes.

sanebay (approved):

PR #183, opened 2023-09-25
ReplDev context type safety and blkread tracker race condition fix

hkadayam:
* Consumer passes some context as part of ReplDev requests, which are then passed back during callback. This was a type unsafe void* pointer. This resulted in either caller allocating a surrounding request structure then pass as a void* or worse caller pass a stack context. Latter is dangerous since callback can be done as part of different thread at different time. Former is safe, but it involves type unsafety and another allocation, since homestore repl layer also has to create a structure to capture its status. Hence this commit opens up a repl_ctx as intrusive pointer and exposes as the context for upper layer, providing type safety and reduce allocations. Using intrusive instead of shared_ptr is because intrusive size is 8 bytes and avoid another allocation during lambda capture, since most of the std::function implementation will capture upto 32 bytes inline and better to keep them always inline.

* Allow consumer to veto pre_commit on callback

* Without this fix, BlkReadTracker gets the blk_track_record and then makes decision to update or delete.  When this happens on multiple threads, each thread makes update or delete independently and step on each other resulting in either callback never happens or ref_count goes below 0. Fixed this by calling SimpleHashMap new api to get/update/delete atomically

### DIFF
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -1,17 +1,43 @@
 #pragma once
 
-// #include <nuraft_mesg/messaging_if.hpp>
+#include <boost/intrusive_ptr.hpp>
+#include <boost/smart_ptr/intrusive_ref_counter.hpp>
 #include <sisl/fds/buffer.hpp>
 
 #include <homestore/replication/repl_decls.h>
 
 namespace homestore {
+class ReplDev;
+
+struct repl_journal_entry;
+struct repl_req_ctx : public boost::intrusive_ref_counter< repl_req_ctx, boost::thread_safe_counter > {
+    friend class SoloReplDev;
+
+public:
+    virtual ~repl_req_ctx();
+
+private:
+    sisl::blob header;                          // User header
+    sisl::blob key;                             // Key to replicate
+    sisl::sg_list value;                        // Raw value - applicable only to leader req
+    MultiBlkId local_blkid;                     // List of corresponding local blkids for the value
+    RemoteBlkId remote_blkid;                   // List of remote blkid for the value
+    std::unique_ptr< uint8_t[] > journal_buf;   // Buf for the journal entry
+    repl_journal_entry* journal_entry{nullptr}; // pointer to the journal entry
+    int64_t lsn{0};                             // Lsn for this replication req
+
+    void alloc_journal_entry(uint32_t size);
+};
+
 //
 // Callbacks to be implemented by ReplDev users.
 //
 class ReplDevListener {
 public:
     virtual ~ReplDevListener() = default;
+
+    void set_repl_dev(ReplDev* rdev) { m_repl_dev = std::move(rdev); }
+    virtual ReplDev* repl_dev() { return m_repl_dev; }
 
     /// @brief Called when the log entry has been committed in the replica set.
     ///
@@ -22,10 +48,10 @@ public:
     /// @param header - Header originally passed with replica_set::write() api
     /// @param key - Key originally passed with replica_set::write() api
     /// @param blkids - List of blkids where data is written to the storage engine.
-    /// @param ctx - User contenxt passed as part of the replica_set::write() api
+    /// @param ctx - Context passed as part of the replica_set::write() api
     ///
     virtual void on_commit(int64_t lsn, sisl::blob const& header, sisl::blob const& key, MultiBlkId const& blkids,
-                           void* ctx) = 0;
+                           cintrusive< repl_req_ctx >& ctx) = 0;
 
     /// @brief Called when the log entry has been received by the replica dev.
     ///
@@ -44,8 +70,9 @@ public:
     /// @param lsn - The log sequence number
     /// @param header - Header originally passed with repl_dev::write() api
     /// @param key - Key originally passed with repl_dev::write() api
-    /// @param ctx - User contenxt passed as part of the repl_dev::write() api
-    virtual void on_pre_commit(int64_t lsn, const sisl::blob& header, const sisl::blob& key, void* ctx) = 0;
+    /// @param ctx - Context passed as part of the replica_set::write() api
+    virtual bool on_pre_commit(int64_t lsn, const sisl::blob& header, const sisl::blob& key,
+                               cintrusive< repl_req_ctx >& ctx) = 0;
 
     /// @brief Called when the log entry has been rolled back by the replica set.
     ///
@@ -59,8 +86,9 @@ public:
     /// @param lsn - The log sequence number getting rolled back
     /// @param header - Header originally passed with repl_dev::write() api
     /// @param key - Key originally passed with repl_dev::write() api
-    /// @param ctx - User contenxt passed as part of the repl_dev::write() api
-    virtual void on_rollback(int64_t lsn, const sisl::blob& header, const sisl::blob& key, void* ctx) = 0;
+    /// @param ctx - Context passed as part of the replica_set::write() api
+    virtual void on_rollback(int64_t lsn, const sisl::blob& header, const sisl::blob& key,
+                             cintrusive< repl_req_ctx >& ctx) = 0;
 
     /// @brief Called when replication module is trying to allocate a block to write the value
     ///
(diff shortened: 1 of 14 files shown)


src/tests/test_solo_repl_dev.cpp
```diff
@@ -97,19 +99,37 @@ struct Runner {
...
+        if ((actual_comp.fetch_add(1) + 1) >= expected_comp.load()) { comp_promise.setValue(); }
+    }
+};
+
+struct test_repl_req : public repl_req_ctx {
     sisl::byte_array header;
     sisl::byte_array key;
     sisl::sg_list write_sgs;
     sisl::sg_list read_sgs;
     int64_t lsn;
```
szmyd, line 125: There seems to be some redundant data between this and the base; should it just be public?

src/lib/blkdata_svc/blk_read_tracker.cpp
```diff
@@ -28,59 +28,39 @@ void BlkReadTracker::merge(const BlkId& blkid, int64_t new_ref_count,
...
-            if (rec_found) {
-                // if some read is already happening on this record, just update the ref_cnt;
+    while (cur_base_blk_num <= last_base_blk_num) {
+        BlkId base_blkid{cur_base_blk_num, entries_per_record(), blkid.chunk_num()};
+
+        if (new_ref_count > 0) {
+            m_pending_reads_map.upsert_or_delete(base_blkid,
+                                                 [&base_blkid, new_ref_count](BlkTrackRecord& rec, bool existing) {
+                                                     if (!existing) { rec.m_key = base_blkid; }
+                                                     rec.m_ref_cnt += new_ref_count;
```
yamingk, line 44: nit: I knwn `m_ref_cnt` is defaulted to zero, would suggest set the ref_cnt if not exist for readability. 
```if (!existing) { 
    rec.m_key = base_blkid;
    rec.m_ref_cnt = new_ref_cnt;
}  else { rec.m_ref_cnt += new_ref_cnt; }```

File: src/lib/blkdata_svc/blk_read_tracker.cpp [hunk 3]
```diff
@@ -28,59 +28,39 @@ void BlkReadTracker::merge(const BlkId& blkid, int64_t new_ref_count,
...
-                BlkTrackRecord dummy_rec;
-                m_pending_reads_map.erase(base_blkid, dummy_rec);
-            }
+                return (rec.m_ref_cnt == 0);
+            });
         } else {
-            // this is wait_on operation
-            if (rec_found) {
-                // apply waiter to this record;
+            m_pending_reads_map.update(base_blkid, [&waiter_rescheduled, &waiter](BlkTrackRecord& rec) {
```
yamingk, line 54: Can we put the comment back to indicate this is a wait_on operation?

src/lib/blkdata_svc/blk_read_tracker.cpp
```diff
@@ -28,59 +28,39 @@ void BlkReadTracker::merge(const BlkId& blkid, int64_t new_ref_count,
...
-
-        if (new_ref_count != 0) {
-            // this is insert/remove operations
-            if (rec_found) {
-                // if some read is already happening on this record, just update the ref_cnt;
+    while (cur_base_blk_num <= last_base_blk_num) {
+        BlkId base_blkid{cur_base_blk_num, entries_per_record(), blkid.chunk_num()};
+
+        if (new_ref_count > 0) {
+            m_pending_reads_map.upsert_or_delete(base_blkid,
```
yamingk, line 41: comment: // this is an insert operation

src/lib/blkdata_svc/blk_read_tracker.cpp
```diff
@@ -28,59 +28,39 @@ void BlkReadTracker::merge(const BlkId& blkid, int64_t new_ref_count,
...
-                m_pending_reads_map.upsert(base_blkid, rec);
                 waiter_rescheduled = true;
-            }
-
-            // not found, nothing needs to be done; fall through and visit remaining records;
+            });
         }
 
-        cur_blk_num_aligned += entries_per_record();
+        /*LOGDEBUG("[{}] {} rec_found={} rec_blkid=[{}] cur_rec_ref_cnt={}", base_blkid.to_string(),
```
yamingk, line 60: want to keep it or remove it? I prefer to keep it.

src/lib/replication/service/repl_service_impl.cpp
```diff
@@ -19,19 +19,26 @@
...
 namespace homestore {
-ReplicationServiceImpl& repl_service() { return hs()->repl_service(); }
+ReplicationService& repl_service() { return hs()->repl_service(); }
 
-ReplicationServiceImpl::ReplicationServiceImpl(repl_impl_type impl_type, std::unique_ptr< ReplServiceCallbacks > cbs) :
-        m_svc_cbs{std::move(cbs)}, m_repl_type{impl_type} {
+ReplicationServiceImpl::ReplicationServiceImpl(repl_impl_type impl_type) : m_repl_type{impl_type} {
     meta_service().register_handler(
-        "replication",
+        "repl_dev",
```
yamingk, line 26: Should we add group_id along with "repl_dev" for uniqueness of this metablk client?

src/tests/test_solo_repl_dev.cpp
hkadayam: this is just in test code, but yes I should remove this lsn and use base class lsn.

src/include/homestore/replication/repl_dev.h
```diff
@@ -1,18 +1,44 @@
...
 
 namespace homestore {
+class ReplDev;
+
+struct repl_journal_entry;
+struct repl_req_ctx : public boost::intrusive_ref_counter< repl_req_ctx, boost::thread_safe_counter > {
+    friend class SoloReplDev;
+
+public:
+    virtual ~repl_req_ctx();
```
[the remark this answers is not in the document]
hkadayam: Technically no, its not needed. repl layer will not destroy itself, but slightly concerned that every upper layer needs to ensure they type cast it correctly to the one originally created and not something else. If you think that skepticism is not worth it, I can remove vtable.

src/lib/blkdata_svc/blk_read_tracker.cpp
hkadayam: Done

src/lib/blkdata_svc/blk_read_tracker.cpp
hkadayam: done

src/lib/blkdata_svc/blk_read_tracker.cpp
```diff
@@ -28,59 +28,39 @@ void BlkReadTracker::merge(const BlkId& blkid, int64_t new_ref_count,
...
+
+        if (new_ref_count > 0) {
+            m_pending_reads_map.upsert_or_delete(base_blkid,
+                                                 [&base_blkid, new_ref_count](BlkTrackRecord& rec, bool existing) {
+                                                     if (!existing) { rec.m_key = base_blkid; }
+                                                     rec.m_ref_cnt += new_ref_count;
+                                                     return false;
+                                                 });
+        } else if (new_ref_count < 0) {
+            m_pending_reads_map.upsert_or_delete(base_blkid, [new_ref_count](BlkTrackRecord& rec, bool existing) {
```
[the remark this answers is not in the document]
hkadayam: done

src/lib/blkdata_svc/blk_read_tracker.cpp
hkadayam: put that as trace

src/include/homestore/replication/repl_dev.h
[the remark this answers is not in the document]
szmyd: I assume consumers of Homestore will have bugs; if they misuse this class because they didn't read it; is that worse than other bugs they could have?

szmyd (dismissed):
lgtm

yamingk (dismissed):

src/include/homestore/replication/repl_dev.h
```diff
@@ -98,10 +130,10 @@ class ReplDev {
     /// cases
     /// @param value - vector of io buffers that contain value for the key. It is an optional field and if the value
     /// list size is 0, then only key is written to replicadev without data.
-    /// @param user_ctx - User supplied opaque context which will be passed to listener
+    /// @param ctx - User supplied context which will be passed to listener
     /// callbacks
     virtual void async_alloc_write(sisl::blob const& header, sisl::blob const& key, sisl::sg_list const& value,
-                                   void* user_ctx) = 0;
+                                   intrusive< repl_req_ctx > ctx) = 0;
```
zichanglai, line 136: do you mind add a interface: get_blk_size() to ReplicationService,  because value size must be multiples of blk size , but for upper layer(HO) does not what is blk size at all. HS datasvc has such a interface, but it will be strange to call datasvc get_blk_size, it will be better to call replsvc get_blk_size().
it is a small change and I am both ok to add it in this pr or use another pr.

hkadayam: Make sense, will add as part of this PR itself.

Issue #184, opened 2023-09-25
BlkReadTracker test to have multithreaded test


hkadayam:
BlkReadTracker needs to have multithreaded tests. It is currently being tested by outside classes hitting those patterns, but an inline test is important enhancement.

PR #190, opened 2023-09-27
issue #184: threaded test for readblktracker

yamingk:
Added three new test cases:
========================
1. concurrent insert and concurrent remove.   // >>>>> to expose race window for concurrent get element, update element and then put back to map, same thing for remove.
2. concurrent insert, then concurrent wait, then concurrent remove. //  >>>> similar as "1", but add wait on operations.
3. concurrent insert/wait/remove.  // >>>>> this is pure concurrent operations close to real life concurrency.
All test passed.

### DIFF
--- a/src/tests/test_blk_read_tracker.cpp
+++ b/src/tests/test_blk_read_tracker.cpp
@@ -408,7 +408,7 @@ TEST_F(BlkReadTrackerTest, TestThreadedInsertWaitonThenRemove) {
                 LOGINFO("wait_on called on blkid: {};", bids[i].to_string());
             });
         });
-        t.detach();
+        op_threads.push_back(std::move(t));
     }
 
     LOGINFO("Step 3: threaded wait_on issued on all bids.");
--- a/src/tests/test_blk_read_tracker.cpp
+++ b/src/tests/test_blk_read_tracker.cpp
@@ -403,7 +403,7 @@ TEST_F(BlkReadTrackerTest, TestThreadedInsertWaitonThenRemove) {
         std::thread t([this, &bids, i, &mtx, &called]() {
             get_inst()->wait_on(bids[i], [i, &bids, &mtx, &called]() {
                 std::unique_lock lk(mtx);
-                assert(!called[i]); // callback shouldn't be called more than once;
+                LOGMSG_ASSERT(called[i] == false, "not expecting callback to be called more than once!");
                 called[i] = true;
                 LOGINFO("wait_on called on blkid: {};", bids[i].to_string());
             });
@@ -427,8 +427,8 @@ TEST_F(BlkReadTrackerTest, TestThreadedInsertWaitonThenRemove) {
 
     LOGINFO("Step 4: all threads joined.");
 
-    for (const auto x : called) {
-        assert(x); // all callbacks should be called;
+    for (auto i = 0ul; i < called.size(); ++i) {
+        LOGMSG_ASSERT(called[i] == true, "expecting all waiters to be called");
     }
 
     LOGINFO("Step 5: all bids wait_on cb called.");
--- a/src/tests/test_blk_read_tracker.cpp
+++ b/src/tests/test_blk_read_tracker.cpp
@@ -374,9 +374,9 @@ TEST_F(BlkReadTrackerTest, TestThreadedInsertAndRemove) {
     LOGINFO("Step 3: threaded insert joined.");
     op_threads.clear();
 
-    for (auto i = 0ul; i < bids.size(); ++i) {
+    for (const auto& b : bids) {
         for (auto j = 0ul; j < repeat; ++j) {
-            std::thread t([this, &bids, i]() { get_inst()->remove(bids[i]); });
+            std::thread t([this, &b]() { get_inst()->remove(b); });
             op_threads.push_back(std::move(t));
         }
     }
@@ -397,10 +397,10 @@ TEST_F(BlkReadTrackerTest, TestThreadedInsertWaitonThenRemove) {
     std::vector< BlkId > bids{{12, 6, 0}, {18, 5, 0}, {25, 8, 0}, {36, 16, 0}, {57, 4, 0}, {66, 18, 0}, {92, 14, 0}};
     const auto repeat = 100ul;
     std::vector< std::thread > op_threads;
-    for (auto i = 0ul; i < bids.size(); ++i) {
-        std::thread t([this, &bids, i]() {
+    for (const auto& b : bids) {
+        std::thread t([this, &b]() {
             for (auto j = 0ul; j < repeat; ++j) {
-                get_inst()->insert(bids[i]);
+                get_inst()->insert(b);
             }
         });
         op_threads.push_back(std::move(t));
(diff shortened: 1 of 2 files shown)


src/tests/test_blk_read_tracker.cpp
```diff
@@ -352,6 +334,106 @@ TEST_F(BlkReadTrackerTest, TestInsRmWithWaiterOverlapMultiReads2) {
...
+    LOGINFO("Step 1: set entries per record to {}.", align);
+    get_inst()->set_entries_per_record(align);
+
+    std::vector< BlkId > bids{{10, 8, 0}, {20, 5, 0}, {25, 6, 0}, {43, 16, 0}, {56, 4, 0}, {72, 18, 0}, {122, 4, 0}};
+
+    const auto repeat = 100ul;
+    std::vector< std::thread > op_threads;
+    for (auto i = 0ul; i < bids.size(); ++i) {
+        std::thread t([this, &bids, i]() {
+            for (auto j = 0ul; j < repeat; ++j) {
```
yamingk, line 357: this loop is a intentional inside thread (to create less threads), compared to remove operation outside of thread

src/tests/test_blk_read_tracker.cpp
```diff
@@ -352,6 +334,124 @@ TEST_F(BlkReadTrackerTest, TestInsRmWithWaiterOverlapMultiReads2) {
...
+    LOGINFO("Step 1: set entries per record to {}.", align);
+    get_inst()->set_entries_per_record(align);
+
+    std::vector< BlkId > bids{{10, 8, 0}, {20, 5, 0}, {25, 6, 0}, {43, 16, 0}, {56, 4, 0}, {72, 18, 0}, {122, 4, 0}};
+
+    const auto repeat = 100ul;
+    std::vector< std::thread > op_threads;
+
+    for (auto i = 0ul; i < bids.size(); ++i) {
+        std::thread t([this, &bids, i]() {
```
hkadayam, line 357: Nit: You don't need to iterate with old fashioned bids[i] and can use
```for (auto const& bid: bids) {..}```
that way lambda doesn't need to capture bids and i, just bid by value also.

Comment (yamingk) [reply src/tests/test_blk_read_tracker.cpp:357]: done

File: src/tests/test_blk_read_tracker.cpp [hunk 3]
```diff
@@ -352,6 +358,198 @@ TEST_F(BlkReadTrackerTest, TestInsRmWithWaiterOverlapMultiReads2) {
...
+    get_inst()->set_entries_per_record(align);
+
+    std::vector< BlkId > bids{{10, 8, 0}, {20, 5, 0}, {25, 6, 0}, {43, 16, 0}, {56, 4, 0}, {72, 18, 0}, {122, 4, 0}};
+
+    const auto repeat = 100ul;
+    std::vector< std::thread > op_threads;
+
+    for (const auto& b : bids) {
+        std::thread t([this, &b]() {
+            for (auto j = 0ul; j < repeat; ++j) {
```
sanebay, line 381: repeat is simulating multiple reads ?

src/tests/test_blk_read_tracker.cpp
```diff
@@ -352,6 +358,198 @@ TEST_F(BlkReadTrackerTest, TestInsRmWithWaiterOverlapMultiReads2) {
...
+    }
+
+    LOGINFO("Step 2: threaded insert issued.");
+
+    for (auto& t : op_threads) {
+        t.join();
+    }
+
+    // remove has to wait for insert to complete because if insert thread runs slower than remove, it might assert
+    // complaining no elements are found in map;
```
sanebay, line 395: Couldnt we solve this with insert at line 382 also insert to a set and remove thread uses that as a candidate.

src/tests/test_blk_read_tracker.cpp
```diff
@@ -352,6 +358,198 @@ TEST_F(BlkReadTrackerTest, TestInsRmWithWaiterOverlapMultiReads2) {
...
+    LOGINFO("Step 1: set entries per record to {}.", align);
+    get_inst()->set_entries_per_record(align);
+
+    std::vector< BlkId > bids{{10, 8, 0}, {20, 5, 0}, {25, 6, 0}, {43, 16, 0}, {56, 4, 0}, {72, 18, 0}, {122, 4, 0}};
+
+    const auto repeat = 100ul;
+    std::vector< std::thread > op_threads;
+
+    for (const auto& b : bids) {
+        std::thread t([this, &b]() {
```
sanebay, line 380: using reference in lambda capture will cause it to reuse old objects. Capture by value. Same in other places.

src/tests/test_blk_read_tracker.cpp
```diff
@@ -352,6 +358,198 @@ TEST_F(BlkReadTrackerTest, TestInsRmWithWaiterOverlapMultiReads2) {
...
+        });
+        op_threads.push_back(std::move(t));
+    }
+
+    // wait for all insert to complete, otherwise it will race with remove thread;
+    for (auto& t : op_threads) {
+        t.join();
+    }
+
+    op_threads.clear();
```
sanebay, line 453: Is it safe to clear op_threads as we are using it in line 445 for wait_on ? May be separate vec for wait on ?

src/tests/test_blk_read_tracker.cpp
```diff
@@ -352,6 +358,198 @@ TEST_F(BlkReadTrackerTest, TestInsRmWithWaiterOverlapMultiReads2) {
...
+    std::mutex mtx;
+    std::list< BlkId > inserted_bids;
+
+    std::atomic< uint32_t > outstanding_wait_bids_cnt = 0ul;
+
+    LOGINFO("Step 2: randome threaded insert/remove/wait_on operation:");
+    std::list< std::thread > op_threads;
+    auto nitr = 0ul;
+    while (nitr++ < 200ul || inserted_bids.empty() == false) {
+        std::thread t([this, &outstanding_wait_bids_cnt, &nitr, &inserted_bids, &mtx]() {
```
sanebay, line 497: Is it more stress testing to create 200 threads and each doing one operation vs creating N threads and each doing M iteration of random operations

src/tests/test_blk_read_tracker.cpp
yamingk: it is done at the last test case. This test case is not intended for that.

src/tests/test_blk_read_tracker.cpp
yamingk: it won't clear until all threads are joined, right?

src/tests/test_blk_read_tracker.cpp
yamingk: the purpose of the test is to create 200 threads, doing more iterations in a single thread won't help create racing window.

src/tests/test_blk_read_tracker.cpp
yamingk: repeat here with insert is just to put more ref_cnt on same record. Not simulating multiple threads

src/tests/test_blk_read_tracker.cpp
yamingk: it is a ref to different element in bids, it will use capture different elements.

src/tests/test_blk_read_tracker.cpp
sanebay: Shouldnt remove call below cause the wait_on() callbacks to get triggered ? But if we clear here, is it still ok ?

yamingk: it is okay. The wait_on thread itself is just to simulate a free blk call, and once applied waiter, free operation thread was gone. The wait_on lambda is supposed to be called by remove operation's thread, which in real life is a read_completion routine's thread. I hope this explains.

sanebay (approved):

PR #194, opened 2023-09-28
Add cp flush during shutdown in index btree

shosseinimotlagh:


src/lib/index/index_service.cpp
```diff
@@ -75,11 +75,29 @@ void IndexService::start() {
...
 }
 
+void IndexService::stop() {
+    std::unique_lock lg(m_index_map_mtx);
+    auto fut = homestore::hs()->cp_mgr().trigger_cp_flush(true /* force */);
+    auto on_complete = [&](auto success) {
+        HS_REL_ASSERT_EQ(success, true, "CP Flush failed");
+        LOGINFO("CP Flush completed");
+    };
+    on_complete(std::move(fut).get());
```
hkadayam, line 85: What is the purpose of this lamba, can we not just have statements 82-83 here itself?

hkadayam (approved):

sanebay (approved):

PR #198, opened 2023-10-07
fix chunk_size of virtual device may not be aligned with blk_size issue

zichanglai:


hkadayam (approved):

yamingk:
@zichanglai can you elaborate a little more why this can only be caught by HO test case, not HomeStore. We might need a new test case in HomeStore to catch that.

zichanglai:
> @zichanglai can you elaborate a little more why this can only be caught by HO test case, not HomeStore. We might need a new test case in HomeStore to catch that.

@yamingk , I think it is because HO use a specified and rare-use chunk num 65000,  in HO tests case,   HS data service will use 80% percent of HS space, which is actually 1717986919 bytes and block size is 1024. so :

1.  vdev size will be round-up to 1730560000 first 
2. chunk num is adjust from 65000 to 128.
3. re-calculate the chunk size = 1730560000 / 128 = 13520000.0,  but it is not aligned with 1024 any more.

I think HS can added such case to make sure future changes will not  cause same issue again.

PR #199, opened 2023-10-07
fix solo repldev will be nullptr in map when homestore restart

zichanglai:
This PR fix issue when HS restarts,  the Solo ReplDev in map will be set to nullptr. it is caused by an unexpected std::move() operation in src/lib/replication/service/repl_service_impl.cpp Line80

### DIFF
--- a/src/lib/replication/service/repl_service_impl.cpp
+++ b/src/lib/replication/service/repl_service_impl.cpp
@@ -74,7 +74,7 @@ ReplicationServiceImpl::open_repl_dev(uuid_t group_id, std::unique_ptr< ReplDevL
     auto it = m_rd_map.find(group_id);
     if (it != m_rd_map.end()) {
         // We already loaded the ReplDev, just call the group_id and attach the listener
-        auto& repl_dev = it->second;
+        auto repl_dev = it->second;
         listener->set_repl_dev(repl_dev.get());
         repl_dev->attach_listener(std::move(listener));
         return make_async_success< shared< ReplDev > >(std::move(repl_dev));
--- a/conanfile.py
+++ b/conanfile.py
@@ -5,7 +5,7 @@ from conans import CMake
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "4.5.4"
+    version = "4.5.5"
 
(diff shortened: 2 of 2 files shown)


src/lib/replication/service/repl_service_impl.cpp
```diff
@@ -52,7 +52,6 @@ ReplicationServiceImpl::create_repl_dev(uuid_t group_id, std::set< std::string,
     superblk< repl_dev_superblk > rd_sb{"repl_dev"};
     rd_sb.create(sizeof(repl_dev_superblk));
     rd_sb->gid = group_id;
-
```
hkadayam, line 55: Nit: I would like to avoid contiguous pieces of lines of code and would rather have this new line. It gives a clear block of code.

src/lib/replication/service/repl_service_impl.cpp
```diff
@@ -74,7 +73,7 @@ ReplicationServiceImpl::open_repl_dev(uuid_t group_id, std::unique_ptr< ReplDevL
     auto it = m_rd_map.find(group_id);
     if (it != m_rd_map.end()) {
         // We already loaded the ReplDev, just call the group_id and attach the listener
-        auto& repl_dev = it->second;
+        auto repl_dev = it->second;
```
hkadayam, line 76: I am not clear how is this fixes the problem. At this time, someone already opened the repl_dev and group_id -> repl_dev has been mapped. So how is this storing by referencing or incrementing the ref_count of repl_dev solves this?

src/lib/replication/service/repl_service_impl.cpp
zichanglai: updated

src/lib/replication/service/repl_service_impl.cpp
zichanglai: but at Line80, this repl_dev will be moved, that is why I changed to make a copy of shared_ptr

xiaoxichen (approved):
Approving this as it is clear, 
repl_dev was by-reference referencing it->second , later in L80 repl_dev has been std::move() which will set to repl_dev to null and as it is by-reference, it->second be set to null

PR #200, opened 2023-10-09
recover metablk in subtype order

JacksonYao287:
In HO , when metaservice restart,  we actually want to recover repl-dev before PGinfo  and recover PGinfo before Shardinfo. 
but metaservice will recover all the metablk only according to blkid and we have to use some pending method to solve this problem.
https://github.com/eBay/HomeObject/blob/dbde1bb6b2eb705fc55f7539be08d9c9fdbbd508/src/lib/homestore_backend/hs_pg_manager.cpp#L126

https://github.com/eBay/HomeObject/blob/dbde1bb6b2eb705fc55f7539be08d9c9fdbbd508/src/lib/homestore_backend/hs_homeobject.cpp#L105

this PR do some change in metaservice recovery logic to make sure all the metablk will be recovered by subtype, and the subtype order is that appear order in the metablk chain.

### DIFF
--- a/src/include/homestore/meta_service.hpp
+++ b/src/include/homestore/meta_service.hpp
@@ -18,11 +18,13 @@
 #include <cstdint>
 #include <functional>
 #include <map>
+#include <unordered_map>
 #include <memory>
 #include <mutex>
 #include <string>
 #include <system_error>
 #include <vector>
+#include <optional>
 
 #include <sisl/fds/buffer.hpp>
 #include <sisl/metrics/metrics.hpp>
@@ -46,10 +48,12 @@ struct vdev_info;
 // new blk found subsystem callback
 typedef std::function< void(meta_blk* mblk, sisl::byte_view buf, size_t size) > meta_blk_found_cb_t;
 typedef std::string meta_sub_type;
+typedef std::vector< meta_sub_type > meta_subtype_vec_t;
 typedef std::function< void(bool success) > meta_blk_recover_comp_cb_t; // recover complete subsystem callbacks;
 typedef std::map< uint64_t, meta_blk* > meta_blk_map_t;                 // blkid to meta_blk map;
 typedef std::map< uint64_t, meta_blk_ovf_hdr* > ovf_hdr_map_t;          // ovf_blkid to ovf_blk_hdr map;
 typedef std::map< meta_sub_type, MetaSubRegInfo > client_info_map_t;    // client information map;
+typedef std::unordered_map< meta_sub_type, std::vector< meta_sub_type > > subtype_graph_t;
 
 class MetablkMetrics : public sisl::MetricsGroupWrapper {
 public:
@@ -87,6 +91,7 @@ private:
     MetablkMetrics m_metrics;
     bool m_inited{false};
     std::unique_ptr< meta_vdev_context > m_meta_vdev_context;
+    subtype_graph_t m_dep_topo_graph;
 
 public:
     MetaBlkService(const char* name = "MetaBlkStore");
@@ -121,7 +126,7 @@ public:
      * @param cb : subsystem cb
      */
     void register_handler(meta_sub_type type, const meta_blk_found_cb_t& cb, const meta_blk_recover_comp_cb_t& comp_cb,
-                          bool do_crc = true);
+                          bool do_crc = true, std::optional< meta_subtype_vec_t > deps = std::nullopt);
 
     /**
      * @brief
@@ -356,6 +361,8 @@ private:
 
     bool scan_and_load_meta_blks(meta_blk_map_t& meta_blks, ovf_hdr_map_t& ovf_blk_hdrs, BlkId* last_mblk_id,
                                  client_info_map_t& sub_info);
+
+    void recover_meta_block(meta_blk* meta_block);
 
 public:
     bool verify_metablk_store();
--- a/src/include/homestore/meta_service.hpp
+++ b/src/include/homestore/meta_service.hpp
@@ -363,6 +363,7 @@ private:
                                  client_info_map_t& sub_info);
 
     void recover_meta_block(meta_blk* meta_block);
+    void recover_meta_sub_type(bool do_comp_cb, const meta_sub_type&);
 
 public:
     bool verify_metablk_store();
--- a/src/lib/meta/meta_blk_service.cpp
+++ b/src/lib/meta/meta_blk_service.cpp
@@ -377,10 +377,11 @@ void MetaBlkService::register_handler(meta_sub_type type, const meta_blk_found_c
     m_sub_info[type].cb = cb;
     m_sub_info[type].comp_cb = comp_cb;
     m_sub_info[type].do_crc = do_crc ? 1 : 0;
-    m_dep_topo_graph[type];
     if (deps.has_value()) {
-        for (const auto& dep : deps.value()) {
-            m_dep_topo_graph[dep].push_back(type);
+        m_sub_info[type].has_deps = true;
+        for (auto const& x : deps.value()) {
+            m_sub_info[x].has_deps = true;
+            m_dep_topo_graph[x].push_back(type);
         }
     }
 }
@@ -1081,11 +1082,9 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
 void MetaBlkService::recover(bool do_comp_cb) {
     // for each registered subsystem, look up in cache for their meta blks;
     std::lock_guard< decltype(m_shutdown_mtx) > lg{m_shutdown_mtx};
-
-    meta_subtype_vec_t independent_subtypes;
     meta_subtype_vec_t ordered_subtypes;
 
-    if (hs_utils::topological_sort(m_dep_topo_graph, ordered_subtypes, &independent_subtypes)) {
+    if (hs_utils::topological_sort(m_dep_topo_graph, ordered_subtypes)) {
         throw homestore::homestore_exception(
             "MetaBlkService has circular dependency, please check the dependency graph", homestore_error::init_failed);
     }
@@ -1093,28 +1092,25 @@ void MetaBlkService::recover(bool do_comp_cb) {
     // all the subsystems are divided into two parts.
     // for subsystems in ordered_subtypes, we need to recover in order.
     for (const auto& subtype : ordered_subtypes) {
-        for (const auto& m : m_sub_info[subtype].meta_bids) {
-            auto mblk = m_meta_blks[m];
-            recover_meta_block(mblk);
-        }
-
-        if (do_comp_cb && m_sub_info[subtype].comp_cb) {
-            m_sub_info[subtype].comp_cb(true);
-            HS_LOG(DEBUG, metablk, "[type={}] completion callback sent.", subtype);
-        }
+        recover_meta_sub_type(do_comp_cb, subtype);
     }
 
-    // TODO: for subsystems in independent_subtypes, we can use concurrent recovery if necessary.
-    for (const auto& subtype : independent_subtypes) {
-        for (const auto& m : m_sub_info[subtype].meta_bids) {
-            auto mblk = m_meta_blks[m];
-            recover_meta_block(mblk);
-        }
+    // TODO: for independent subsystems, we can use concurrent recovery if necessary.
+    for (auto const& x : m_sub_info) {
+        auto& reg_info = x.second;
+        if (!reg_info.has_deps) { recover_meta_sub_type(do_comp_cb, x.first); }
+    }
+}
 
-        if (do_comp_cb && m_sub_info[subtype].comp_cb) {
-            m_sub_info[subtype].comp_cb(true);
-            HS_LOG(DEBUG, metablk, "[type={}] completion callback sent.", subtype);
-        }
+void MetaBlkService::recover_meta_sub_type(bool do_comp_cb, const meta_sub_type& sub_type) {
+    for (const auto& m : m_sub_info[sub_type].meta_bids) {
+        auto mblk = m_meta_blks[m];
+        recover_meta_block(mblk);
+    }
+
+    if (do_comp_cb && m_sub_info[sub_type].comp_cb) {
+        m_sub_info[sub_type].comp_cb(true);
+        HS_LOG(DEBUG, metablk, "[type={}] completion callback sent.", sub_type);
     }
 }
 
--- a/src/lib/meta/meta_blk_service.cpp
+++ b/src/lib/meta/meta_blk_service.cpp
@@ -351,6 +351,7 @@ void MetaBlkService::deregister_handler(meta_sub_type type) {
     const auto it = m_sub_info.find(type);
     if (it != std::end(m_sub_info)) {
         m_sub_info.erase(it);
+        m_dep_topo_graph.erase(type);
         HS_LOG(INFO, metablk, "[type={}] deregistered Successfully", type);
     } else {
         HS_LOG(INFO, metablk, "[type={}] not found in registered list, no-op", type);
@@ -358,7 +359,8 @@ void MetaBlkService::deregister_handler(meta_sub_type type) {
 }
 
 void MetaBlkService::register_handler(meta_sub_type type, const meta_blk_found_cb_t& cb,
-                                      const meta_blk_recover_comp_cb_t& comp_cb, bool do_crc) {
+                                      const meta_blk_recover_comp_cb_t& comp_cb, bool do_crc,
+                                      std::optional< meta_subtype_vec_t > deps) {
     std::lock_guard< decltype(m_meta_mtx) > lk(m_meta_mtx);
     HS_REL_ASSERT_LT(type.length(), MAX_SUBSYS_TYPE_LEN, "type len: {} should not exceed len: {}", type.length(),
                      MAX_SUBSYS_TYPE_LEN);
@@ -375,7 +377,12 @@ void MetaBlkService::register_handler(meta_sub_type type, const meta_blk_found_c
     m_sub_info[type].cb = cb;
     m_sub_info[type].comp_cb = comp_cb;
     m_sub_info[type].do_crc = do_crc ? 1 : 0;
-    HS_LOG(INFO, metablk, "[type={}] registered with do_crc: {}", type, do_crc);
+    m_dep_topo_graph[type];
+    if (deps.has_value()) {
+        for (const auto& dep : deps.value()) {
+            m_dep_topo_graph[dep].push_back(type);
+        }
+    }
 }
 
 void MetaBlkService::add_sub_sb(meta_sub_type type, const uint8_t* context_data, uint64_t sz, void*& cookie) {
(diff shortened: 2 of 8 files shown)


JacksonYao287:
after this PR is merge , we can change the recovery logic [ in this patch](https://github.com/eBay/HomeObject/pull/68) to a simple way

src/lib/meta/meta_blk_service.cpp
```diff
@@ -1074,83 +1080,81 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
...
-                const auto crc = crc32_ieee(init_crc32, s_cast< const uint8_t* >(buf->bytes), mblk->hdr.h.context_sz);
-
-                HS_REL_ASSERT_EQ(crc, uint32_cast(mblk->hdr.h.crc),
-                                 "[type={}], CRC mismatch: {}/{}, on mblk bid: {}, context_sz: {}", mblk->hdr.h.type,
-                                 crc, uint32_cast(mblk->hdr.h.crc), mblk->hdr.h.bid.to_string(),
-                                 uint64_cast(mblk->hdr.h.context_sz));
-            } else {
-                HS_LOG(DEBUG, metablk, "[type={}] meta blk found with bypassing crc.", mblk->hdr.h.type);
-            }
+    for (const auto& subtype : m_subtype_recovery_order) {
```
xiaoxichen, line 1083: Agree to recovery by subtype however doesnt understand why you think 1 is better than 2
1 . subtype order based on appear order in the metablk chain
2. subtype order based on the order of subtype_name(string).

As 2 by nature can be done by something like
```
for (const auto& sub : m_sub_info)
    for (const auto& sub : sub->second().meta_bids)
```

yamingk (dismissed):

src/lib/meta/meta_blk_service.cpp
xiaoxichen: And also not sure how this could help HO , i.e how can we achieve
```
we actually want to recover repl-dev before PGinfo and recover PGinfo before Shardinfo.
```
It seems to me neither  `metablk chain` order nor `subtype dictionary order` is the order we want (dependency order),  unless we explicitly pass-in recovery-order(from hard-code or config).

JacksonYao287: >why you think 1 is better than 2

suppose we have two subsystems in metaservice , the name is `A` and `B`, respectively.
if we add a metablk B1, then add a metablk A1, then A2, then B2. finally,  the metablk chain will look like
`ssb->B1->A1->A2->B2`

the first metablk of subsystem B (B1) comes before the first metablk of subsystem A (A1), so we want to recover all the metablks of B before A

`client_info_map_t m_sub_info ` is a `std::map` 
```
std::map< meta_sub_type, MetaSubRegInfo > client_info_map_t;
```
and, all the items in it will be ordered by the key(A<B). so if we use `for (const auto& sub : m_sub_info)` as you mentioned, subsystem A will be recovered before subsystem B, that is not what we want here.

>unless we explicitly pass-in recovery-order

as i mentioned above, actually the recovery order of all the subsystem here will be the order of the presence of the first metablk of each subsystem in the metablk chain. so if shardInfo depends on PGinfo, PGinfo will be created before shardInfo and thus the metablk of PGinfo will be previous of shardinfo, and then the subsystem of PG will be recovered before shard.

pass-in recovery-order in unnecessary here. if we do this , the user of homestore will have to take the responsibility to pass in the correct order, which is not what we want. So , unless we want to recovery shardInfo before PGInfo, for example, we need a mechanism to specify the particular order of subsystem recovery. but i do not think there will be a scenario like this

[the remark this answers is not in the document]
JacksonYao287: > MetaService as a key-value provider shouldn't udnerstand each component's recovery logic.

yea , it is. and here, we just want metaservice to recover according to the order of the presence of the first metablk of each subsystem in the metablk chain. it also does not udnerstand each component's recovery logic.

JacksonYao287:
I think it is better to clarify this PR again.

**1 what is the current metaservice recovery process?** 

the current recovery process of metaservice is recovering all the metablks according to BlkID, and know nothing about the upper layer logic. so from the perspective of upper layer , no recovery order guaranteeed.

for the upper layer user, they should adapt the out-of-order of metaservice recovery and implement the upper layer recovery logic carefully by themselves. that is it.

**2 what does this PR do?** 

this PR changes the default recovery order of metaservice from out-of-order to the order of  the presence of  subsystem in the metablk chain. the presence of subsystem means the presence of the first metablk of the subsystem in the metablk chain.  for example , suppose we have 3  subsystems A, B and C , and in each subsystem we have 3 metablk( A1, A2 ,A3 ),(B1,B2,B3),(C1,C2,C3). the matablk chain is as follows:
`A1->C1->C2->A2->B1->C3->B2->A3->B3`
current recovery order will be ordered by the blkid among all the subsystems, which is not definite. 

since A1 comes before C1 and C1 comes before B1, so subsystem A comes before C and C come before B. after the PR, the recovery order will be
`(A1,A2,A3) -> (C1,C2,C3) ->(B1,B2,B3)`
in each subsystem, it will be ordered by blkid. But among subsystems, it will be ordered by the presence of subsystem

this gives the upper layer a chance to untilize this recovery order to simplify the upper layer recovery logic. well , the upper lay can also ignore this change and do like before. note that , here, the upper layer can not control the order of metaservice recovery, and we also do not want to give the chance to upper layer to control until we really need. 

so , in fact , what this PR does is to give a default natural order of metaservice recovery, and no extra performance issue will be involved.

**3 is it necessary for the upper layer to control the recover order of metaservice in fine granularity** 

i don`t think it is necessary for now at least. the reason is :
1 upper layer user  will take the responsibility to pass in the correct order.
2 inside homestore, there are already some metaservice user, such as CPmanager. how do we handle the recovery order of these subsystem with those passed in subsystem. although this will not be difficult to choose a policy and implement, but it seems not uniformed that the upper layer can only partially control the recovery order. they can not control the order of those existed subsystem in the homestore.

so my suggestion is we can delay to implement fine-granularity control of the metaservice recovery order until we have a clear scenario . for now, we can just use the newly added default recovery order to simplify HO Implementation

xiaoxichen:
another example of ordering requirement 
https://github.com/eBay/HomeObject/pull/77/files#diff-3a2d9832e51b506b2866103eae7a445b4f67e23b3c8d215e6012d33466060fb9R149-R162

src/tests/test_meta_blk_mgr.cpp
```diff
@@ -553,6 +565,106 @@ class VMetaBlkMgrTest : public ::testing::Test {
...
+                    / \   \
+                   D   E   F
+        */
+
+        // register with dependencies
+        m_mbm->register_handler(
+            "A",
+            [this](meta_blk* mblk, sisl::byte_view buf, size_t size) {
+                meta_sub_type subType(mblk->hdr.h.type);
+                subtypes_order_after_sort_for_cb.push_back(subType);
```
xiaoxichen, line 590: s/subtypes_order_after_sort_for_cb/actual_cb_order

src/tests/test_meta_blk_mgr.cpp
```diff
@@ -614,6 +726,45 @@ TEST_F(VMetaBlkMgrTest, single_read_test) {
...
+    this->register_client_inlcuding_dependencies();
+
+    // simulate reboot case that MetaBlkMgr will scan the disk for all the metablks that were written;
+    this->scan_blks();
+
+    this->recover_with_on_complete();
+
+    std::unordered_map< meta_sub_type, int > subtypes_order_after_sort_for_cb_map;
+
+    // verify the order of callback
```
xiaoxichen, line 741: Suggesting moving https://github.com/eBay/HomeStore/pull/200/files#diff-820453c6d78ac456e2f86d1c2960472e447e98c0311c9392253b663e68f43989R633-R647 to test cases.

Better to enhance the checks as we want to ensure all A are before all B/C, i.e last A < first B. The implementation right now doesnt works if we add UT to have more than one records per subsys.

xiaoxichen (approved):
lgtm in implementation

A few improvement area in UT but the existing UT covered the logic

src/lib/meta/meta_blk_service.cpp
```diff
@@ -1074,85 +1077,125 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
...
 
-    if (do_comp_cb) {
-        // for each registered subsystem, do recovery complete callback;
-        for (auto& sub : m_sub_info) {
-            if (sub.second.comp_cb) {
-                sub.second.comp_cb(true);
-                HS_LOG(DEBUG, metablk, "[type={}] completion callback sent.", sub.first);
-            }
+std::vector< meta_sub_type >
+MetaBlkService::topological_sort(std::unordered_map< meta_sub_type, std::vector< meta_sub_type > > topological_graph) {
```
zichanglai, line 1161: it will be better to use reference&. LG to me for the rest part.

zichanglai (approved):
it will be better to use reference & for MetaBlkService::topological_sort parameter

src/include/homestore/meta_service.hpp
```diff
@@ -87,6 +88,7 @@ class MetaBlkService {
     MetablkMetrics m_metrics;
     bool m_inited{false};
     std::unique_ptr< meta_vdev_context > m_meta_vdev_context;
+    std::unordered_map< meta_sub_type, std::vector< meta_sub_type > > m_dependency_topological_graph;
```
yamingk, line 91: Can we move this to m_sub_info which mantains all the registration info for all clients?

xiaoxichen: step 2 is not that easy, B may have a dependency C (assuming it already registered ), and D may depends on B((assuming it also registered).  Now you need to insert B after C, but before A/D.

Actually you are doing topo-sort, in another form, standard topo-sort algorithm is much easier to follow.

JacksonYao287: @yamingk we do not know what kind of dependency(single or multi) will we meet in the future,  i think topological sort is thorough solution

src/lib/meta/meta_blk_service.cpp
JacksonYao287: ack

src/tests/test_meta_blk_mgr.cpp
JacksonYao287: ack

src/tests/test_meta_blk_mgr.cpp
JacksonYao287: good suggestion, will enhance the checks

src/include/homestore/meta_service.hpp
```diff
@@ -121,7 +123,7 @@ class MetaBlkService {
      * @param cb : subsystem cb
      */
     void register_handler(meta_sub_type type, const meta_blk_found_cb_t& cb, const meta_blk_recover_comp_cb_t& comp_cb,
-                          bool do_crc = true);
+                          bool do_crc = true, std::vector< meta_sub_type > dependencies = {});
```
szmyd, line 126: Maybe
```c++
bool do_crc = true, std::optional< meta_subtype_vec_t > = std::nullopt);
```
so we don't have to instantiate a `std::vector` in the case it's not needed? The map default doesn't need to be re-assigned to another empty vector right?

src/lib/meta/meta_blk_service.cpp
```diff
@@ -1074,85 +1077,125 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
...
+
+    // Calculate in-degree of each vertex
+    for (const auto& [vertex, edges] : topological_graph) {
+        in_degree[vertex] = 0;
+    }
+    for (const auto& [vertex, edges] : topological_graph) {
+        for (const auto& edge : edges) {
+            in_degree[edge]++;
+        }
+    }
```
szmyd, line 1174: Why does this need to be looped over twice? It seems reasonable to combine them into a single iteration no?

JacksonYao287: no, if we have some vertexes who`s in_degree is 0 and we combine the two loops, then, there will be no chance to put these vertexes into `in_degree`.

JacksonYao287: I think it is more clear to use a loop to add every vertex, but we can also combine the them, let me make this change

src/include/homestore/meta_service.hpp
JacksonYao287: got your point here.

src/include/homestore/meta_service.hpp
yamingk: I think the logic will be much more easier to follow if it is one-to-one dependency. But it could also be subjective. 
Bottom line is I am okay with any solution (algorithm) to solve the current problem (which is a one-to-one dependency), but we should not be solving problems that there is no real use case.

yamingk (changes requested):
I think the logic will be much more easier to follow if it is one-to-one dependency. But it could also be subjective.
Bottom line is I am okay with any solution (algorithm) to solve the current problem (which is a one-to-one dependency), but we should not be solving problems that there is no real use case.

xiaoxichen:
> I think the logic will be much more easier to follow if it is one-to-one dependency. But it could also be subjective. Bottom line is I am okay with any solution (algorithm) to solve the current problem (which is a one-to-one dependency), but we should not be solving problems that there is no real use case.

As the discussion in https://github.com/eBay/HomeStore/pull/200#discussion_r1370545309,  we all agree that even with one-to-one dependency we need to do some sort of topology sort.   The topology sort takes 38 lines of code,  with complexity of O(V+E).  I cannot find a solution,  if limit to one-to-one dependency , can significant reduce the #LOC or complexity. 

Not to say topology sort is well-known to every developer that is super easy to review, without worrying the correctness.

So the point here comes to "we should not support one to N dependency even we could",   it is simply just add one line of code in `register_handler` that checks the size of deps.   I dont think such check bring in business values,  but if that line is the turn-key to let us move forward,  lets add that line.

JacksonYao287:
@yamingk 

no matter multi or single dependency we face now , essentially we need to do topological sort to get a correct `partial order relation`.  [the comments](https://github.com/eBay/HomeStore/pull/200#discussion_r1370545309) you mentioned here is also essentially a kind of topological sort(suitable for only single dependency ) ,  right? 

based on the actual strong requirement of topological sort, the point here is not whether we need a topological sort, but what kind of topological sort we need. 

coming to this point , Obviously， a common thorough solution with extremely low complexity of algorithm and implementation is better than a specific one for only one case yea?

yamingk:
@JacksonYao287 Can we do one thing: 
1. only do order sort on those clients who participant the dependency (either being depended or depending on others), given that most of the clients doesn't have dependency with others. 
2. during recovery, write a new function to do recover on order based clients and have current MetaBlkService::recover to call this new function, and leave non-order based clients to recover with existing flow?

JacksonYao287:
@yamingk 

>only do order sort on those clients who participant the dependency (either being depended or depending on others), given that most of the clients doesn't have dependency with others.

we use a `std::optional` as a dependency parameter now, so actually it has given the client a change to indicate whether it has any dependecy. meanwhile,  it gives us a uniform function declaration.

topological sort is fully in memory , so it is fast. and it is only be called once when recovery, so it is one-time shot. we can basically ignore the performance cost it brings. actually , if no dependency passed in, topological sort will do nothing, since no edge exists, right?

if we want do concurrent recovery with no dependency, actually the current logic can also support this . what topological sort does is generating a recovery order. if any dependency exists,  it will generate a partial relation order. if no dependency exist, in fact a random order returned, just like the current random recovery order.

the benefit we keep this code style is it gives us a clean and uniform function, which will be better for the code reader and UT.

JacksonYao287:
@yamingk 

i follow you suggestion to  devided the subsystem into two part: dependent and independent. if we want to speed up recovery time in the future, we can concurrent recover the independent part.

src/lib/meta/meta_blk_service.cpp
```diff
@@ -1074,85 +1081,165 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
...
-    if (do_comp_cb) {
-        // for each registered subsystem, do recovery complete callback;
-        for (auto& sub : m_sub_info) {
-            if (sub.second.comp_cb) {
-                sub.second.comp_cb(true);
-                HS_LOG(DEBUG, metablk, "[type={}] completion callback sent.", sub.first);
-            }
+    // Add vertices with in-degree 0 to the queue
+    for (const auto& [vertex, degree] : in_degree) {
+        if (degree == 0) { q.push(vertex); }
```
xiaoxichen, line 1222: the whole https://github.com/eBay/HomeStore/pull/200/files#diff-428b89aa263f097554416e42d736778169b33e2b6d72febef88b1c981b84bd0aR1086-R1102
can be moved here 
```
if (degree == 0) { 
    if (len(topological_graph[vertex]) ==0) 
          indep_subs.push_back(vertx)
     else 
           q.push(vertex); 
}
```

need to change the function signature as
```
bool MetaBlkService::topological_sort(std::unordered_map< meta_sub_type, std::vector< meta_sub_type > >& topological_graph,  std::vector< meta_sub_type> &indep_subs, std::vector< meta_sub_type > &ordered_subs)
```

src/lib/meta/meta_blk_service.cpp
```diff
@@ -375,7 +377,12 @@ void MetaBlkService::register_handler(meta_sub_type type, const meta_blk_found_c
     m_sub_info[type].cb = cb;
     m_sub_info[type].comp_cb = comp_cb;
     m_sub_info[type].do_crc = do_crc ? 1 : 0;
-    HS_LOG(INFO, metablk, "[type={}] registered with do_crc: {}", type, do_crc);
+    m_dependency_topological_graph[type];
```
yamingk, line 380: what is this trying to achieve?

yamingk, line 380: You are trying to put every client into the graph and during recover do a few for loops to determin independent clients and then erase from the graph, which is unncessary. 

We already have a client_map, m_sub_info, which contains every client's private information and put a new field: bool no_dep {true}; 

during register_handler, if this client appears in anyone's dep vector, set this client's no_dep to false, your graph's existing operation stays the same. 

During recovery, loop this m_sub_info, if no_dep is true, call recover_meta_blk(m_sub_info(client).meta_blk). 
Then loop your graph to call recover_meta_blk() in order.

src/lib/meta/meta_blk_service.cpp
```diff
@@ -1074,85 +1081,165 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
...
-                        HS_REL_ASSERT_EQ(uint64_cast(mblk->hdr.h.src_context_sz),
-                                         uint64_cast(decompressed_size)); /* since decompressed_size is >=0 it
-                                                                             is safe to cast to uint64_t */
-                        HS_LOG(DEBUG, metablk,
-                               "[type={}] Successfully decompressed, compressed_sz: {}, src_context_sz: {}, "
-                               "decompressed_size: {}",
-                               mblk->hdr.h.type, uint64_cast(mblk->hdr.h.compressed_sz),
-                               uint64_cast(mblk->hdr.h.src_context_sz), decompressed_size);
-                    }
+    meta_subtype_vec_t independent_subsystems;
```
yamingk, line 1085: line: 1085 ~1105 can all be removed if we rely on m_sub_info for independent clients. The graph doesn't need to interact with independent clients.

src/lib/meta/meta_blk_service.cpp
```diff
@@ -1074,85 +1081,165 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
...
 #if 0
             // should never arrive here since we do assert on type before write to disk;
             HS_LOG_ASSERT( false, "[type={}] not registered for mblk found on disk. Skip this meta blk. ",
                       mblk->hdr.h.type);
 #endif
+    }
+}
+
+std::vector< meta_sub_type >
+MetaBlkService::topological_sort(std::unordered_map< meta_sub_type, std::vector< meta_sub_type > >& topological_graph) {
```
yamingk, line 1205: Move this to `class hs_utils`, replace meta_sub_type to std::string, so that it can be reused as homestore utility.

src/lib/meta/meta_blk_service.cpp
yamingk, line 380: s/m_dependency_topological_graph/m_dep_topo_graph/g

src/lib/meta/meta_blk_service.cpp
```diff
@@ -1074,85 +1081,165 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
...
+        if (do_comp_cb && m_sub_info[subtype].comp_cb) {
+            m_sub_info[subtype].comp_cb(true);
+            HS_LOG(DEBUG, metablk, "[type={}] completion callback sent.", subtype);
+        }
+    }
+
+    // for subsystems in independent_subsystems, we can use concurrent recovery if necessary.
+    // for now, we recover independent_subsystems in current thread.
+    // we can start a thread to recover independent_subsystems concurrently to improve performace in the future.
+    for (const auto& subtype : independent_subsystems) {
```
yamingk, line 1123: no need to maintain independent_subsystems, use m_sub_info and deduce independent clients during register_handler.

src/lib/meta/meta_blk_service.cpp
```diff
@@ -1074,85 +1081,165 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
...
+        result.push_back(vertex);
+
+        for (const auto& edge : topological_graph[vertex]) {
+            in_degree[edge]--;
+            if (in_degree[edge] == 0) { q.push(edge); }
         }
     }
+
+    // Check for cycle
+    if (result.size() != topological_graph.size()) {
```
yamingk, line 1238: My concern is now we are failing the recovery (which is a no-go), instead of catching the dependency violation during register_handler to fail the client trying to mess up the dependency.

src/include/homestore/meta_service.hpp
```diff
@@ -121,7 +125,7 @@ class MetaBlkService {
      * @param cb : subsystem cb
      */
     void register_handler(meta_sub_type type, const meta_blk_found_cb_t& cb, const meta_blk_recover_comp_cb_t& comp_cb,
```
yamingk, line 127: It needs to return bool if the dep requested can cause circle or any invalidation to the graph.

src/lib/meta/meta_blk_service.cpp
xiaoxichen: you cannot define yourself as no_dep as you only know you dont depends on anyone, but you dont know if there is anyone depends on you.

src/lib/meta/meta_blk_service.cpp
xiaoxichen: same as above

src/lib/meta/meta_blk_service.cpp
JacksonYao287: >during register_handler, if this client appears in anyone's dep vector, set this client's no_dep to false, your graph's existing operation stays the same.

a subsystem can be excluded from the graph only if it depends on none, and none depends on it too. if this client does not appear in anyone's dep vector, it means none depends on it. but you have no knowledge whether it depends on anyone.

src/lib/meta/meta_blk_service.cpp
JacksonYao287: we can not do this， since during register_handler， a subsystem only know whether it depends on anyone, but it does not know anyone depends on it

src/lib/meta/meta_blk_service.cpp
JacksonYao287: not fully get your point . if we fails the recovery, what we can do is only crushing immediately. or can you describe this in more detail?

src/include/homestore/meta_service.hpp
```diff
@@ -332,6 +337,8 @@ class MetaBlkService {
     uint64_t max_compress_memory_size() const;
     uint64_t init_compress_memory_size() const;
 
+    std::vector< meta_sub_type > topological_sort(std::unordered_map< meta_sub_type, std::vector< meta_sub_type > >&);
```
zichanglai, line 340: if we have define meta_subtype_vec_t, please use it instead of std::vector< meta_sub_type > in the whole PR.

JacksonYao287: this func has been moved to hs_utils, we will use the original type there

src/lib/meta/meta_blk_service.cpp
JacksonYao287: ack

xiaoxichen (dismissed):
lgtm

src/lib/meta/meta_blk_service.cpp
yamingk: It is deduced after all register completes. For every resigter_hander, for this clicent, and every client in dep vec, set m_sub_info[client].no_dep = false. It can't be determined by one register call.

src/lib/meta/meta_blk_service.cpp
```diff
@@ -1074,84 +1081,108 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
...
-                        HS_REL_ASSERT_EQ(uint64_cast(mblk->hdr.h.src_context_sz),
-                                         uint64_cast(decompressed_size)); /* since decompressed_size is >=0 it
-                                                                             is safe to cast to uint64_t */
-                        HS_LOG(DEBUG, metablk,
-                               "[type={}] Successfully decompressed, compressed_sz: {}, src_context_sz: {}, "
-                               "decompressed_size: {}",
-                               mblk->hdr.h.type, uint64_cast(mblk->hdr.h.compressed_sz),
-                               uint64_cast(mblk->hdr.h.src_context_sz), decompressed_size);
-                    }
+    meta_subtype_vec_t independent_subtypes;
```
yamingk, line 1085: same as other comment, no need to have this and do for loop to deduce this unnecessary.

yamingk (changes requested):

src/lib/meta/meta_blk_service.cpp
```diff
@@ -375,7 +377,12 @@ void MetaBlkService::register_handler(meta_sub_type type, const meta_blk_found_c
     m_sub_info[type].cb = cb;
     m_sub_info[type].comp_cb = comp_cb;
     m_sub_info[type].do_crc = do_crc ? 1 : 0;
-    HS_LOG(INFO, metablk, "[type={}] registered with do_crc: {}", type, do_crc);
+    m_dep_topo_graph[type];
```
yamingk, line 380: this is not needed. 
if (deps) {
    m_sub_info[type].no_dep = false;
    for (auto const& x : deps) {
        m_sub_info[x].no_deps = false;
    }
}

src/lib/meta/meta_blk_service.cpp
```diff
@@ -1074,84 +1081,108 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
...
+        }
+
+        if (do_comp_cb && m_sub_info[subtype].comp_cb) {
+            m_sub_info[subtype].comp_cb(true);
+            HS_LOG(DEBUG, metablk, "[type={}] completion callback sent.", subtype);
+        }
+    }
+
+    // TODO: for subsystems in independent_subtypes, we can use concurrent recovery if necessary.
+    for (const auto& subtype : independent_subtypes) {
```
yamingk, line 1108:
```
for (auto const& x : m_sub_info) {
    if (m_sub_info[x].no_dep) {
         /// do recover stuff
    }
}
```

src/lib/meta/meta_blk_service.cpp
```diff
@@ -1074,84 +1081,108 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
...
-                                         uint64_cast(decompressed_size)); /* since decompressed_size is >=0 it
-                                                                             is safe to cast to uint64_t */
-                        HS_LOG(DEBUG, metablk,
-                               "[type={}] Successfully decompressed, compressed_sz: {}, src_context_sz: {}, "
-                               "decompressed_size: {}",
-                               mblk->hdr.h.type, uint64_cast(mblk->hdr.h.compressed_sz),
-                               uint64_cast(mblk->hdr.h.src_context_sz), decompressed_size);
-                    }
+    meta_subtype_vec_t independent_subtypes;
+    meta_subtype_vec_t ordered_subtypes;
```
yamingk, line 1086: also not needed.

src/lib/meta/meta_blk_service.cpp
```diff
@@ -1074,84 +1081,108 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
...
+
+    if (hs_utils::topological_sort(m_dep_topo_graph, ordered_subtypes, &independent_subtypes)) {
```
yamingk, line 1088: only topo sort is needed, no need to deduce the other two

yamingk, line 1088: Can we move this sort out of MetaBlkService::recover and put it into either MetaBlkService::start (before ::recover) or into a new function (e.g. validate_register_dep(...)) to indicate this is a validation of register dependency, since this is not part of recovery.

src/lib/meta/meta_blk_service.cpp
xiaoxichen: @yamingk that looks good through seems we dont need to maintain the `no_dep`
```
std::unordered_set<std::string> m_independent_subtypes;
```
```
if (deps.has_value()) {
   for (auto const& x : deps.value()) {
        m_independent_subtypes.erase(x);
        m_dep_topo_graph[x].push_back(type);
    }
    m_dep_topo_graph[type];
 } else {
     m_independent_subtypes.insert(type);
```

JacksonYao287: no. when client register , it only know whether it depends on others, it does not know whether others depend on it until all the register action finishes. what`s more, the way we determine whether a subtype is independent is to put all the subtype in the graph and check the in_degree and out_degree. only those subtypes with both 0 in_degree and 0 out_degree are actually independent

src/lib/meta/meta_blk_service.cpp
JacksonYao287: as i mentioned, no matter it is a fresh boot or not, the order is `do_start` -> `metaservice#start`->`recover()`.  we can delay this change, until we move recover out of recover. this change will bring a lot of changes in current metaservice UT, and so we better to do all the related works in a dedicated PR

JacksonYao287: for now, we use `hs_utils::topological_sort` to both check circle and get the ordered_subtypes. if we call this in another function out of recovery, we need to pass the ordered_subtypes to `MetaBlkService::recover`, which need to change the signature of `MetaBlkService::recover` , and thus we need to do a lot of changes in the whole metaservice UT. can we do it in a seperate PR?

if we do not want to change the signature of `MetaBlkService::recover`, we can add ordered_subtypes as a member in MetaBlkService so that `MetaBlkService::recover` can read it, but it is generate from the `m_dep_topo_graph` and only be used once, i do not think it is a good fashion

src/lib/meta/meta_sb.hpp
```diff
@@ -91,6 +91,7 @@ struct MetaSubRegInfo {
     std::set< uint64_t > meta_bids; // meta blk id
     meta_blk_found_cb_t cb{nullptr};
     meta_blk_recover_comp_cb_t comp_cb{nullptr};
+    bool has_deps;
```
yamingk, line 94: Needs to be defaulted to false

src/tests/test_meta_blk_mgr.cpp
```diff
@@ -43,6 +42,8 @@
 #include "test_common/bits_generator.hpp"
 #include "test_common/homestore_test_common.hpp"
 
+#define private public
```
yamingk, line 45: is this expected?

src/tests/test_meta_blk_mgr.cpp
```diff
@@ -731,6 +895,7 @@ SISL_OPTION_GROUP(
     (bitmap, "", "bitmap", "bitmap test", ::cxxopts::value< bool >()->default_value("false"), "true or false"));
 
 int main(int argc, char* argv[]) {
+    ::testing::GTEST_FLAG(filter) = "*register_dependency_test*";
```
yamingk, line 898: wouldn't it overwrite the next line?

src/lib/meta/meta_blk_service.cpp
```diff
@@ -1074,84 +1082,109 @@ sisl::byte_array MetaBlkService::read_sub_sb_internal(const meta_blk* mblk) cons
...
+    meta_subtype_vec_t ordered_subtypes;
+
+    if (hs_utils::topological_sort(m_dep_topo_graph, ordered_subtypes)) {
+        throw homestore::homestore_exception(
+            "MetaBlkService has circular dependency, please check the dependency graph", homestore_error::init_failed);
+    }
+
+    // all the subsystems are divided into two parts.
+    // for subsystems in ordered_subtypes, we need to recover in order.
+    for (const auto& subtype : ordered_subtypes) {
```
yamingk, line 1094: recovery_meta_block should take client type and move line: 1095~ 1101, 1110 ~1118 to a new recover_meta_block(client_type) which takes client type as input. This function now provides recovery (return all metablks) for this client type, not just for a given metablk cookie. 

```
for (const auto& subtype : ordered_subtypes) {
    recover_meta_blk(subtype);
}

 for (auto const& x : m_sub_info) {
        auto& reg_info = x.second;
        if (!reg_info.has_deps) {
            recover_meta_block(x.first);
        }
```

```
void recover_meta_block(meta_sub_type client) {
        for (const auto& m : meta_sub_info[client].meta_bids) {
                auto mblk = m_meta_blks[m];
                recover_meta_block(mblk);
            }

            if (do_comp_cb && reg_info.comp_cb) {
                reg_info.comp_cb(true);
                HS_LOG(DEBUG, metablk, "[type={}] completion callback sent.", x.first);
            }
}
```

src/lib/meta/meta_blk_service.cpp
xiaoxichen: the sort in `recovery` or in `validate_register_dep` doesnt have any behavior change,  more for the code style. It looks to me also seems wired that `validate_register_dep` returns a order and passed to `MetaBlkService::recover`

Maybe do the topo-sort twice is a proper solution for the sake of code-style...

Honestly I dont think we need to spend more cycle on this style/preference and blocking the PR (other comments needed to be addressed),   as it is significant communication overhead to let other believe this is better compare to having another PR to fix the style to your prefer one.

src/lib/meta/meta_sb.hpp
JacksonYao287: the default value of bool is false, no need to set it explicitly

src/lib/meta/meta_blk_service.cpp
JacksonYao287: ack

src/tests/test_meta_blk_mgr.cpp
JacksonYao287: fixed

src/tests/test_meta_blk_mgr.cpp
JacksonYao287: removed

src/lib/meta/meta_sb.hpp
yamingk: Isn't it depending on the compiler? Why should we stop a good habit of always initializing the var before accessing it, not to mention any complaining code analysis's tool that would report issues on it.

src/lib/meta/meta_blk_service.cpp
yamingk: One can argue either way, right, why do we need another set compared to using existing one?

src/lib/meta/meta_blk_service.cpp
yamingk: I don't get why we want to put more stuff in a function that is apparent not part of its job, and also would cause some confusion to the reader. The validation simply is NOT part of recovery, it is part of registration, if it fails, it shouldn't start recover process. Another benefit is if some use case want to call ::recover multiple times (either to measure the speed of recovery or test the dep/non-dep order correctness), it shouldn't do the sort multiple times, but just do recover with its clients.

xiaoxichen: Logically it is ,  the point is the cycle-detection is actually a topo-sort,  so we either 

1. do the topo-sort twice , one for validation at the end of registration(validate_register_dep) , and one for recover.
2. let recovery() accept `ordered_subtypes` and `independent_subtypes`, which returned from `validate_register_dep` and passed in here.
3. Put the validation in recovery()

I think 2) will makes the flow more complicate and other caller of `recovery` will be  more confusing.   I am fine with 1 or 3. Though honestly dont see too much benefit.

yamingk (approved):

src/lib/meta/meta_blk_service.cpp
yamingk: There is no need for independent_subtypes, where we can get from m_sub_info. 
This is not a major comment. Feel free to merge as it as. The rest looks good to me. 
But please expect some refactoring as to this part in next PR.

PR #209, opened 2023-10-18
get defrag nblks api

yamingk:
1. provide get defrag nblks api
2. pct check for each service
3.  allow test framework to customize num_chunks while starting homestore service, apply 64K num_chunks for appendblk allocator test cases.

Tested with homeobject passed.


src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -179,6 +175,12 @@ std::string AppendBlkAllocator::to_string() const {
     return fmt::format("{}, last_append_offset: {}", get_name(), m_last_append_offset);
 }
 
+blk_num_t AppendBlkAllocator::available_blks() const { return get_total_blks() - get_used_blks(); }
+
+blk_num_t AppendBlkAllocator::get_used_blks() const { return m_last_append_offset; }
+
 blk_num_t AppendBlkAllocator::get_freeable_nblks() const { return m_freeable_nblks; }
 
+blk_num_t AppendBlkAllocator::get_defrag_nblks() const { return get_freeable_nblks() - available_blks(); }
```
zichanglai, line 184: actually I am little confuse about the definition of defrag_nblks() and why it use free_blks - available_blks ?
from HO GC perspective,  the chunk with larger free blks is more worth to do the compaction, right? why need to minus available blks

src/lib/homestore.cpp
```diff
@@ -145,6 +145,16 @@ bool HomeStore::start(const hs_input_params& input, hs_before_services_starting_
 }
 
 void HomeStore::format_and_start(std::map< uint32_t, hs_format_params >&& format_opts) {
+    auto total_pct_sum = 0.0f;
+    for (const auto& [svc_type, fparams] : format_opts) {
+        total_pct_sum += fparams.size_pct;
+    }
+
+    if (total_pct_sum > 100.0f) {
```
JacksonYao287, line 153: https://github.com/eBay/HomeStore/blob/202f07fa57361f657d74a60c13c907c6391e256f/src/lib/device/device_manager.cpp#L222

since we use `round_up`, so there may be a case that every v_dev will get little more size of expected, so the final sum of all dev will cap the total size, right? should we use `round_down`

src/lib/blkalloc/append_blk_allocator.cpp
xiaoxichen: i think this is should be `available_blks()- get_freeable_nblks()` = `get_total_blks() - get_used_blks()- get_used_blks()` , indicating the live data on the chunk

JacksonYao287: if i am right , defrag_nblks() means the actual deleted number of blocks, which can be gc.
`m_freeable_nblks` means the theorically freeable number of blks, which will be updated in every `async_free`.

`available_blks()` means the actuall freeable number of blks, which means the left space never be allocated .

so the `defrag_nblks()` is ` theorically freeable` minus `actual freeable`, that is the blks need to be GC.

pls correct me if i am wrong, @yamingk

zichanglai: ok, m_freeable_nblks will be initialize to all blk num of this chunk during format and will be increase during every async_free(). it is much clear to me. @JacksonYao287 @yamingk .

JacksonYao287 (approved):

xiaoxichen (approved):

xiaoxichen:
@yamingk rebased to solved the conanfile conflicts, feel free to merge.

JacksonYao287 (approved):
LGTM

xiaoxichen:
@JacksonYao287 next time please if you can leave the merge to author,  as we do squash merge,  the author should take the responsibility to write a meaningful commit message (that will stay in git history) .

This is a bad example of commit w/o any context in commit message.
image

JacksonYao287:
@xiaoxichen sorry , for this.
this is a small PR and i need to use it for my test, so i take a quick merge.
will leave the merge to author and follow the guideline next time!

PR #211, opened 2023-10-21
Changes to support PrefixNode and consolidation of tree

hkadayam:
Following are the major changes in btree library:

1. A new BtreeKey and BtreeValue type called IntervalKey and IntervalValue which allows a particular key to be part of an larger interval (say 1-3 will result in key1, key2, key3 and each key is an BtreeIntervalKey).

2. Implemented a new tree node called PrefixNode which can store interval key and value effectively stores them with prefix and suffix to store the intervals compactly.

3. Removed the Extent code which was not used and replaced it with the IntervalKey/Value as mentioned above.

4. Added filtering for read/write/remove operations, so that consumer can change the way which key/value to be custom modify. This change effectively provides compare_and_set or compare_and_remove operations atomically. This change also removed the previous on_read_cb etc, which was passed on every callback

5. Removed the requirement for BtreeKey and Value to implement some static methods like get_key_size() etc, by creating a dummyKey, dummyValue. Now the requirement of the implementor of btree key and value only needs to ensure it is default constructible (which is already the case today)

6. Created new internal btree node APIs such as multi_put, multi_get instead of each btree operations repeat similar code to read or remove operations in leaf node. As a result, a method called match_range() is introduced in the btree node, which consolidates and searches for all range operation requirements. Moved most of the common operations into variant_node class and now variant_node is derived from BtreeNode. PrefixNode overrides variant_node and implements its own version of multi_put etc. range operations.

7. Removed obsolete btree return status.

8. Consolidated the mem_btree test, mem_btree concurrent test and index_btree test into one btree_test_helper library. Modified the way range_scheduler picks the existing and working keys from boost icl to sis::Bitset as it simplifies them and also provides some essential feature which is always pick some keys to schedule instead of failing.

### DIFF
--- a/src/include/homestore/btree/btree_kv.hpp
+++ b/src/include/homestore/btree/btree_kv.hpp
@@ -31,11 +31,10 @@ ENUM(MultiMatchOption, uint16_t,
 )
 
 ENUM(btree_put_type, uint16_t,
-     INSERT_ONLY_IF_NOT_EXISTS,     // Insert
-     REPLACE_ONLY_IF_EXISTS,        // Update
-     REPLACE_IF_EXISTS_ELSE_INSERT, // Upsert
-     APPEND_ONLY_IF_EXISTS,         // Update
-     APPEND_IF_EXISTS_ELSE_INSERT)
+     INSERT, // Insert only if it doesn't exist
+     UPDATE, // Update only if it exists
+     UPSERT  // Update if exists, insert otherwise
+)
 
 // The base class, btree library expects its key to be derived from
 class BtreeKey {
@@ -45,23 +44,33 @@ public:
     // Deleting copy constructor forces the derived class to define its own copy constructor
     // BtreeKey(const BtreeKey& other) = delete;
     // BtreeKey(const sisl::blob& b) = delete;
-    BtreeKey(const BtreeKey& other) = default;
+    BtreeKey(BtreeKey const& other) = default;
     virtual ~BtreeKey() = default;
 
-    virtual BtreeKey& operator=(const BtreeKey& other) {
-        clone(other);
-        return *this;
-    };
-
-    virtual void clone(const BtreeKey& other) = 0;
-    virtual int compare(const BtreeKey& other) const = 0;
+    virtual int compare(BtreeKey const& other) const = 0;
 
     virtual sisl::blob serialize() const = 0;
     virtual uint32_t serialized_size() const = 0;
-    virtual void deserialize(const sisl::blob& b, bool copy) = 0;
+    virtual void deserialize(sisl::blob const& b, bool copy) = 0;
 
     virtual std::string to_string() const = 0;
-    virtual bool is_extent_key() const { return false; }
+    virtual bool is_interval_key() const { return false; }
+};
+
+// An extension of BtreeKey where each key is part of an interval range. Keys are not neccessarily only needs to be
+// integers, but it needs to be able to get next or prev key from a given key in the key range
+class BtreeIntervalKey : public BtreeKey {
+public:
+    virtual void shift(int n) = 0;
+    virtual int distance(BtreeKey const& from) const = 0;
+    bool is_interval_key() const override { return true; }
+
+    virtual sisl::blob serialize_prefix() const = 0;
+    virtual sisl::blob serialize_suffix() const = 0;
+
+    virtual uint32_t serialized_prefix_size() const = 0;
+    virtual uint32_t serialized_suffix_size() const = 0;
+    virtual void deserialize(sisl::blob const& prefix, sisl::blob const& suffix, bool copy) = 0;
 };
 
 template < typename K >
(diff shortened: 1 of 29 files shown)


yamingk:
need CompactBitset to compile.

shosseinimotlagh (changes requested):

src/include/homestore/btree/detail/varlen_node.hpp
```diff
@@ -63,10 +71,6 @@ class VariableNode : public BtreeNode {
 
     virtual ~VariableNode() = default;
 
-    uint32_t occupied_size(const BtreeConfig& cfg) const override {
-        return (cfg.node_data_size() - sizeof(var_node_header) - available_size(cfg));
```
shosseinimotlagh, line 67: It is overriden here because in merge_nodes(), for varlen node type  `available_size -= old_nodes[i]->occupied_size();` results in miscalculation. It can happen frequently with reduced node size (for instance 512). 
In merge if the total entries of a node is needed to move to left, only the size of actual data matters not `var_node_header`.

src/include/homestore/btree/detail/btree_remove_impl.ipp
```diff
@@ -80,21 +65,22 @@ retry:
...
-        BT_NODE_REL_ASSERT_NE(count, 0, my_node, "get_all returns 0 entries for interior node is not valid pattern");
+        auto const matched = my_node->match_range< K >(req.m_range, start_idx, end_idx);
+        if (!matched) {
+            ret = btree_status_t::not_found;
+            goto out_return;
+        }
         end_idx = start_idx = (end_idx - start_idx) / 2; // Pick the middle, TODO: Ideally we need to pick random
     }
-    if (go_to_out) { goto out_return; }
+    // if (go_to_out) { goto out_return; }
```
yamingk, line 83: same here

src/include/homestore/btree/btree_kv.hpp
```diff
@@ -45,37 +44,43 @@ class BtreeKey {
...
+    virtual int compare(BtreeKey const& other) const = 0;
 
     virtual sisl::blob serialize() const = 0;
     virtual uint32_t serialized_size() const = 0;
-    virtual void deserialize(const sisl::blob& b, bool copy) = 0;
+    virtual void deserialize(sisl::blob const& b, bool copy) = 0;
 
     virtual std::string to_string() const = 0;
-    virtual bool is_extent_key() const { return false; }
+    virtual bool is_interval_key() const { return false; }
```
yamingk, line 57: I know it is from the legacy code, a little odd to me that base class has knowledge about subclass. 
it looks like same goal can be achieved by try to dynamic cast from a base class pointer. Is there any problem if we remove this api?

src/include/homestore/btree/detail/btree_mutate_impl.ipp
```diff
@@ -102,17 +102,7 @@ retry:
...
-            // TODO remove the split_node retry logic and use key max size.
-            if (!my_node->can_accomodate(m_bt_cfg, K::get_estimate_max_size(), BtreeLinkInfo::get_fixed_size())) {
-                // Mark the parent_node itself to be split upon next retry.
-                bt_thread_vars()->force_split_node = my_node;
-                unlock_node(child_node, child_cur_lock);
-                ret = btree_status_t::retry;
-                goto out;
-            }
-
+        if (is_split_needed(child_node, req) || is_repair_needed(child_node, child_info)) {
```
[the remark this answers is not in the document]
hkadayam: Ok, did that.

src/include/homestore/btree/detail/varlen_node.hpp
hkadayam: I am probably missing something, but doing that overridden would defeat the generic purpose right? I am not clear why available_size() doesn't take discount the `var_node_header` area?

src/include/homestore/btree/detail/btree_remove_impl.ipp
```diff
@@ -68,7 +53,7 @@ btree_status_t Btree< K, V >::do_remove(const BtreeNodePtr& my_node, locktype_t
         return modified ? btree_status_t::success : btree_status_t::not_found;
     }
 
-    bool go_to_out = false;
+    // bool go_to_out = false;
```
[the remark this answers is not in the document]
hkadayam: Done

src/include/homestore/btree/btree_kv.hpp
hkadayam: No it is not that base class is aware of derived class, but by default they are not interval key. Only BtreeIntervalKey can claim that. If this is not there, then we need to define additional BtreeNonIntervalKey which does only this method. We can't do it with dynamic_cast either, because btree main library will take key and call `is_interval_key()`,

src/include/homestore/btree/detail/btree_remove_impl.ipp
hkadayam: Done

yamingk (approved):

sanebay (approved):

PR #215, opened 2023-10-26
Implementation of bitmap blk allocator

hkadayam:
This PR is a part of several stages to make sure BlkAllocation is persisted under single umbrella. In order to ease the review, splitting this into multiple PRs.

**BitmapBlkAllocator**
At present we have 3 allocators, Fixed, Varsize, Append. 
Fixed uses list for in-memory side and bitmap for persistence
Varsize uses bitmap and slab for in-memory side and separate bitmap for persistence
Append uses only a location and does in-memory and persistence within single structure and no bitmap.

This PR provides class hierarchy as base 
```
                                  -> FixedBlkAllocator
                                  |
            -> BitmapBlkAllocator -
            |                     |
            |                     -> VarsizeBlkAllocator
BlkAllocator 
            |                                       
            |                                       
            -> AppendBlkAllocator 
```
All on-disk implementations are moved to BitmapBlkAllocator. 

**Realtime Bitmap**
In this full implementation of this requirement, we would have upper layers to just call one of the 3 apis and everything else is handled inside, viz, _alloc_blk()_, _commit_blk()_, _free_blk()_. All other apis such alloc_on_disk, free_on_disk etc are handled internally. That is same with realtime_bitmap. Currently in homestore 3.x we have realtime_bm being driven by caller. We will achieve the same thing to keep track of realtime bitmap slightly differently and internally, which will come in next PR. 

**Other adjustments**
* Also renamed auto_recovery flag in blkallocator to is_persistent which is more apt name from blk allocator perspective.
* Removed explicit inited() call for each chunk/blkallocator which is unnecessary. We need one global check if homestore initialize is completed or not, when we are checking for asserts.


src/lib/blkalloc/blk_allocator.h
```diff
@@ -184,155 +151,42 @@ class BlkAllocator {
...
-    BlkAllocPortion& get_blk_portion(blk_num_t portion_num) {
-        HS_DBG_ASSERT_LT(portion_num, get_num_portions(), "Portion num is not in range");
-        return m_blk_portions[portion_num];
-    }
-
-    virtual void inited();
+    virtual std::string to_string() const = 0;
+    virtual void cp_flush(CP* cp) = 0;
 
     void incr_alloced_blk_count(blk_count_t nblks) { m_alloced_blk_count.fetch_add(nblks, std::memory_order_relaxed); }
```
yamingk, line 169: since appendblkallocator can decude this m_alloced_blk_count and doesn't call incr/decr in its implementation, should we move these three apis (inc/dec/get) into bitmap_blk_allocator.h? It will be inaccurate anyway for append if we keep it here.

src/lib/blkalloc/bitmap_blk_allocator.cpp
```diff
@@ -0,0 +1,201 @@
...
+        hs_utils::extract_byte_array(buf, meta_service().is_aligned_buf_needed(size), meta_service().align_size())}};
+
+    m_alloced_blk_count.store(m_disk_bm->get_set_count(), std::memory_order_relaxed);
+    do_init();
+
+    load();
+}
+
+void BitmapBlkAllocator::cp_flush(CP*) {
+    if (m_is_disk_bm_dirty.load()) {
```
yamingk, line 63: is it expected for is_persistent set to false, we still do cp flush for this allocator for the 1st cp, since m_is_disk_bm_dirty is defaulted to true? The later cp_flush won't do anything as no one is going to set it back to dirty.

src/lib/blkalloc/bitmap_blk_allocator.h
```diff
@@ -0,0 +1,121 @@
...
+        return m_blk_portions[portion_num];
+    }
+
+    blk_num_t blknum_to_portion_num(const blk_num_t blknum) const { return blknum / m_blks_per_portion; }
+    BlkAllocPortion& blknum_to_portion(blk_num_t blknum) { return m_blk_portions[blknum_to_portion_num(blknum)]; }
+    BlkAllocPortion const& blknum_to_portion_const(blk_num_t blknum) const {
+        return m_blk_portions[blknum_to_portion_num(blknum)];
+    }
+
+    sisl::Bitset const* get_disk_bitmap() const { return is_persistent() ? m_disk_bm.get() : nullptr; }
```
yamingk, line 94: So the rule is if the api doesn't have xxx_const, it is by default is const, for mutable, we need xxx_mutable.

src/lib/blkalloc/bitmap_blk_allocator.cpp
```diff
@@ -0,0 +1,201 @@
...
+    auto list = get_alloc_blk_list();
+    if (list) {
+        // cp has started, accumulating to the list
+        list->push_back(bid);
+    } else {
+        auto set_on_disk_bm = [this](auto& b) {
+            BlkAllocPortion& portion = blknum_to_portion(b.blk_num());
+            {
+                auto lock{portion.portion_auto_lock()};
+                if (!hs()->is_initializing()) {
```
yamingk, line 101: Can we also put a comment here by saying, during recovery, log replay might do alloc

yamingk:
LGTM

src/lib/blkalloc/bitmap_blk_allocator.cpp
```diff
@@ -0,0 +1,201 @@
...
+        meta_service().register_handler(
+            get_name(),
+            [this](meta_blk* mblk, sisl::byte_view buf, size_t size) {
+                on_meta_blk_found(voidptr_cast(mblk), std::move(buf), size);
+            },
+            nullptr);
+    }
+
+    if (is_fresh || !is_persistent()) {
+        m_disk_bm = std::make_unique< sisl::Bitset >(m_num_blks, m_chunk_id, m_align_size);
```
yamingk, line 35: We are allocating memory for disk_bm when is_persistent return false, is it expected because we always access this bitmap on 1st cp? 
If so can we avoid doing that if it is not a persistent bitmap to save some memory budget which will not be used?

src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -212,7 +212,10 @@ folly::Future< std::error_code > BlkDataService::async_free_blk(MultiBlkId const
     auto f = promise.getFuture();
 
     m_blk_read_tracker->wait_on(bids, [this, bids, p = std::move(promise)]() mutable {
-        m_vdev->free_blk(bids);
+        {
+            auto cpg = hs()->cp_mgr().cp_guard();
```
yamingk, line 216: We only take cp_guard for bitmapblk allocator, for appendblkallocator, it would just call m_vdev->free_blk(), right?

src/lib/blkalloc/blk_allocator.h
hkadayam: Make sense, however, we may need to look out for future blk allocators which need this, but I agree for now I can move that there.

src/lib/blkalloc/bitmap_blk_allocator.cpp
hkadayam: Good catch. I think its a mistake to call this !is_persistent(). I think we need to create only if
`is_fresh && is_persistent`

src/lib/blkalloc/bitmap_blk_allocator.cpp
hkadayam: Yes good catch again. I will add a check if is_persistent check here as well.

src/lib/blkalloc/bitmap_blk_allocator.cpp
hkadayam: Ok, I think there is a comment in free_blks, will add here as well.

src/lib/blkalloc/bitmap_blk_allocator.h
hkadayam: I don't think there is any set rule, but new code we can try to do this.

src/lib/blkdata_svc/blkdata_service.cpp
hkadayam: I didn't want to do a specific call for different blk allocators. Taking CPGuard is an extremely light weight operation and IMO we don't want to optimize, rather like to have clean code.

src/lib/blkalloc/bitmap_blk_allocator.cpp
```diff
@@ -0,0 +1,201 @@
...
+    }
+}
+
+void BitmapBlkAllocator::do_init() {
+    // NOTE:  Blocks per portion must be modulo word size so locks do not fall on same word
+    m_blks_per_portion = sisl::round_up(m_blks_per_portion, m_disk_bm ? m_disk_bm->word_size() : 64u);
+
+    m_blk_portions = std::make_unique< BlkAllocPortion[] >(get_num_portions());
+    for (blk_num_t index{0}; index < get_num_portions(); ++index) {
+        m_blk_portions[index].set_portion_num(index);
```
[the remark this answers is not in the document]
hkadayam: We have removed the set_available_nblks() altogether, because after restart it is not consistent and we don't load the correct number of blks available. So it doesn't make sense to do that.

src/lib/blkalloc/varsize_blk_allocator.cpp
```diff
@@ -232,13 +231,15 @@ bool VarsizeBlkAllocator::allocator_state_machine() {
     return active_state;
 }
 
-void VarsizeBlkAllocator::inited() {
-    m_cache_bm->copy(*(get_disk_bm_const()));
-    BlkAllocator::inited();
+void VarsizeBlkAllocator::load() {
+    m_cache_bm->copy(*get_disk_bitmap());
```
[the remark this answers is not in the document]
hkadayam: load() is not called for non persistent case, but I will add an assert here to catch that condition

src/include/homestore/homestore.hpp
```diff
@@ -154,6 +153,7 @@ class HomeStore {
 
     // cap_attrs get_system_capacity() const; // Need to move this to homeblks/homeobj
     bool is_first_time_boot() const;
+    bool is_initializing() const { return !m_init_done; }
```
hkadayam, line 156: This call is used by all subsystems/chunks instead of each of them maintaining is_inited flag and we need to walk through all chunks to set this after initialization. This is especially bad for Object storage since it can have 64K chunks.

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -119,6 +119,12 @@ BlkAllocStatus AppendBlkAllocator::alloc(blk_count_t nblks, const blk_alloc_hint
     return BlkAllocStatus::SUCCESS;
 }
 
+BlkAllocStatus AppendBlkAllocator::alloc_on_disk(BlkId const&) { return BlkAllocStatus::SUCCESS; }
+
+void AppendBlkAllocator::free_on_disk(BlkId const&) {}
+
+bool AppendBlkAllocator::is_blk_alloced_on_disk(BlkId const&, bool) const { return false; }
+
```
hkadayam, line 127: These APIs are necessary for chunk to be unaware of which blk allocator

src/lib/blkalloc/blk_allocator.h
```diff
@@ -65,37 +63,28 @@ struct BlkAllocConfig {
...
+    const bool m_persistent{false};
     const std::string m_unique_name;
-    bool m_auto_recovery{false};
-    bool m_realtime_bm_on{false}; // only specifically turn off in BlkAlloc Test;
 
 public:
-    BlkAllocConfig(uint32_t blk_size, uint32_t align_size, uint64_t size, const std::string& name = "",
-                   bool realtime_bm_on = true) :
+    BlkAllocConfig(uint32_t blk_size, uint32_t align_size, uint64_t size, bool persistent,
+                   const std::string& name = "") :
```
hkadayam, line 71: Removed realtime_bm_on, since everything is folded inside VDev. Will introduce similar functionality for debugging on subsequent PRs.

src/lib/blkalloc/blk_allocator.h
```diff
@@ -65,37 +63,28 @@ struct BlkAllocConfig {
...
     BlkAllocConfig& operator=(BlkAllocConfig const&) = default;
     BlkAllocConfig& operator=(BlkAllocConfig&&) noexcept = delete;
     virtual ~BlkAllocConfig() = default;
-    void set_auto_recovery(bool is_auto_recovery) { m_auto_recovery = is_auto_recovery; }
 
     virtual std::string to_string() const {
-        return fmt::format("BlkSize={} TotalBlks={} BlksPerPortion={} auto_recovery={}", in_bytes(m_blk_size),
-                           in_bytes(m_capacity), m_blks_per_portion, m_auto_recovery);
+        return fmt::format("BlkSize={} TotalBlks={} BlksPerPortion={} persistent={}", in_bytes(m_blk_size),
+                           in_bytes(m_capacity), m_blks_per_portion, m_persistent);
```
hkadayam, line 87: From allocator perspective the term `auto_recovery` doesn't mean much, but it cares whether the bitmap is persistent or not.

src/lib/blkalloc/blk_cache.h
```diff
@@ -156,7 +156,24 @@ struct blk_cache_refill_status {
...
+    constexpr auto parse(ParseContext& ctx) {
+        return ctx.begin();
+    }
+
+    template < typename FormatContext >
+    auto format(const homestore::blk_cache_refill_status& s, FormatContext& ctx) {
+        return format_to(ctx.out(), "{}/{}", s.slab_refilled_count, s.slab_required_count);
+    }
+};
+} // namespace fmt
```
hkadayam, line 174: Necessary to support C++-20

src/lib/blkdata_svc/blkdata_service.cpp
yamingk: Taking additional cpguard is fine. 
With this change, free_blk won't take effect in next cp_flush for appendblkallocator, but for next cp_flush, since appendblkallocator::free() will take cpgurad under vdev::cp_flush() (which should be fine, right?), and it will be flushed until next cp, which is not a big deal.

But it seems accumulating the free blk list is unnecessary for appendblkallocator, which seems a little odd.

hkadayam: I see, yes we don't want to accumulate that for append_blk_allocator. But without adding additional parameters such as do_instant_free, we can't do that, let me see how to overcome this. Thanks!

yamingk: How about let vdev::free_blk take cpguard and differentiate whether to accumulate or not based on allocator type?

src/lib/blkalloc/bitmap_blk_allocator.cpp
```diff
@@ -0,0 +1,199 @@
...
+    if (list) {
+        // cp has started, accumulating to the list
+        list->push_back(bid);
+    } else {
+        auto set_on_disk_bm = [this](auto& b) {
+            BlkAllocPortion& portion = blknum_to_portion(b.blk_num());
+            {
+                auto lock{portion.portion_auto_lock()};
+                if (!hs()->is_initializing()) {
+                    // During recovery we might try to free the entry which is already freed while replaying the
```
yamingk, line 99: free -> alloc ?

src/lib/device/virtual_dev.cpp
```diff
@@ -262,26 +252,26 @@ BlkAllocStatus VirtualDev::alloc_blks_from_chunk(blk_count_t nblks, blk_alloc_hi
...
-    return chunk->blk_allocator_mutable()->free_on_realtime(b);
-}*/
+void VirtualDev::free_blk(BlkId const& bid, VDevCPContext* vctx) {
+    auto do_free_action = [this](auto const& b, VDevCPContext* vctx) {
+        if (vctx && (m_allocator_type != blk_allocator_type_t::append)) {
+            // We don't want to accumulate here for append blk allocator.
+            vctx->m_free_blkid_list.push_back(b);
+        } else {
+            BlkAllocator* allocator = m_dmgr.get_chunk_mutable(b.chunk_num())->blk_allocator_mutable();
+            if (m_auto_recovery) { allocator->free_on_disk(b); }
```
yamingk, line 262: nit: is this needed since it does nothing.

yamingk (dismissed):

yamingk:
LGTM except two minor comments

src/lib/device/virtual_dev.cpp
hkadayam: It does free_on_disk on cases where we use persistent blkallocator, but not part of CP. So this code is needed.

src/lib/blkalloc/bitmap_blk_allocator.cpp
hkadayam: Done

src/lib/device/virtual_dev.cpp
yamingk: I see. It also fall through to non-append blkalloctor with vctx equal to nullptr;

yamingk (approved):

PR #216, opened 2023-10-27
Add `get_freeable_nblks` and `get_defrag_nblks` api to Vchunk

JacksonYao287:
the two apis will be used by HeapChunkSelector to select a best candidate chunk for GC

### DIFF
--- a/src/lib/blkalloc/varsize_blk_allocator.cpp
+++ b/src/lib/blkalloc/varsize_blk_allocator.cpp
@@ -731,6 +731,17 @@ bool VarsizeBlkAllocator::is_blk_alloced(BlkId const& bid, bool use_lock) const 
 }
 
 blk_num_t VarsizeBlkAllocator::available_blks() const { return get_total_blks() - get_used_blks(); }
+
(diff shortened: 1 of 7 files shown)


src/lib/blkalloc/varsize_blk_allocator.cpp
```diff
@@ -731,6 +731,17 @@ bool VarsizeBlkAllocator::is_blk_alloced(BlkId const& bid, bool use_lock) const
 }
 
 blk_num_t VarsizeBlkAllocator::available_blks() const { return get_total_blks() - get_used_blks(); }
+
+blk_num_t VarsizeBlkAllocator::get_freeable_nblks() const {
+    // TODO: implement this
```
yamingk, line 736: Can we put an assert here to indicate we are not supposed to arrive here? It can help unexpected error.
Same for fixed blk allocator.

JacksonYao287: ACK

yamingk (approved):

yamingk:
LGTM, thanks.

PR #220, opened 2023-11-01
Issue 187#: Long duration test for data service

yamingk:
1. add long duration test for data service with random read, write and free. 
2. fixed one issue in MultiBlkId iterator
3. add/update api comments for data service

Testing:
======
a few hundred iterations run for the mixed io load (num_io=300) test case and passed. --- which included in conan build test.
Long duration run for a few hours and still running.

### DIFF
--- a/src/lib/blkalloc/blk.cpp
+++ b/src/lib/blkalloc/blk.cpp
@@ -127,8 +127,7 @@ std::string MultiBlkId::to_string() const {
     std::string str = "MultiBlks: {";
     auto it = iterate();
     while (auto const b = it.next()) {
-        str += b->to_string();
-        str += " ";
+        str += (b->to_string() + " ");
     }
     str += std::string("}");
     return str;
--- a/src/lib/blkalloc/blk.cpp
+++ b/src/lib/blkalloc/blk.cpp
@@ -128,6 +128,7 @@ std::string MultiBlkId::to_string() const {
     auto it = iterate();
     while (auto const b = it.next()) {
         str += b->to_string();
+        str += " ";
     }
     str += std::string("}");
     return str;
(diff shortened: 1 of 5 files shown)


src/lib/blkalloc/blk.cpp
```diff
@@ -128,6 +128,7 @@ std::string MultiBlkId::to_string() const {
     auto it = iterate();
     while (auto const b = it.next()) {
         str += b->to_string();
+        str += " ";
```
hkadayam, line 131: nit: could fold this into one line by 
```str += b->to_string() + " "```

File: src/tests/test_data_service.cpp [hunk 2]
```diff
@@ -56,13 +60,15 @@ constexpr uint64_t Mi{Ki * Ki};
...
+    uint64_t num_io{1000};
+    uint64_t run_time{200};        // in seconds
+    uint32_t min_io_size{4 * Ki};  // blk_size aligned;
+    uint32_t max_io_size{32 * Mi}; // blk_size aligned;
 };
 
 static Param gp;
 
-ENUM(DataSvcOp, uint8_t, WRITE, READ, FREE_BLK, COMMIT_BLK, RESERVE_STREAM, ALLOC_STREAM, FREE_STREAM)
+ENUM(DataSvcOp_t, uint8_t, async_alloc_write = 1, async_read = 2, async_free = 3, max_op = 4);
```
hkadayam, line 71: If you want to allocate the value for it we should use `VENUM(..`. I don't see much benefit in assigning the value here.

yamingk: acked

src/lib/blkalloc/blk.cpp
yamingk: done

hkadayam (approved):

Issue #221, opened 2023-11-08
HomeStore Replication Service to report capacity stats for Consumers


yamingk:
DM needs information about each SM to provide total/used space for decision making of its load balancing. 

This API will be consumed by HomeObject and HomeObject will also provide API based on this to SM.

PR #222, opened 2023-11-08
issue #221: HomeStore to report capacity stats

yamingk:
HomeStore to provide used/total space which will be consumed by HomeObject/SM

Testing:
======
```
[11/07/23 18:14:40-07:00] [info] [test_solo_repl_dev] [206264] [test_solo_repl_dev.cpp:254:write_io] Before write, cap stats: used=0 total=1277747200
[11/07/23 18:14:40-07:00] [info] [test_solo_repl_dev] [206265] [test_solo_repl_dev.cpp:254:write_io] Before write, cap stats: used=0 total=1277747200
[11/07/23 18:14:40-07:00] [info] [test_solo_repl_dev] [206264] [test_solo_repl_dev.cpp:254:write_io] Before write, cap stats: used=8192 total=1277747200
[11/07/23 18:14:40-07:00] [info] [test_solo_repl_dev] [206265] [test_solo_repl_dev.cpp:254:write_io] Before write, cap stats: used=8192 total=1277747200
[11/07/23 18:14:40-07:00] [info] [test_solo_repl_dev] [206265] [test_solo_repl_dev.cpp:254:write_io] Before write, cap stats: used=16384 total=1277747200
[11/07/23 18:14:40-07:00] [info] [test_solo_repl_dev] [206264] [test_solo_repl_dev.cpp:254:write_io] Before write, cap stats: used=16384 total=1277747200
[11/07/23 18:14:40-07:00] [info] [test_solo_repl_dev] [206264] [test_solo_repl_dev.cpp:254:write_io] Before write, cap stats: used=24576 total=1277747200
[11/07/23 18:14:40-07:00] [info] [test_solo_repl_dev] [206264] [test_solo_repl_dev.cpp:254:write_io] Before write, cap stats: used=28672 total=1277747200
[11/07/23 18:14:40-07:00] [info] [test_solo_repl_dev] [206265] [test_solo_repl_dev.cpp:293:on_write_complete] Write complete with cap stats: used=32768 total=1277747200
[11/07/23 18:14:40-07:00] [info] [test_solo_repl_dev] [206264] [test_solo_repl_dev.cpp:254:write_io] Before write, cap stats: used=32768 total=1277747200
```

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -5,7 +5,7 @@ from conans import CMake
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "4.5.11"
+    version = "4.6.1"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/conanfile.py
+++ b/conanfile.py
@@ -5,7 +5,7 @@ from conans import CMake
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "4.5.10"
+    version = "4.5.11"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/include/homestore/replication_service.hpp
+++ b/src/include/homestore/replication_service.hpp
@@ -6,6 +6,7 @@
 
 #include <folly/futures/Future.h>
 
+#include <homestore/homestore.hpp>
 #include <homestore/replication/repl_decls.h>
 
 namespace homestore {
@@ -84,5 +85,8 @@ public:
     /// @brief Iterate over all repl devs and then call the callback provided
     /// @param cb Callback with repl dev
     virtual void iterate_repl_devs(std::function< void(cshared< ReplDev >&) > const& cb) = 0;
+
+    /// @brief Get the current term of the repl dev
+    virtual hs_stats get_cap_stats() const = 0;
 };
 } // namespace homestore
--- a/src/include/homestore/replication_service.hpp
+++ b/src/include/homestore/replication_service.hpp
@@ -86,7 +86,8 @@ public:
     /// @param cb Callback with repl dev
     virtual void iterate_repl_devs(std::function< void(cshared< ReplDev >&) > const& cb) = 0;
 
-    /// @brief Get the current term of the repl dev
+    /// @brief get the capacity stats form underlying backend;
+    /// @return the capacity stats;
     virtual hs_stats get_cap_stats() const = 0;
 };
 } // namespace homestore
--- a/src/include/homestore/replication_service.hpp
+++ b/src/include/homestore/replication_service.hpp
@@ -6,7 +6,6 @@
 
 #include <folly/futures/Future.h>
 
-#include <homestore/homestore.hpp>
 #include <homestore/replication/repl_decls.h>
 
 namespace homestore {
@@ -32,6 +31,7 @@ VENUM(ReplServiceError, int32_t,
 
 class ReplDev;
 class ReplDevListener;
+struct hs_stats;
 
 template < typename V, typename E >
 using Result = folly::Expected< V, E >;
--- a/src/lib/replication/service/repl_service_impl.cpp
+++ b/src/lib/replication/service/repl_service_impl.cpp
@@ -18,6 +18,7 @@
 #include "replication/service/repl_service_impl.h"
 #include "replication/repl_dev/solo_repl_dev.h"
 #include "homestore/blkdata_service.hpp"
+#include "homestore/homestore.hpp"
 
 namespace homestore {
 ReplicationService& repl_service() { return hs()->repl_service(); }
--- a/src/lib/replication/service/repl_service_impl.cpp
+++ b/src/lib/replication/service/repl_service_impl.cpp
@@ -17,6 +17,7 @@
 #include "common/homestore_assert.hpp"
 #include "replication/service/repl_service_impl.h"
 #include "replication/repl_dev/solo_repl_dev.h"
+#include "homestore/blkdata_service.hpp"
 
 namespace homestore {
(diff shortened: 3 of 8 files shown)


conanfile.py
```diff
@@ -5,7 +5,7 @@
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "4.5.10"
+    version = "4.5.11"
```
szmyd, line 8: This should be a MINOR bump (4.6.1) since we adding an new interface call. That way the consumer knows not to try to use something prior to `4.6.x`

yamingk: done

src/lib/replication/service/repl_service_impl.cpp
```diff
@@ -46,6 +47,14 @@ void ReplicationServiceImpl::stop() {
     m_rd_map.clear();
 }
 
+hs_stats ReplicationServiceImpl::get_cap_stats() const {
```
xiaoxichen, line 50: we should be able to get total/used space of a particular repl_dev(a.k.a PG).
As we limit one PG to one pdev in HO and that information is only available in there,  we may better handle the information there?

xiaoxichen (changes requested):
We need some more clarification on the capacity stats.

From SM->CM/DM perspective, we need
1. capacity stats PER PG.
2. capacity stats of whole SM/HS.

As in HO we enforce one PDEV per PG policy and more general the HO take control of the chunk selection,  it makes more sense to leave 1) to HO as HS doesnt have any information regarding allocation policy.

This PR on HS is more on 2) and actually it returns the available/total capacity of the whole data service,  not individual repl_dev/PG.  In this sense , this API should not be put into ReplicationService,  but just `HomeStore::instance().getCapacityStat()`

src/include/homestore/replication_service.hpp
```diff
@@ -6,6 +6,7 @@
 
 #include <folly/futures/Future.h>
 
+#include <homestore/homestore.hpp>
```
raakella1, line 9: Do we need the header here? We can move it to the cpp file

raakella1:
> We need some more clarification on the capacity stats.
> 
> From SM->CM/DM perspective, we need
> 
> 1. capacity stats PER PG.
> 2. capacity stats of whole SM/HS.
> 
> As in HO we enforce one PDEV per PG policy and more general the HO take control of the chunk selection, it makes more sense to leave 1) to HO as HS doesnt have any information regarding allocation policy.
> 
> This PR on HS is more on 2) and actually it returns the available/total capacity of the whole data service, not individual repl_dev/PG. In this sense , this API should not be put into ReplicationService, but just `HomeStore::instance().getCapacityStat()`

@xiaoxichen SM needs 2) as well. In fact, the current heartbeat request proto only includes 2) 
https://github.corp.ebay.com/SDS/nuobject_proto/blob/main/src/proto/cm_svc.proto#L12
and the storage info is the total stats.  
https://github.corp.ebay.com/SDS/nuobject_proto/blob/main/src/proto/sm_svc.proto#L14
So, this has to be part of replication service for HO to access it and pass it to SM above

src/lib/replication/service/repl_service_impl.cpp
yamingk: We thought about doing it in PG level, but current nuobject proto buffer definition only need total/used per SM. 

We can explore the needs for per PG level stats if there is a use case.

src/include/homestore/replication_service.hpp
yamingk: done.

raakella1 (approved):
LG!

yamingk:
@xiaoxichen The best place to get Per PG stats (if we were to have it) is from HomeObject, as HomeStore don't have knowledge about PG. And HomeObj heapChunkSel maitains per PG chunk list, vchunk can query underlying blkallocator to get total/used stats.

xiaoxichen:
@yamingk  @raakella1  thanks for clarification and Yes we need #2.
But if we are collecting the metrics of the whole SM, do we really want to put this into the ReplicationService interface? Or maybe better just put it into `class HomeStore `

I was got confused as it put into ReplicationService so I was assume it is doing something related with PG.  And yes as @yamingk said the PG stats need to be in HO.

xiaoxichen:
> > We need some more clarification on the capacity stats.
> > From SM->CM/DM perspective, we need
> > 
> > 1. capacity stats PER PG.
> > 2. capacity stats of whole SM/HS.
> > 
> > As in HO we enforce one PDEV per PG policy and more general the HO take control of the chunk selection, it makes more sense to leave 1) to HO as HS doesnt have any information regarding allocation policy.
> > This PR on HS is more on 2) and actually it returns the available/total capacity of the whole data service, not individual repl_dev/PG. In this sense , this API should not be put into ReplicationService, but just `HomeStore::instance().getCapacityStat()`
> 
> @xiaoxichen SM needs 2) as well. In fact, the current heartbeat request proto only includes 2) https://github.corp.ebay.com/SDS/nuobject_proto/blob/main/src/proto/cm_svc.proto#L12 and the storage info is the total stats. https://github.corp.ebay.com/SDS/nuobject_proto/blob/main/src/proto/sm_svc.proto#L14 So, this has to be part of replication service for HO to access it and pass it to SM above

https://github.corp.ebay.com/SDS/nuobject_proto/blob/main/src/proto/pg.proto#L28
The per PG stat needed here,  we may need avail_bytes into the proto as when we define the proto we assumed all PGs share the entire SM so PG do not have a capacity. But after we apply one to one mapping it makes senes to let PG have a avail_bytes.
As you said this need to be done on HO and not part of this PR/

yamingk:
> ReplicationService

@xiaoxichen ReplicationService Instance is a singletone.

xiaoxichen (approved):

PR #230, opened 2023-11-14
Next set of code to get the raft based replication.

hkadayam:
Major changes in this PR i
* Foundational code and framework for Raft replication with RaftReplService override
and RaftReplDev/RaftStateMachine

* It does have barebone code and no testing related to the Raft code, but at this time
it compiles and existing test cases run successfully.


src/CMakeLists.txt
```diff
@@ -8,12 +8,15 @@ find_package(isa-l REQUIRED)
...
 find_package(GTest REQUIRED)
+find_package(nuraft_mesg REQUIRED)
 
 set (COMMON_DEPS
         iomgr::iomgr
         farmhash::farmhash
         isa-l::isa-l
         sisl::sisl
+        nuraft::nuraft
+        nuraft_mesg::nuraft_mesg
```
szmyd, line 19: NuraftMesg::proto

src/include/homestore/homestore_decl.hpp
```diff
@@ -198,7 +198,7 @@ struct cap_attrs {
 
 ////////////// Misc ///////////////////
 #define HOMESTORE_LOG_MODS                                                                                             \
-    btree_structures, btree_nodes, btree_generics, btree, cache, device, blkalloc, vol_io_wd, volume, flip, cp,        \
-        metablk, indx_mgr, wbcache, logstore, replay, transient, IOMGR_LOG_MODS
+    btree, device, blkalloc, flip, cp, metablk, wbcache, logstore, transient, replication, nuraft, nuraft_mesg,        \
+        grpc_server, IOMGR_LOG_MODS
```
szmyd, line 202: `NURAFTMESG_LOG_MODS` will cover `nuraft`, `nuraft_mesg` and `grpc_server`

src/lib/replication/log_store/endian_encode.h
```diff
@@ -0,0 +1,86 @@
...
+#define __ENDIAN_SAFE
+#ifdef __ENDIAN_SAFE
+#define _enc(v)                                                                                                        \
+    ((sizeof(v) == 8) ? (_enc64(v)) : ((sizeof(v) == 4) ? (_enc32(v)) : ((sizeof(v) == 2) ? (_enc16(v)) : (v))))
+#define _dec(v)                                                                                                        \
+    ((sizeof(v) == 8) ? (_dec64(v)) : ((sizeof(v) == 4) ? (_dec32(v)) : ((sizeof(v) == 2) ? (_dec16(v)) : (v))))
+#else
+#define _enc(v) (v)
+#define _dec(v) (v)
+#endif
```
szmyd, line 86: Why are we not using POSIX calls to do these conversions? (https://linux.die.net/man/3/ntohl)

src/lib/replication/CMakeLists.txt
```diff
@@ -5,27 +5,22 @@ include_directories (BEFORE .)
 
 flatbuffers_generate_headers(
     TARGET hs_replication_fb 
-    SCHEMAS rpc/push_data_rpc.fbs
+    SCHEMAS push_data_rpc.fbs
     FLAGS "--cpp"
```
szmyd, line 9: Do you want to enable c++17 extensions?

src/lib/replication/repl_dev/mapper.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include <sisl/logging/logging.h>
+#include <sisl/fds/utils.hpp>
+#include <sisl/fds/vector_pool.hpp>
+
+#include "repl_dev/raft_state_machine.h"
+#include "repl_dev/raft_repl_dev.h"
+
+namespace homestore {
+repl_req_ptr_t RaftReplDev::follower_create_req(repl_key const& rkey, sisl::blob const& user_header,
+                                                sisl::blob const& user_key) {
```
zichanglai, line 11: do we need to remove this mapper.cpp? it seems follower_create_req/notify_after_data_written functions are duplicated with the implementation in raft_repl_dev.cpp and map_fetch_write_pop() is never used.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -0,0 +1,379 @@
...
+    MultiBlkId blkid;
+    auto const status = data_service().alloc_blks(sisl::round_up(size, data_service().get_blk_size()), hints, blkid);
+    RELEASE_ASSERT_EQ(status, BlkAllocStatus::SUCCESS, "alloc_blks returned null, no space left!");
+    return blkid;
+}
+
+repl_req_ptr_t RaftReplDev::follower_create_req(repl_key const& rkey, sisl::blob const& user_header,
+                                                sisl::blob const& user_key, uint32_t data_size) {
+    auto const [it, happened] = m_repl_key_req_map.try_emplace(rkey, repl_req_ptr_t(new repl_req_ctx()));
+    RD_DBG_ASSERT((it == m_repl_key_req_map.end()), "Unexpected error in map_repl_key_to_req");
```
zichanglai, line 146: it != m_repl_key_req_map.end()

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -0,0 +1,510 @@
...
+uint64_t RaftStateMachine::last_commit_index() { return uint64_cast(m_rd.get_last_commit_lsn()); }
+
+void RaftStateMachine::propose_to_raft(repl_req_ptr_t rreq) {
+    uint32_t entry_size = sizeof(repl_journal_entry) + rreq->header.size() + rreq->key.size() +
+        (rreq->value.size ? rreq->local_blkid.serialized_size() : 0);
+    rreq->alloc_journal_entry(entry_size, true /* raft_buf */);
+    rreq->journal_entry->code = (rreq->value.size) ? journal_type_t::HS_LARGE_DATA : journal_type_t::HS_HEADER_ONLY;
+    rreq->journal_entry->user_header_size = rreq->header.size();
+    rreq->journal_entry->key_size = rreq->key.size();
+    rreq->journal_entry->dsn = rreq->dsn();
```
zichanglai, line 67: forget to set rreq->journal_entry->value_size?

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -0,0 +1,510 @@
...
+
+    // From the repl_key, get the repl_req. In cases where log stream got here first, this method will create a new
+    // repl_req and return that back. Fill up all of the required journal entry inside the repl_req
+    auto rreq = m_rd.follower_create_req(
+        repl_key{.server_id = jentry->server_id, .term = lentry->get_term(), .dsn = jentry->dsn}, header, key,
+        jentry->value_size);
+    rreq->journal_buf = lentry->serialize();
+
+    MultiBlkId entry_blkid;
+    entry_blkid.deserialize(sisl::blob{key.cbytes() + key.size(), jentry->value_size}, true /* copy */);
```
zichanglai, line 121: jentry->value_size is the data value size, right? actually what we need here is the PBA serialize size.

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -0,0 +1,510 @@
...
+raft_buf_ptr_t RaftStateMachine::commit_ext(nuraft::state_machine::ext_op_params const& params) {
+    int64_t lsn = s_cast< int64_t >(params.log_idx);
+    raft_buf_ptr_t data = params.data;
+
+    RD_LOG(DEBUG, "apply_commit: {}, size: {}", lsn, data->size());
+
+    repl_req_ptr_t rreq = lsn_to_req(lsn);
+    if (m_rd.is_leader()) {
+        // This is the time to ensure flushing of journal happens in leader
+        if (m_rd.m_data_journal->last_durable_index() < uint64_cast(lsn)) { m_rd.m_data_journal->flush(); }
```
zichanglai, line 48: I am little confused about this behaviors:

1. will journal log flush only happens when it can be commited? 
2. why only happens in leader side? follower need flush the journal log too, right?

src/CMakeLists.txt
hkadayam: Will do.

src/lib/replication/log_store/endian_encode.h
hkadayam: This is not new code, but copied NuRaft code to make it run. Didn't want to invest time on that :-)

src/lib/replication/CMakeLists.txt
hkadayam: Will do.

src/lib/replication/service/generic_repl_svc.cpp
```diff
@@ -33,7 +33,8 @@ std::unique_ptr< GenericReplService > GenericReplService::create(cshared< ReplAp
     }
 }
 
-GenericReplService::GenericReplService(cshared< ReplApplication >& repl_app) : m_repl_app{repl_app} {
+GenericReplService::GenericReplService(cshared< ReplApplication >& repl_app) :
```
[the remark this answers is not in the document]
hkadayam: Yes, but I will do that as part of my testing effort.

src/lib/replication/service/raft_repl_service.h
```diff
@@ -0,0 +1,78 @@
...
+    std::string lookup_peer(nuraft_mesg::peer_id_t const&) override;
+    std::shared_ptr< nuraft_mesg::mesg_state_mgr > create_state_mgr(int32_t srv_id,
+                                                                    nuraft_mesg::group_id_t const& group_id) override;
+    shared< nuraft_mesg::Manager > msg_manager() { return m_msg_mgr; }
+
+protected:
+    ///////////////////// Overrides of GenericReplService ////////////////////
+    void start() override;
+    AsyncReplResult< shared< ReplDev > > create_repl_dev(uuid_t group_id,
+                                                         std::set< uuid_t, std::less<> >&& members) override;
```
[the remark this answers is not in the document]
hkadayam: We don't need to implement create_repl_dev. I will remove this declaration.

src/lib/replication/repl_dev/mapper.cpp
hkadayam: Yes I thought I removed it in next commit, if not, will remove that.

src/lib/replication/repl_dev/raft_repl_dev.cpp
hkadayam: Good catch, will fix that.

src/lib/replication/repl_dev/raft_state_machine.cpp
hkadayam: Few points to note:
The data_journal->flush() call here ensures that logs up to now has been flushed. But it might have been already flushed because of timer or other group commit, in that case will not do anything. In general you can call this method as late as possible.

To your question:
1. Leader will need to ensure flushing only after the Raft confirms it is committed. Until that time, it is not required to be flushed, because after we commit, we need to record the commit lsn in the superblk (in-memory initially), before we respond to the user. So it is right to flush here.

2. Follower will flush on_pre_commit itself. Without flushing if it reports to the leader that it is precommitted and then crashed (and it happened on say all replicas), leader will go ahead and commit and respond to user, only later that data is lost. So we must flush during on_pre_commit. This is done `ReplLogStore::end_of_append_batch()`

Now regarding changes to be done:
This reminds me a change, I was thinking of doing, which is instead of relying on current state of Replica Set from RAFT (if its a leader or follower), I will track this in request structure itself (to make this decision of leader/follower). This way we don't need to worry about role changing in the middle. Will make that change in next commit.

src/lib/replication/repl_dev/raft_state_machine.cpp
hkadayam: Will check that and add if it is missing. Thanks for the catch.

src/lib/replication/service/generic_repl_svc.cpp
```diff
@@ -68,17 +69,22 @@ AsyncReplResult< shared< ReplDev > > GenericReplService::create_repl_dev(uuid_t
...
-    if (!bool(result)) { return make_async_error< shared< ReplDev > >(result.error()); }
-
     // Now we need to create local repl dev instance which is actually creates entry, state machine etc
     // according the the underlying implementation
     superblk< repl_dev_superblk > rd_sb{get_meta_blk_name()};
     rd_sb.create(rd_super_blk_size());
     rd_sb->group_id = group_id;
 
     auto repl_dev = create_local_repl_dev_instance(rd_sb, false /* load_existing */);
+
```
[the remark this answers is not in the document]
hkadayam: I think you are worried about rd_sb being copied inside the `create_local_repl_dev_instance` and this and other method are working on 2 different in-memory versions, right? If so that is an excellent catch. 

I will disable copy constructor and `=` operator on superblk_handler as it can happen even otherwise. So will do a move only and we shall move the superblk and access them from repl_dev->superblk() as reference. In that case, cp_flush() will not do rd_sb write() until this method finishes rd_sb.write() as repl_dev is not added into the map yet.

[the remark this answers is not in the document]
zichanglai: yes, it will have 2 different in-memory versions, but that is not the point I was worried.  what I worries is two different version super_block will allocate 2 metablks on HS meta service when they are called with write() and we need recovery twice during restart.  that is because member m_meta_mgr_cookie in the both super_block is still nullptr and sb.write() function will allocate metablk instead of using the exist one.

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -0,0 +1,510 @@
...
+        RD_LOG(DEBUG, "pre_commit: {}, size: {}", lsn, data->size());
+        repl_req_ptr_t rreq = lsn_to_req(lsn);
+
+        m_rd.m_listener->on_pre_commit(rreq->lsn, rreq->header, rreq->key, rreq);
+    }
+    return m_success_ptr;
+}
+
+void RaftStateMachine::after_precommit_in_leader(nuraft::raft_server::req_ext_cb_params const& params) {
+    repl_req_ptr_t rreq = repl_req_ptr_t(r_cast< repl_req_ctx* >(params.context));
```
JacksonYao287, line 34: log is stored before `pre_commit_ext`

https://github.com/eBay/NuRaft/blob/ff3d2c7a8d73754d3c09dc4b26b7968181f5bcf6/src/handle_client_request.cxx#L135
`params.context` seems not changed between `pre_commit_ext` and `after_pre_commit`, so lsn should be known in `pre_commit_ext`?

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -0,0 +1,268 @@
...
+                            fmt::vformat_to(fmt::appender{buf}, fmt::string_view{"[{}={}] "},                          \
+                                            fmt::make_format_args("replstore", m_logstore_id));                        \
+                            fmt::vformat_to(fmt::appender{buf}, fmt::string_view{msgcb},                               \
+                                            fmt::make_format_args(std::forward< decltype(args) >(args)...));           \
+                            return true;                                                                               \
+                        }),                                                                                            \
+                        msg, ##__VA_ARGS__);
+
+namespace homestore {
+static constexpr store_lsn_t to_store_lsn(uint64_t raft_lsn) { return s_cast< store_lsn_t >(raft_lsn) - 1; }
```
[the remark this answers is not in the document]
hkadayam: Homestore logstore id is int64_t because it uses streamtracker which is int64_t. It is 0 based index and first invalid id is -1 (and not 0). 0 is a valid lsn, whereas raft 0 is an invalid lsn.

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -0,0 +1,268 @@
...
+static constexpr store_lsn_t to_store_lsn(repl_lsn_t repl_lsn) { return repl_lsn - 1; }
```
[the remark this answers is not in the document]
hkadayam: For reasons mentioned above we can't have that. Raft unfortunately uses 1 as starting lsn index. Ideally it should make it 0, so that it works inside logstore most efficiently to store (in vector etc).

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -0,0 +1,379 @@
...
+static MultiBlkId do_alloc_blk(uint32_t size, blk_alloc_hints const& hints) {
+    MultiBlkId blkid;
+    auto const status = data_service().alloc_blks(sisl::round_up(size, data_service().get_blk_size()), hints, blkid);
+    RELEASE_ASSERT_EQ(status, BlkAllocStatus::SUCCESS, "alloc_blks returned null, no space left!");
+    return blkid;
+}
+
+repl_req_ptr_t RaftReplDev::follower_create_req(repl_key const& rkey, sisl::blob const& user_header,
+                                                sisl::blob const& user_key, uint32_t data_size) {
+    auto const [it, happened] = m_repl_key_req_map.try_emplace(rkey, repl_req_ptr_t(new repl_req_ctx()));
```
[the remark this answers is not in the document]
hkadayam: This map is folly::ConcurrentHashmap

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -0,0 +1,268 @@
...
+static constexpr repl_lsn_t to_repl_lsn(store_lsn_t store_lsn) { return store_lsn + 1; }
+
+static nuraft::ptr< nuraft::log_entry > to_nuraft_log_entry(const log_buffer& log_bytes) {
+    uint8_t const* raw_ptr = log_bytes.bytes();
+    uint64_t term = *r_cast< uint64_t const* >(raw_ptr);
+    raw_ptr += sizeof(uint64_t);
+    nuraft::log_val_type type = static_cast< nuraft::log_val_type >(*raw_ptr);
+    raw_ptr += sizeof(uint8_t);
```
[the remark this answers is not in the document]
hkadayam: Done

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -0,0 +1,268 @@
...
+void HomeRaftLogStore::write_at(ulong index, nuraft::ptr< nuraft::log_entry >& entry) {
+    auto buf = entry->serialize();
+    write_at(index, buf);
+}
+
+void HomeRaftLogStore::write_at(ulong index, raft_buf_ptr_t& buffer) {
+    m_log_store->rollback_async(to_store_lsn(index) - 1, nullptr);
+    // we need to reset the durable lsn, because its ok to set to lower number as it will be updated on next flush
+    // calls, but it is dangerous to set higher number.
+    m_last_durable_lsn = -1;
```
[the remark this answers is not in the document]
hkadayam: It is ok to call `get_contiguous_issued_seq_num(-1)`, just a little inefficient, since it has to search for bits which are set.

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -0,0 +1,268 @@
...
+    pack.pos(0);
+    auto num_entries = pack.get_int();
+
+    auto slot = next_slot();
+    if (index < slot) {
+        // We are asked to apply/insert data behind next slot, so we must rollback before index and then append
+        m_log_store->rollback_async(to_store_lsn(index) - 1, nullptr);
+    } else if (index > slot) {
+        // We are asked to apply/insert data after next slot, so we need to fill in with dummy entries upto the slot
+        // before append the entries
```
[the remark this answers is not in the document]
hkadayam: Those dummy entries are not touched upon replay. There is no option of failing to apply the pack. We have only 2 options, apply or crash, but upon coming back it will possibly do the same thing. This is an error in the system and we simply log, work it around.

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    for (int i{0}; i < num_entries; ++i) {
+        size_t entry_len;
+        auto* entry = const_cast< nuraft::byte* >(pack.get_bytes(entry_len));
+        [[maybe_unused]] auto store_sn =
+            m_log_store->append_async(sisl::io_blob{entry, uint32_cast(entry_len), false}, nullptr, nullptr);
+        REPL_STORE_LOG(TRACE, "unpacking nth_entry={} of size={}, lsn={}", i + 1, entry_len, to_repl_lsn(store_sn));
+    }
+    m_log_store->flush_sync(to_store_lsn(index) + num_entries - 1);
```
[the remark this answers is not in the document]
hkadayam: flush_sync needs upto (inclusive)
https://github.com/eBay/HomeStore/blob/b202457b9bcb95921c2c77fa76be0e3b504ea9e2/src/include/homestore/logstore/log_store.hpp#L251

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -0,0 +1,268 @@
...
+bool HomeRaftLogStore::compact(ulong compact_lsn) {
+    auto cur_max_lsn = m_log_store->get_contiguous_issued_seq_num(m_last_durable_lsn);
+    if (cur_max_lsn < to_store_lsn(compact_lsn)) {
+        // We need to fill the remaining entries with dummy data.
+        for (auto lsn{cur_max_lsn + 1}; lsn <= to_store_lsn(compact_lsn); ++lsn) {
+            append(m_dummy_log_entry);
+        }
+    }
+    m_log_store->flush_sync(to_store_lsn(compact_lsn));
+    m_log_store->truncate(to_store_lsn(compact_lsn));
```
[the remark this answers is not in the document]
hkadayam: Logstore in homestore uses bitmap to track contiguous completion, truncation etc. Without dummy log entry we run into the hole and all contiguous completions are in trouble. This layer is to adapt to what homestore provides and what raft needs (and fill those gaps).

src/lib/replication/repl_dev/raft_state_machine.cpp
hkadayam: Relying on `read_sync` to ensure flush is done may not the right approach. Like you said our implementation of `entry_at` or even `read` without flush is on cards, so we must do a flush which is a no-op in those cases.

src/lib/replication/repl_dev/raft_state_machine.cpp
hkadayam: I think this has been take care now

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -0,0 +1,451 @@
...
+        // Write the data
+        data_service().async_write(rreq->value, rreq->local_blkid).thenValue([this, rreq](auto&& err) {
+            HS_REL_ASSERT(!err, "Error in writing data"); // TODO: Find a way to return error to the Listener
+            rreq->state.fetch_or(uint32_cast(repl_req_state_t::DATA_WRITTEN));
+            m_state_machine->propose_to_raft(std::move(rreq));
+            // if (rreq->state.load() & uint32_cast(repl_req_state_t::LOG_FLUSHED)) { report_committed(std::move(rreq));
+            // }
+        });
+    } else {
+        RD_LOG(INFO, "Skipping data channel send since value size is 0");
```
zichanglai, line 98: we need to support the case without value, right? so we still need to call m_state_machine->propose_to_raft() in here

src/lib/homestore.cpp
```diff
@@ -231,6 +231,11 @@ void HomeStore::shutdown() {
 
     LOGINFO("Homestore shutdown is started");
 
+    if (has_repl_data_service()) {
+        s_cast< GenericReplService* >(m_repl_service.get())->stop();
+        m_repl_service.reset();
+    }
+
     if (has_index_service()) {
         m_index_service->stop();
```
zichanglai, line 240: when indexservice stops, it will call CPManager to flush the CP, but RaftReplDev cp flush will crash if we reset the replicationservice in L236 before.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -0,0 +1,379 @@
...
+}
+
+void RaftReplDev::use_config(json_superblk raft_config_sb) { m_raft_config_sb = std::move(raft_config_sb); }
+
+void RaftReplDev::async_alloc_write(sisl::blob const& header, sisl::blob const& key, sisl::sg_list const& value,
+                                    repl_req_ptr_t rreq) {
+    if (!rreq) { auto rreq = repl_req_ptr_t(new repl_req_ctx{}); }
+    rreq->header = header;
+    rreq->key = key;
+    rreq->value = std::move(value);
```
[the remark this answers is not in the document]
hkadayam: Done

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -0,0 +1,379 @@
...
+        auto status = data_service().alloc_blks(uint32_cast(rreq->value.size),
+                                                m_listener->get_blk_alloc_hints(rreq->header, rreq->value.size),
+                                                rreq->local_blkid);
+        HS_REL_ASSERT_EQ(status, BlkAllocStatus::SUCCESS);
+
+        // Write the data
+        data_service().async_write(rreq->value, rreq->local_blkid).thenValue([this, rreq](auto&& err) {
+            HS_REL_ASSERT(!err, "Error in writing data"); // TODO: Find a way to return error to the Listener
+            rreq->state.fetch_or(uint32_cast(repl_req_state_t::DATA_WRITTEN));
+            if (rreq->state.load() & uint32_cast(repl_req_state_t::LOG_FLUSHED)) { report_committed(std::move(rreq)); }
```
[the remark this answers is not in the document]
hkadayam: Done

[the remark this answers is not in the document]
hkadayam: Code has been modified so that proposal to raft happens only after data has been written on primary.

src/lib/replication/repl_dev/raft_repl_dev.cpp
hkadayam: Yes, modified the code.

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -0,0 +1,510 @@
...
+
+namespace homestore {
+
+RaftStateMachine::RaftStateMachine(RaftReplDev& rd) : m_rd{rd}, m_group_id{m_rd.group_id()} {
+    m_success_ptr = nuraft::buffer::alloc(sizeof(int));
+    m_success_ptr->put(0);
+}
+
+raft_buf_ptr_t RaftStateMachine::pre_commit_ext(nuraft::state_machine::ext_op_params const& params) {
+    // Leader precommit is processed in next callback, since lsn would not have been known to repl layer till we get
```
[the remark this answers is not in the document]
hkadayam: I agree that comment is bit misleading, I changed the comment. The purpose of this callback is that `pre_commit_ext` callback params doesn't provide a way to stick the context. Hence it is useful in follower where we use lsn information to create a request structure at that time. However, in leader we create the req structure before and put all the values inside it. So we have another callback as part of `nuraft::raft_server::req_ext_params` where we can stick our context and callback, which is called only on the proposer (in our case only on leader)

src/lib/replication/repl_dev/raft_state_machine.cpp
hkadayam: As explained on Xioaxi's comments, the purpose is our ability to stick the context (which is req structure) on callback, which is not possible with pre_commit_ext. That is why we have another callback to put the context.

src/lib/homestore.cpp
hkadayam: Raised an issue https://github.com/eBay/HomeStore/issues/263 to track this issue.

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -0,0 +1,532 @@
...
+        // we've got some blkids not completed yet, let's fetch from remote;
+        fetch_blkid_data_from_leader(std::move(remote_fetch_blkids));
+    }
+}
+
+//
+// if return false /* no need to wait */, no cb will be triggered;
+// if return true /* need to wait */ , cb will be triggered after all local blkids completed writting;
+//
+bool RaftStateMachine::async_fetch_write_blkids(std::vector< RemoteBlkId > const& remote_blkids,
```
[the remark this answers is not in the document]
hkadayam: That has been removed. It is under #if 0, but my latest changed removed that.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -0,0 +1,451 @@
...
+    /*RD_LOG(INFO, "Data Channel: Pushing data to all followers: rreq=[{}] data=[{}]", rreq->to_string(),
+           flatbuffers::FlatBufferToString(builder.GetBufferPointer() + sizeof(flatbuffers::uoffset_t),
+                                           PushDataRequestTypeTable()));*/
+
+    RD_LOG(INFO, "Data Channel: Pushing data to all followers: rreq=[{}]", rreq->to_compact_string());
+
+    group_msg_service()
+        ->data_service_request_unidirectional(nuraft_mesg::role_regex::ALL, PUSH_DATA, rreq->pkts)
+        .via(&folly::InlineExecutor::instance())
+        .thenValue([this, rreq = std::move(rreq)](auto e) {
```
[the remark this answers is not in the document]
hkadayam: A move on const& would be treated as copy assignment I believe. Anyways it is not serving the purpose, I can make it copy by value and do a move.

PR #236, opened 2023-11-17
index service long running

shosseinimotlagh:
Long running test for index service runs on **86-05**.
To run the test on stability,  [homestore-btree-manual test ](https://github.corp.ebay.com/SDS/nublox-verify/pull/433) is added 

`launcher.py --setup_once --profile homestore-btree-manual -it dev `


szmyd (approved):
lgtm

PR #237, opened 2023-11-18
Index Btree Benchmark

JacksonYao287:
test btree types:

1 FixedLenBtree
2 VarKeySizeBtree
3 VarValueSizeBtree
4 VarObjSizeBtree
5 PrefixIntervalBtree

we can run benchmark from command line with different parameters to simulate different case , for example:
`./bin/index_btree_benchmark --num_threads=5 --num_fibers=5 --operation_list=put:100`
Screenshot 2023-12-12 at 20 26 40


shosseinimotlagh (changes requested):
1- You don't need put_random_in_parallel or other variation. the only function that is needed is BtreeConcurrentTest/*ConcurrentAllOps. 
2- apply changes for ops count (`m_num_ops`) in helper functions (specifically `run_in_parallel`) instead of rewriting them.

src/tests/index_btree_benchmark.cpp
```diff
@@ -0,0 +1,396 @@
...
+        iota(vec.begin(), vec.end(), 0);
+        std::random_shuffle(vec.begin(), vec.end());
+        std::condition_variable done_cv;
+        std::mutex done_mtx;
+        auto thread_count{num_threads};
+
+        for (uint i{0}; i < num_threads; ++i) {
+            auto start = i * num_entries;
+            auto end = (i + 1) * num_entries;
+            std::thread([&, this, start, end]() {
```
shosseinimotlagh, line 322: We also support fibers. Please use iomgr instead of std::thread

src/tests/index_btree_benchmark.cpp
```diff
@@ -0,0 +1,396 @@
...
+        std::vector< std::pair< K, V > > out_vector;
+        auto rqreq = BtreeQueryRequest< K >{BtreeKeyRange< K >{start_key, true, end_key, true}};
+        m_btree->query(rqreq, out_vector);
+    }
+
+    void put_random_in_parallel() {
+        run_in_parallel([this](uint32_t i) { do_put(i, btree_put_type::INSERT); });
+    }
+
+    void get_random_in_parallel() {
```
shosseinimotlagh, line 248: not needed! Nonetheless, concurrent ALL ops can serve this purpose

src/tests/index_btree_benchmark.cpp
JacksonYao287: ack

src/tests/index_btree_benchmark.cpp
```diff
@@ -0,0 +1,262 @@
...
+        this->m_bt = std::make_shared< typename T::BtreeType >(uuid, parent_uuid, 0, this->m_cfg);
+        hs()->index_service().add_index_table(this->m_bt);
+    }
+
+    void TearDown() {
+        BtreeHelper< TestType >::TearDown();
+        test_common::HSTestHelper::shutdown_homestore();
+    }
+
+    void ConcurrentOps(std::vector< std::string > input_ops) {
```
shosseinimotlagh, line 174: NIT: This part is needed to be defined in helper since here, membtree and index use it. I will do it after your PR is merged

src/tests/index_btree_benchmark.cpp
```diff
@@ -0,0 +1,262 @@
...
+                    // Invalid frequency, ignore this element
+                }
+            }
+            return std::make_pair(std::string(), 0);
+        });
+
+        this->run_in_parallel(ops);
+    }
+
+    void put() { this->ConcurrentOps({"put:100"}); }
```
shosseinimotlagh, line 208: Not sure if we need these functions?

JacksonYao287: the advantage we have a seperate put(), remove() and other functions is if we wan to take some actions more than just passing the operation_list, we can implement this logic inside these functions.

furthermore, if we want to implement a complex benchmark test ,  we just need to the logic in a seperate function and then add the items in the macro.

src/tests/index_btree_benchmark.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+template < typename TestType >
+struct IndexBtreeBenchmark : public BtreeHelper< TestType > {
+    using T = TestType;
+    using K = typename TestType::KeyType;
+    using V = typename TestType::ValueType;
+    class TestIndexServiceCallbacks : public IndexServiceCallbacks {
+    public:
+        TestIndexServiceCallbacks(IndexBtreeBenchmark* test) : m_benchmark(test) {}
+        std::shared_ptr< IndexTableBase > on_index_table_found(const superblk< index_table_sb >& sb) {
```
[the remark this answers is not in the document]
JacksonYao287: will remove

src/tests/btree_helpers/btree_test_helper.hpp
```diff
@@ -397,4 +402,36 @@ struct BtreeTestHelper : public testing::Test {
...
+                    return std::make_pair(tokens[0], (int)(100.0 * std::stoi(tokens[1]) / total));
+                } catch (const std::exception&) {
+                    // Invalid frequency, ignore this element
+                }
+            }
+            return std::make_pair(std::string(), 0);
+        });
+
+        return ops;
+        // this->run_in_parallel(ops);
```
shosseinimotlagh, line 435: NIT: remove the comment

shosseinimotlagh (dismissed):
LGTM

JacksonYao287:
test failure is caused by iomgr, not the test case itself

shosseinimotlagh (dismissed):

shosseinimotlagh (approved):

PR #238, opened 2023-11-21
fix HS metablk allocate duplicated blkId after restart

zichanglai:
HomeStore MetaBlkService is using varisize blk allocator to allocate new metablk, but for now the allocated bitmap is not persisted at all which will may cause HS re-allocated the same blkId after restarted.


src/lib/device/virtual_dev.cpp
```diff
@@ -145,9 +145,14 @@ bool VirtualDev::is_blk_alloced(BlkId const& blkid) const {
...
-BlkAllocStatus VirtualDev::commit_blk(BlkId const& blkid) {
+BlkAllocStatus VirtualDev::commit_blk(BlkId const& blkid, bool recovering) {
     Chunk* chunk = m_dmgr.get_chunk_mutable(blkid.chunk_num());
     HS_LOG(DEBUG, device, "commit_blk: bid {}", blkid.to_string());
+    if (!recovering) {
+        HS_DBG_ASSERT(is_blk_alloced(blkid), "commiting blkid {} not allocated", blkid.to_string());
+    } else {
+       chunk->blk_allocator_mutable()->mark_blk_allocated(blkid);
+    }
     return chunk->blk_allocator_mutable()->alloc_on_disk(blkid);
```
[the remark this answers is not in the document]
zichanglai: for non-persisted BitmapBlkAllocator::alloc_on_disk(), it do not need to do anything, so yes, it make sense to always return success.

src/lib/device/virtual_dev.hpp
```diff
@@ -149,6 +149,7 @@ class VirtualDev {
     /// during recovery where alloc_blk is called but before it was checkpointed, it crashed and we are trying to
     /// recover Please note that even calling this method is not guaranteed to persisted until checkpoint is taken.
     /// @param blkid BlkId to commit explicitly.
+    /// @param recovering : Is this commit happening during recovery
```
yamingk, line 152: this can be removed

src/tests/test_meta_blk_mgr.cpp
```diff
@@ -776,6 +781,31 @@ TEST_F(VMetaBlkMgrTest, random_dependency_test) {
     this->shutdown();
 }
 
+TEST_F(VMetaBlkMgrTest, recovery_test) {
```
yamingk, line 784: Can we add this test to conan build test by default, e.g. add this test to the googlt test filter?

src/lib/blkalloc/varsize_blk_allocator.cpp
```diff
@@ -626,6 +626,24 @@ blk_count_t VarsizeBlkAllocator::alloc_blks_direct(blk_count_t nblks, blk_alloc_
     return (nblks - nblks_remain);
 }
 
+//since this function will only be called during HS recovery, we can safe to update the cache bitmap directly without
+//touching the slab caches.
+BlkAllocStatus VarsizeBlkAllocator::mark_blk_allocated(BlkId const& bid) {
+    BlkAllocPortion& portion = blknum_to_portion(bid.blk_num());
+    {
+        auto const start_blk_id = portion.get_portion_num() * get_blks_per_portion();
```
yamingk, line 634: Can we move line: 634 ~ 639 inside "ifndef NDEBUG" to speed up a little for release build since they are just doing some assert in debug mode?

src/lib/device/virtual_dev.cpp
```diff
@@ -148,6 +148,12 @@ bool VirtualDev::is_blk_alloced(BlkId const& blkid) const {
 BlkAllocStatus VirtualDev::commit_blk(BlkId const& blkid) {
     Chunk* chunk = m_dmgr.get_chunk_mutable(blkid.chunk_num());
     HS_LOG(DEBUG, device, "commit_blk: bid {}", blkid.to_string());
+    auto const recovering = homestore::hs()->is_initializing();
+    if (!recovering) {
+        HS_DBG_ASSERT(is_blk_alloced(blkid), "commiting blkid {} not allocated", blkid.to_string());
```
yamingk, line 153: nit: Can we put "(non-recovery mode)" at the end of the asserted message.

src/lib/blkalloc/varsize_blk_allocator.cpp
zichanglai: but on L636, we need to hold the  portion lock, right?

src/tests/test_meta_blk_mgr.cpp
zichanglai: yes, now I had add this unit tests to set of default test cases of HS metablk

src/lib/blkalloc/varsize_blk_allocator.cpp
yamingk: right, 636 should be out of the macro.

src/tests/test_meta_blk_mgr.cpp
yamingk: is this not committed to this PR yet? Sorry, I didn't find where this test case was made default.

yamingk (approved):

PR #239, opened 2023-11-30
FIX btree remove

shosseinimotlagh:
There exist multiple multiple problems in btree that this pr will address them. 

1. In varlen node type, some room has been reserved for `var_node_header` that should be deducted for `occupied_size() ` . This change is needed to taken account when the entries transfer size is getting calculated for moving entries from a node to its next adjacent one. 
2. Fixing the calculation for parent. For sake of simplicity, suppose that there is no excess node to remove during `merge_nodes()` and the number of new nodes is the same as old nodes. However, the last key of some of new nodes is different. In this case the new size of parent node must be checked even though the number of new nodes is the same. 
3. (Improvement): If an entry is large and cannot accommodate in the last node according to calculated balanced size, but it can be fitted to the last new node; let it be there instead of making a new sparse node.

### Testing:
The remove op is tested 100 times with node size of 512 bytes with over 100K entries and iterations.

### DIFF
--- a/src/include/homestore/btree/detail/btree_remove_impl.ipp
+++ b/src/include/homestore/btree/detail/btree_remove_impl.ipp
@@ -335,14 +335,23 @@ btree_status_t Btree< K, V >::merge_nodes(const BtreeNodePtr& parent_node, const
 
         auto& old_ith_node = old_nodes[src_cursor.ith_node];
         auto const nentries = new_node->copy_by_size(m_bt_cfg, *old_ith_node, src_cursor.nth_entry, available_size);
+        total_size -= new_node->occupied_size();
         if (old_ith_node->total_entries() == (src_cursor.nth_entry + nentries)) {
             // Copied entire node
             ++src_cursor.ith_node;
             src_cursor.nth_entry = 0;
             available_size = balanced_size - new_node->occupied_size();
         } else {
-            src_cursor.nth_entry += nentries;
-            available_size = 0;
+            //  If it is the last node supposed to be, check if the remaining entries can be copied and not creating a
+            //  new nodes. This will make the last new node a little skewed from balanced size due to large key/values but
+            //  avoid making extra new node.
+            if (new_nodes.size() == num_nodes - 1 && total_size < new_node->available_size()) {
+                available_size = new_node->available_size();
+                src_cursor.nth_entry += nentries;
+            } else {
+                src_cursor.nth_entry += nentries;
+                available_size = 0;
+            }
         }
     }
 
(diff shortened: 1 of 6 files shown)


src/include/homestore/btree/detail/btree_remove_impl.ipp
```diff
@@ -335,14 +335,23 @@ btree_status_t Btree< K, V >::merge_nodes(const BtreeNodePtr& parent_node, const
...
             ++src_cursor.ith_node;
             src_cursor.nth_entry = 0;
             available_size = balanced_size - new_node->occupied_size();
         } else {
-            src_cursor.nth_entry += nentries;
-            available_size = 0;
+            //  If it is the last node supposed to be, check if the remaining entries can be copied and not creating a
+            //  new nodes. This will make the last new node a little skewed from balanced size due to large key/values but
+            //  avoid making extra new node.
+            if (new_nodes.size() == num_nodes - 1 && total_size < new_node->available_size()) {
```
shosseinimotlagh, line 348: Explained improvement

src/include/homestore/btree/detail/btree_remove_impl.ipp
```diff
@@ -375,14 +385,13 @@ btree_status_t Btree< K, V >::merge_nodes(const BtreeNodePtr& parent_node, const
...
-#if 0
+#endif
         // we first calculate the least amount of space being released after removing excess children. the key size
         // cannot be taken account; so we know for sure that value (i.e., linkinfo) and also its record will be freed.
         // If the end_idx is the parent's edge, the space is not released eventually.
         auto excess_releasing_nodes =
             old_nodes.size() - new_nodes.size() - (parent_node->total_entries() == end_idx) ? 1 : 0;
-        auto minimum_releasing_excess_size =
-            excess_releasing_nodes * (BtreeLinkInfo::get_fixed_size() + parent_node->get_record_size());
+        auto minimum_releasing_excess_size = excess_releasing_nodes * (BtreeLinkInfo::get_fixed_size());
```
shosseinimotlagh, line 394: this part is enabled for the problem item 2

src/include/homestore/btree/detail/varlen_node.hpp
```diff
@@ -71,6 +71,10 @@ class VariableNode : public VariantNode< K, V > {
 
     virtual ~VariableNode() = default;
 
+    uint32_t occupied_size() const override {
```
shosseinimotlagh, line 74: this part is for solving the problem item 1.

JacksonYao287 (commented):
i

JacksonYao287:
change version to `4.9.3` in conanfile.py

JacksonYao287 (approved):

raakella1 (approved):
LG!

PR #246, opened 2023-12-06
disable copy construction for homestore superblock

zichanglai:
with copy construction, superblock will be easy to cause duplicated metablk when call write() function if it is not carefully enough. we can avoid this by disable copy construction and only allow move construction.

### DIFF
--- a/src/include/homestore/superblk_handler.hpp
+++ b/src/include/homestore/superblk_handler.hpp
@@ -32,6 +32,30 @@ public:
 
     superblk(const std::string& meta_name = "") { set_name(meta_name); }
 
+    superblk(const superblk&) = delete;
+    superblk& operator=(const superblk&) = delete;
+
+    superblk(superblk&& rhs) noexcept
+         : m_meta_mgr_cookie(rhs.m_meta_mgr_cookie)
+           , m_raw_buf(std::move(rhs.m_raw_buf))
+           , m_sb(rhs.m_sb)
+           , m_metablk_name(std::move(rhs.m_metablk_name)) {
+	rhs.m_meta_mgr_cookie = nullptr;
+	rhs.m_sb = nullptr;
+    }
+
+    superblk& operator=(superblk&& rhs) noexcept {
+        if (this != &rhs) {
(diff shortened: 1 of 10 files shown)


zichanglai:
> Can you bump homstore patch version assuming it compiles successfully with homeobj?

yes, I can bump homestore patch version for this.  but homeobject need make little changes to build successfully because indextable interface is changed.

src/include/homestore/superblk_handler.hpp
```diff
@@ -32,6 +32,30 @@ class superblk {
...
+    superblk(const superblk&) = delete;
+    superblk& operator=(const superblk&) = delete;
+
+    superblk(superblk&& rhs) noexcept
+         : m_meta_mgr_cookie(rhs.m_meta_mgr_cookie)
+           , m_raw_buf(std::move(rhs.m_raw_buf))
+           , m_sb(rhs.m_sb)
+           , m_metablk_name(std::move(rhs.m_metablk_name)) {
+	rhs.m_meta_mgr_cookie = nullptr;
+	rhs.m_sb = nullptr;
```
JacksonYao287, line 44: why this two members is initialized in initialization list and set to nullptr here? 
they are already initialized with nullptr

src/include/homestore/superblk_handler.hpp
```diff
@@ -32,6 +32,30 @@ class superblk {
...
+    }
+
+    superblk& operator=(superblk&& rhs) noexcept {
+        if (this != &rhs) {
+            m_meta_mgr_cookie = rhs.m_meta_mgr_cookie;
+            m_raw_buf = std::move(rhs.m_raw_buf);
+            m_sb = rhs.m_sb;
+            m_metablk_name = std::move(rhs.m_metablk_name);
+            rhs.m_meta_mgr_cookie = nullptr;
+            rhs.m_sb = nullptr;
```
JacksonYao287, line 54: the same as above

src/include/homestore/superblk_handler.hpp
zichanglai: this is move construction

yamingk (approved):

PR #250, opened 2023-12-13
Add flexible journal vdev.

sanebay:
Ability to add chunks to a journal vdev dynamically. Create pool of chunks and maintain a list of chunks sliding window. Add api to create chunk. Chunks part of journal vdev are connected like singly linked list with next_chunk in private data. Chunks part of journal vdv have private data. Journal vdev can give independent descriptors to the user to do alloc write and truncate. Modified tests to support this.

Changes made.
1. Earlier only 2 journalvdev were there (data and control) and took fixed size and numchunks. Now user can only provide chunk size and we add chunks dynamically once vdev is full. Most of the state of the list of chunks and its offsets were maintained in the JournalVDev class now its moved to its inner class JournalVDev::Descriptor. Now a single JournalVDev can create multiple descriptors. Each descriptor now maintains the list of chunks, its offsets etc. User call open to create different descriptors.
2. Earlier the offsets of the sliding window could wrap around due to the fixed number of chunks. But now they only increase. Both the left and right side of the window increase. Truncate will increase the left and append blk will increase the right offset.
3. Add a chunk pool in journal vdev so that its faster to append a new chunk to the list.
4. Changes in device_manager is to support to start with zero chunks and api to create and remove single chunk.
5. Test code is almost same. Earlier the api were tested against a JournalVirtualDev but now its against a JournalVirtualDev::Descriptor. Added recovery test case.
6. JournalVDev ......................   Passed   38.81 sec

### DIFF
--- a/src/lib/common/homestore_config.fbs
+++ b/src/lib/common/homestore_config.fbs
@@ -140,6 +140,9 @@ table Generic {
     // percentage of cache used to create indx mempool. It should be more than 100 to 
     // take into account some floating buffers in writeback cache.
     indx_mempool_percent : uint32 = 110;
+
+    // Number of chunks in journal chunk pool.
+    journal_chunk_pool_capacity: uint32 = 5;
 }
 
 table ResourceLimits {
--- a/src/lib/device/device.h
+++ b/src/lib/device/device.h
@@ -42,15 +42,17 @@ struct vdev_info {
     uint32_t num_mirrors{0};                   // 12: Total number of mirrors
     uint32_t blk_size{0};                      // 16: IO block size for this vdev
     uint32_t num_primary_chunks{0};            // 20: number of primary chunks
-    uint8_t slot_allocated{0};                 // 24: Is this current slot allocated
-    uint8_t failed{0};                         // 25: set to true if disk is replaced
-    uint8_t hs_dev_type{0};                    // 26: PDev dev type (as in fast or data)
-    uint8_t multi_pdev_choice{0};              // 27: Choice when multiple pdevs are present (vdev_multi_pdev_opts_t)
-    char name[64];                             // 28: Name of the vdev
-    uint16_t checksum{0};                      // 92: Checksum of this entire Block
-    uint8_t alloc_type;                        // 94: Allocator type of this vdev
-    uint8_t chunk_sel_type;                    // 95: Chunk Selector type of this vdev_id
-    uint8_t padding[160]{};                    // 96: Pad to make it 256 bytes total
+    uint32_t chunk_size{0};                    // 24: chunk size used in vdev.
+    vdev_size_type_t size_type{};              // 28: Whether its a static or dynamic type.
+    uint8_t slot_allocated{0};                 // 29: Is this current slot allocated
+    uint8_t failed{0};                         // 30: set to true if disk is replaced
+    uint8_t hs_dev_type{0};                    // 31: PDev dev type (as in fast or data)
+    uint8_t multi_pdev_choice{0};              // 32: Choice when multiple pdevs are present (vdev_multi_pdev_opts_t)
+    char name[64];                             // 33: Name of the vdev
+    uint16_t checksum{0};                      // 97: Checksum of this entire Block
+    uint8_t alloc_type;                        // 98: Allocator type of this vdev
+    uint8_t chunk_sel_type;                    // 99: Chunk Selector type of this vdev_id
+    uint8_t padding[155]{};                    // 100: Pad to make it 256 bytes total
     uint8_t user_private[user_private_size]{}; // 128: User specific information
 
     uint32_t get_vdev_id() const { return vdev_id; }
@@ -94,11 +96,13 @@ ENUM(chunk_selector_t, uint8_t, // What are the options to select chunk to alloc
 
 struct vdev_parameters {
     std::string vdev_name;                  // Name of the vdev
+    vdev_size_type_t size_type{};           // Wether size is static or dynamic.
     uint64_t vdev_size;                     // Current Vdev size.
-    uint32_t num_chunks;                    // Total number of primary chunks.
+    uint32_t num_chunks{};                  // Total number of primary chunks.
                                             // NOTE: If pdev opts is ALL_PDEV_STRIPED, then num_chunks would round off
                                             // to number of pdevs evenly
     uint32_t blk_size;                      // Block size vdev operates on
+    uint32_t chunk_size{};                  // Chunk size provided for dynamic vdev.
     HSDevType dev_type;                     // Which physical device type this vdev belongs to (FAST or DATA)
     blk_allocator_type_t alloc_type;        // which allocator type this vdev wants to be with;
     chunk_selector_type_t chunk_sel_type;   // which chunk selector type this vdev wants to be with;
@@ -154,21 +158,28 @@ public:
     shared< VirtualDev > create_vdev(vdev_parameters&& vdev_param);
 
     const Chunk* get_chunk(uint32_t chunk_id) const {
+        std::unique_lock lg{m_vdev_mutex};
         return (chunk_id == INVALID_CHUNK_ID) ? nullptr : m_chunks[chunk_id].get();
     }
 
     Chunk* get_chunk_mutable(uint32_t chunk_id) {
+        std::unique_lock lg{m_vdev_mutex};
         return (chunk_id == INVALID_CHUNK_ID) ? nullptr : m_chunks[chunk_id].get();
     }
 
     uint32_t atomic_page_size(HSDevType dtype) const;
     uint32_t optimal_page_size(HSDevType dtype) const;
+    uint32_t align_size(HSDevType dtype) const;
 
     std::vector< PhysicalDev* > get_pdevs_by_dev_type(HSDevType dtype) const;
     std::vector< shared< VirtualDev > > get_vdevs() const;
 
     uint64_t total_capacity() const;
     uint64_t total_capacity(HSDevType dtype) const;
+
+    shared< Chunk > create_chunk(HSDevType dev_type, uint32_t vdev_id, uint64_t chunk_size, const sisl::blob& data);
+    void remove_chunk(shared< Chunk > chunk);
+    void remove_chunk_locked(shared< Chunk > chunk);
 
 private:
     void load_vdevs();
@@ -180,5 +191,53 @@ private:
 
     const std::vector< PhysicalDev* >& pdevs_by_type_internal(HSDevType dtype) const;
 }; // class DeviceManager
+
+// Chunk pool is used to get chunks when there is no space
+// and its cheaper compared to create a chunk on the fly.
+// Creating chunk on the fly causes sync write.
+class ChunkPool {
+public:
+    struct Params {
+        uint64_t pool_capacity;
+        // Private data used when creating chunks.
+        std::function< sisl::blob() > init_private_data_cb;
+        uint8_t hs_dev_type;
+        uint32_t vdev_id;
+        uint64_t chunk_size;
+    };
+
+    ChunkPool(DeviceManager& dmgr, Params&& param);
+    ~ChunkPool();
+
+    // Start the chunk pool.
+    void start();
+
+    // Add chunk to the pool. If the queue is full,
+    // chunk removed from the system. Returns if
+    // if we could reuse chunk by adding back to pool.
+    bool enqueue(shared< Chunk >& chunk);
+
+    // Get a chunk from the pool.
+    shared< Chunk > dequeue();
+
+    // Returns the capacity of the chunk pool.
+    uint64_t capacity() { return m_params.pool_capacity; }
+    uint64_t size() { return m_pool.size(); }
+
+private:
+    // Producer thread.
+    void producer();
+
+private:
+    DeviceManager& m_dmgr;
+    Params m_params;
+    std::list< shared< Chunk > > m_pool;
+    uint32_t m_pool_capacity;
+    std::condition_variable m_pool_cv;
+    std::mutex m_pool_mutex;
+    std::thread m_producer_thread;
+    bool m_run_pool{false};
+    folly::Promise< folly::Unit > m_pool_halt;
+};
 
 } // namespace homestore
--- a/src/lib/device/journal_vdev.cpp
+++ b/src/lib/device/journal_vdev.cpp
@@ -32,115 +32,247 @@
 #include "common/homestore_utils.hpp"
 #include "common/resource_mgr.hpp"
 
+SISL_LOGGING_DECL(journalvdev)
+
 namespace homestore {
 JournalVirtualDev::JournalVirtualDev(DeviceManager& dmgr, const vdev_info& vinfo, vdev_event_cb_t event_cb) :
-        VirtualDev{dmgr, vinfo, std::move(event_cb), false /* is_auto_recovery */} {}
+        VirtualDev{dmgr, vinfo, std::move(event_cb), false /* is_auto_recovery */} {
 
-off_t JournalVirtualDev::alloc_next_append_blk(size_t sz) {
-    if (used_size() + sz > size()) {
-        // not enough space left;
-        HS_LOG(ERROR, device, "No space left! m_write_sz_in_total: {}, m_reserved_sz: {}", m_write_sz_in_total.load(),
-               m_reserved_sz);
-        return INVALID_OFFSET;
+    // Private data stored when chunks are created.
+    init_private_data = std::make_shared< JournalChunkPrivate >();
+    m_chunk_pool = std::make_unique< ChunkPool >(
+        dmgr,
+        ChunkPool::Params{
+            HS_DYNAMIC_CONFIG(generic.journal_chunk_pool_capacity),
+            [this]() {
+                init_private_data->created_at = get_time_since_epoch_ms();
+                sisl::blob private_blob{r_cast< uint8_t* >(init_private_data.get()), sizeof(JournalChunkPrivate)};
+                return private_blob;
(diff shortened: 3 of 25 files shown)


src/lib/common/homestore_config.fbs
```diff
@@ -140,6 +140,9 @@ table Generic {
     // percentage of cache used to create indx mempool. It should be more than 100 to 
     // take into account some floating buffers in writeback cache.
     indx_mempool_percent : uint32 = 110;
+
+    // journal chunk pool size
+    journal_chunk_pool_size: uint32 = 5;
```
hkadayam, line 145: can you please add unit as well, something like journal_chunk_pool_size_mb

src/lib/device/device.h
```diff
@@ -95,10 +96,11 @@ ENUM(chunk_selector_t, uint8_t, // What are the options to select chunk to alloc
 struct vdev_parameters {
     std::string vdev_name;                  // Name of the vdev
     uint64_t vdev_size;                     // Current Vdev size.
-    uint32_t num_chunks;                    // Total number of primary chunks.
+    uint32_t num_chunks{};                  // Total number of primary chunks.
                                             // NOTE: If pdev opts is ALL_PDEV_STRIPED, then num_chunks would round off
                                             // to number of pdevs evenly
     uint32_t blk_size;                      // Block size vdev operates on
+    uint32_t chunk_size{};                  // Chunk size provided for dynamic vdev.
```
hkadayam, line 103: If user has to define either chunk_size or num_chunks, should we put them as std::variant?

src/lib/device/virtual_dev.hpp
```diff
@@ -92,7 +92,7 @@ class VirtualDev {
 
     std::mutex m_mgmt_mutex;          // Any mutex taken for management operations (like adding/removing chunks).
     std::set< PhysicalDev* > m_pdevs; // PDevs this vdev is working on
-    sisl::sparse_vector< shared< Chunk > > m_all_chunks; // All chunks part of this vdev
+    std::vector< shared< Chunk > > m_all_chunks; // All chunks part of this vdev
```
hkadayam, line 95: Why is it replaced with std::vector<>? Chunks could be read in any order right and we wanted to place them in chunk_id order right?

src/lib/logstore/log_stream.cpp
```diff
@@ -22,13 +22,16 @@
...
 namespace homestore {
 SISL_LOGGING_DECL(logstore)
 
-log_stream_reader::log_stream_reader(off_t device_cursor, JournalVirtualDev* store, uint64_t read_size_multiple) :
-        m_vdev{store}, m_first_group_cursor{device_cursor}, m_read_size_multiple{read_size_multiple} {
-    m_vdev->lseek(m_first_group_cursor);
+log_stream_reader::log_stream_reader(off_t device_cursor, JournalVirtualDev* vdev,
+                                     shared< JournalVirtualDev::Descriptor > vdev_jd, uint64_t read_size_multiple) :
+        m_vdev{vdev},
+        m_vdev_jd{vdev_jd},
```
hkadayam, line 31: nit: Can do ```m_vdev_js{std::move(vdev_jd)}```

src/lib/common/homestore_config.fbs
sanebay: @hari, add num_

src/lib/device/journal_vdev.hpp
```diff
@@ -271,66 +397,29 @@ class JournalVirtualDev : public VirtualDev {
...
+    // Chunk pool is used to get chunks when there is no space
+    // and its cheaper compared to create a chunk on the fly.
+    // Creating chunk on the fly causes sync write. Getting a chunk
+    // from the pool causes a single write of the last chunk in the list
+    // to update its end_of_chunk and next_chunk.
+    std::list< shared< Chunk > > m_chunk_pool;
+    uint32_t m_chunk_pool_capacity;
+    std::condition_variable m_chunk_pool_cv;
+    std::mutex m_chunk_pool_mutex;
+    iomgr::io_fiber_t m_chunk_pool_fiber;
```
sanebay, line 420: hari: use normal std thread

src/lib/device/journal_vdev.cpp
```diff
@@ -34,75 +34,300 @@
...
+        uint64_t created_at{};
+    };
+
+    std::unordered_map< logdev_id_t, HeadChunk > logdev_head_map;
+    std::unordered_map< chunk_num_t, shared< Chunk > > chunk_map;
+    std::unordered_set< chunk_num_t > visited_chunks;
+
+    // Traverse the chunks and find the heads of the logdev_id's.
+    for (auto& chunk : m_all_chunks) {
+        auto* data = r_cast< JournalChunkPrivate* >(const_cast< uint8_t* >(chunk->user_private()));
```
sanebay, line 64: descriptor_chunk_info

src/lib/device/journal_vdev.cpp
```diff
@@ -34,75 +34,300 @@
...
+    m_chunk_pool_cv.notify_one();
+    return chunk;
+}
+
+void JournalVirtualDev::release_chunk_to_pool(const shared< Chunk >& chunk) {
+    RELEASE_ASSERT(chunk, "Chunk invalid");
+    bool reuse = false;
+    {
+        std::unique_lock< std::mutex > lk{m_chunk_pool_mutex};
+        if (m_chunk_pool.size() < m_chunk_pool_capacity) {
```
sanebay, line 209: hari: clear the private data before putting back to pool

src/lib/device/journal_vdev.cpp
```diff
@@ -34,75 +34,300 @@
...
+            m_chunk_pool.push_back(chunk);
+            HS_LOG(TRACE, device, "Produced chunk to pool id {}", chunk->chunk_id());
+        }
+    });
+}
+
+shared< Chunk > JournalVirtualDev::get_chunk_from_pool() {
+    shared< Chunk > chunk;
+    {
+        std::unique_lock< std::mutex > lk{m_chunk_pool_mutex};
```
sanebay, line 194: wait if the queue is empty.

src/lib/common/homestore_config.fbs
sanebay: Ack

src/lib/device/virtual_dev.hpp
sanebay: Reverted.

src/lib/logstore/log_stream.cpp
sanebay: Ack

src/lib/device/journal_vdev.hpp
sanebay: Moving the chunk_pool to chunk.cpp

src/lib/device/device.h
sanebay: As discussed added vdev_size_type_t enum {STATIC, DYNAMIC} in format param. If it static it takes size and num_chunks or chunk_size. If its dynamic,  it needs only chunk_size.

src/lib/device/device.h
```diff
@@ -181,4 +189,52 @@ class DeviceManager {
...
+    // Returns the capacity of the chunk pool.
+    uint64_t capacity() { return m_params.pool_capacity; }
+    uint64_t size() { return m_pool.size(); }
+
+private:
+    // Producer thread.
+    void producer();
+
+private:
+    // Chunk pool is used to get chunks when there is no space
```
yamingk, line 226: move this comment to head of the class as some documentation of this class.

src/lib/device/journal_vdev.hpp
```diff
@@ -19,250 +19,367 @@
...
+    // right side (m_end_offset). Truncate increases the left m_data_start_offset
+    // and pop chunks from the front of the list. alloc_next_append_blk adds more chunk to
+    // the back of list if no space and adjusts the right m_end_offset. All offsets
+    // only increase and never wraps around. Each chunk in the descriptor has private
+    // data about the logdev_id its part of, next chunk to maintain the list.
+    struct Descriptor {
+    private:
+        JournalVirtualDev& m_vdev;
+        logdev_id_t m_logdev_id; // Unique id identifying the journal descriptor.
+        // off_t is long. make it uint64_t ?
```
yamingk, line 62: good point, long is only 4 bytes, and it is signed, it will overflow if exceeding 2TB offset, right?

src/lib/device/journal_vdev.cpp
```diff
@@ -34,75 +34,211 @@
...
+    RELEASE_ASSERT_LT(sz, chunk_size, "Size requested greater than chunk size");
+
+    if ((tail_offset() + static_cast< off_t >(sz)) >= m_end_offset) {
+        // not enough space left, add a new chunk.
+        HS_LOG(DEBUG, device,
+               "No space left, desc: {} m_write_sz_in_total: {}, m_reserved_sz: {} tail: 0x{}. Creating chunk",
+               m_logdev_id, m_write_sz_in_total.load(), m_reserved_sz, to_hex(tail_offset()));
+
+#ifdef _PRERELEASE
+        iomgr_flip::test_and_abort("abort_before_update_eof_cur_chunk");
```
yamingk, line 212: Are these flips enabled and running fine somewhere?

src/lib/device/journal_vdev.hpp
```diff
@@ -19,250 +19,367 @@
...
+using logdev_id_t = uint64_t;
+
+// Chunks used for journal vdev has journal related info stored in chunk private data.
+// Each log device has a list of journal chunk data with next_chunk.
+// Journal vdev will arrange the chunks in order during recovery.
+struct JournalChunkPrivate {
+    logdev_id_t logdev_id{0};
+    bool is_head{false};           // Is it the head element.
+    uint64_t created_at{0};        // Creation timestamp
+    uint64_t end_of_chunk_size{0}; // The offset indicates end of chunk.
```
yamingk, line 42: nit: I know the this is migrated from old, can we rename it to 'end_offset', since it is actually a offset not a size.

src/lib/device/virtual_dev.cpp
```diff
@@ -123,6 +123,14 @@ void VirtualDev::add_chunk(cshared< Chunk >& chunk, bool is_fresh_chunk) {
     m_chunk_selector->add_chunk(chunk);
 }
 
+void VirtualDev::remove_chunk(cshared< Chunk >& chunk) {
+    std::unique_lock lg{m_mgmt_mutex};
+    auto iter = std::remove_if(m_all_chunks.begin(), m_all_chunks.end(), [chunk](auto c) { return c == chunk; });
+    m_all_chunks.erase(iter, m_all_chunks.end());
+
+    // m_chunk_selector->remove_chunk(chunk); // TODO
```
yamingk, line 131: won't there be problem if it is not removed from heap_chunk_selector, is there a reason it can't be done within this PR?

src/lib/device/journal_vdev.cpp
```diff
@@ -34,75 +34,211 @@
...
+        return journal_desc;
+    }
+
+    HS_LOG(TRACE, device, "Opened log device descriptor {}", logdev_id);
+    return it->second;
+}
+
+void JournalVirtualDev::Descriptor::append_chunk() {
+    // Get a new chunk from the pool.
+    auto new_chunk = m_vdev.m_chunk_pool->dequeue();
```
yamingk, line 160: will there be a case when m_chunk_pool is empty and new_chunk is null?

src/lib/device/journal_vdev.hpp
```diff
@@ -19,250 +19,367 @@
...
+        off_t dev_offset(off_t nbytes) const;
+
+        /**
+         * @brief : get the start logical offset where data starts;
+         *
+         * @return : the start logical offset where data starts;
+         */
+        off_t data_start_offset() const { return m_data_start_offset; }
+
+        off_t end_offset() const { return m_end_offset; }
```
yamingk, line 228: can you also provide comment of this API, especially the difference between tail_offset, so that who are not familiar with this code can be easier to understand.

src/lib/device/journal_vdev.cpp
```diff
@@ -34,75 +34,211 @@
...
-    if (end_offset < ds_off) { HS_DBG_ASSERT_EQ(size() - used_size(), static_cast< uint64_t >(ds_off - end_offset)); }
-#endif
-    // works for both "end_offset >= ds_off" and "end_offset < ds_off";
-    if (offset_in_chunk + sz <= chunk->size()) {
-        // not acrossing boundary, nothing to do;
-    } else if ((used_size() + (chunk->size() - offset_in_chunk) + sz) <= size()) {
-        // across chunk boundary, still enough space;
+        // Clear the private chunk data.
+        *data = JournalChunkPrivate{};
+        update_chunk_private(chunk, data);
```
yamingk, line 118: shouldn't we set the end_offset in private data here, e.g. like 69 was doing?

src/lib/device/journal_vdev.cpp
```diff
@@ -34,75 +34,211 @@
...
+            m_chunk_pool->enqueue(chunk);
+            LOGWARN("Reuse to pool orphan chunk {} found for logdev {} next {}.", chunk_id, logdev_id, next_chunk);
+        } else {
+            LOGWARN("Removing orphan chunk {} found for logdev {} next {}.", chunk_id, logdev_id, next_chunk);
+            m_dmgr.remove_chunk(chunk);
+        }
+    }
 
-        // Update the overhead to total write size;
-        m_write_sz_in_total.fetch_add(chunk->size() - offset_in_chunk, std::memory_order_relaxed);
```
yamingk, line 66: this line should be kept, no?

src/lib/device/journal_vdev.cpp
```diff
@@ -34,75 +34,211 @@
...
+        auto last_chunk = m_journal_chunks.back();
+        auto* last_chunk_private = r_cast< JournalChunkPrivate* >(const_cast< uint8_t* >(last_chunk->user_private()));
+
+        // Set the next chunk with the newly created chunk id.
+        last_chunk_private->next_chunk = new_chunk->chunk_id();
+
+        // Append the new chunk
+        m_journal_chunks.push_back(new_chunk);
+        auto chunk_size = m_vdev.info().chunk_size;
+        auto offset_in_chunk = (tail_offset() % chunk_size);
```
yamingk, line 179: suggest to call offset_to_chunk to not assume chunk is always same size.

src/lib/device/virtual_dev.cpp
sanebay: Added remove_chunk() default function. HO can override it.

src/lib/device/journal_vdev.cpp
sanebay: We need to add it in our long running tests

src/lib/device/journal_vdev.hpp
sanebay: Made it end_of_chunk

src/lib/device/journal_vdev.hpp
sanebay: Currently off_t is mapped to int64_t which itself is huge.

src/lib/device/journal_vdev.hpp
sanebay: Made it end_of_chunk

src/lib/device/journal_vdev.cpp
sanebay: Added assert.

src/lib/device/journal_vdev.hpp
sanebay: Its added near the class variables. journal_vdev.hpp:50. If its not clear, please let me know.

src/lib/device/journal_vdev.cpp
sanebay: As discussed all chunks in a vdev have same size.

src/lib/device/journal_vdev.cpp
sanebay: This is now done in append_chunk() function line 183

src/tests/test_journal_vdev.cpp
```diff
@@ -315,20 +360,105 @@ class VDevIOTest : public ::testing::Test {
...
-    uint64_t m_total_size = 0;
     std::map< off_t, write_info > m_off_to_info_map;
     Clock::time_point m_start_time;
-    std::shared_ptr< JournalVirtualDev > m_vdev;
+    std::shared_ptr< JournalVirtualDev::Descriptor > m_vdev_jd;
+    friend class VDevJournalIOTest;
 };
 
-TEST_F(VDevIOTest, VDevIOTest) { this->execute(); }
+// TODO add more tests covering unclean shutdown and more corner cases.
```
yamingk, line 375: Is there a issue created for this?

src/tests/test_journal_vdev.cpp
```diff
@@ -315,20 +360,105 @@ class VDevIOTest : public ::testing::Test {
...
+
+TEST_F(VDevJournalIOTest, LongRunning) {
```
yamingk, line 377: Can we also specify how long has it been run in the PR description, would be helpful for reviewers?

It should be good to also add this long running test case to existing logstore long running, right?

src/tests/test_journal_vdev.cpp
sanebay: New issue created. Will add that as future task https://github.com/eBay/HomeStore/issues/283

src/tests/test_journal_vdev.cpp
sanebay: Currently its running just 40secs. Yes we should add it to long running.

yamingk (approved):

PR #251, opened 2023-12-14
fix dirty buf uninit issue

yamingk:
The bug is without initialization of the dirty buffer, it is left with garbage data. 
when cp flushes, it could mistakenly treating it is dirty and flushing garbage offset into metablk.


zichanglai (approved):
LGTM

PR #256, opened 2023-12-19
issue: 188 long duration test for meta svc and logstore svc

yamingk:
1. add long duration py for meta svc and longstore svc.
2. bug fix when failure to allocate the blks, 

- avoid crash but return failure to caller. 
- avoid keep looping when fail to allocate blks. 

Test:
====
Run for a while with the test locally and it is running fine. 
Target is to run them in real machine for long hours and fix any problem if there is any there.


shosseinimotlagh (commented):
Add binaries and script file to docker file.

yamingk:
> Add binaries and script file to docker file.

Binaries are already there, there is no new binary.

shosseinimotlagh (commented):
Add files to Jenkins that Dockerfile can use them to copy in the image. 
Make sure script file has correct permission

yamingk:
> Add files to Jenkins that Dockerfile can use them to copy in the image. Make sure script file has correct permission
If it is copied to .jenkins_nightly, they won't be run by default, right? 
The thing is some of the test might not be run in jenkins_nightly but should be run with real drives.

src/tests/test_scripts/log_meta_test.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env python3
+## @file vol_test.py
```
shosseinimotlagh, line 2: nit: change or remove it

src/tests/test_scripts/log_meta_test.py
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env python3
```
shosseinimotlagh, line 1: Please make sure the permission of this script file is executable.

shosseinimotlagh (approved):
LGTM. It is a good practice to make sure the permission of the python script file is executable.

PR #271, opened 2024-01-08
SDSTOR-12187 :  PartitionUtilization Endpoint

shosseinimotlagh:
Expose PartitionUtilization as volume used size divided by AM total size

### Testing: 
in one terminal, run 
`bin/test_volume --gtest_filter=VolTest.init_io_test --remove_file_on_shutdown=1 --delete_volume=1 --run_time=10000 --http_port=5000 --max_volume=1  --p_volume_size=100 --max_disk_capacity=1 `

in another terminal, run the below command to get the volume uuid:
`nublox-tools/AM/hs_http_cli_tool.py  get_object --name test_files/vol0`

run the below API command:
`curl http://localhost:5000/api/v1/utilization/[volumeUUID]`

### DIFF
--- a/src/api/vol_interface.hpp
+++ b/src/api/vol_interface.hpp
@@ -316,6 +316,7 @@ public:
 
     virtual const char* get_name(const VolumePtr& vol) = 0;
     virtual uint64_t get_size(const VolumePtr& vol) = 0;
+    virtual uint64_t get_used_size(const VolumePtr& vol) = 0;
     virtual uint64_t get_page_size(const VolumePtr& vol) = 0;
     virtual boost::uuids::uuid get_uuid(std::shared_ptr< Volume > vol) = 0;
     virtual sisl::blob at_offset(const boost::intrusive_ptr< BlkBuffer >& buf, uint32_t offset) = 0;
--- a/src/homeblks/home_blks.cpp
+++ b/src/homeblks/home_blks.cpp
@@ -634,6 +634,11 @@ bool HomeBlks::verify_index_bm() {
     return true;
 }
 
+uint64_t HomeBlks::get_used_size(const VolumePtr& vol) {
+    /* Update per volume status */
+   return  vol->get_used_size().used_total_size;
+}
+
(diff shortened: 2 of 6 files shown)


src/homeblks/home_blks.cpp
```diff
@@ -634,6 +634,12 @@ bool HomeBlks::verify_index_bm() {
     return true;
 }
 
+uint64_t HomeBlks::get_used_volume(const boost::uuids::uuid& uuid) {
+    /* Update per volume status */
+   VolumePtr vol = lookup_volume(uuid);
```
yamingk, line 639: you can just assert here I think. return 0 can either mean there is 0 used size.

src/api/vol_interface.hpp
```diff
@@ -316,6 +316,7 @@ class VolInterface {
 
     virtual const char* get_name(const VolumePtr& vol) = 0;
     virtual uint64_t get_size(const VolumePtr& vol) = 0;
+    virtual uint64_t get_used_volume(const boost::uuids::uuid& uuid) = 0;
```
yamingk, line 319: get_used_size

src/homeblks/home_blks.cpp
shosseinimotlagh: if there is no volume found, `HomeBlksHttpServer::get_utilization()` return an error message  `vol not found!`. This is for safety (an edge case)that if between calling HomeBlksHttpServer::get_utilization Line 165 and 172, the volume does not exist in lookup map.

src/api/vol_interface.hpp
shosseinimotlagh: ok

src/homeblks/home_blks.cpp
yamingk: if so, can we pass the volptr to homeblks directly since you've already looked it up, and no need to worry about lookup again and check the result.

shosseinimotlagh: done

yamingk (approved):

raakella1 (approved):
LG!

PR #272, opened 2024-01-09
Fix shutdown sequence with cp mgr.

sanebay:
Add blocking cp flush in cp mgr shutdown. Added atomic to stop cp to trigger another cp.
Destroy in btree and index dont need cp context. As cp mgr already shutdown.
Shut the cp mgr first.
Test meta blk used directly calling meta service api's to simulate recovery, instead use homestore restart to simulate a recovery.


src/lib/checkpoint/cp_mgr.cpp
```diff
@@ -74,10 +74,17 @@ void CPManager::shutdown() {
     LOGINFO("Stopping cp timer");
     iomanager.cancel_timer(m_cp_timer_hdl, true);
     m_cp_timer_hdl = iomgr::null_timer_handle;
+    m_shutdown_initiated = true;
```
hkadayam, line 77: Shouldn't the shutdown initiated be done in Homestore::shutdown()?

src/include/homestore/checkpoint/cp_mgr.hpp
```diff
@@ -154,6 +154,7 @@ class CPManager {
     superblk< cp_mgr_super_block > m_sb;
     std::vector< iomgr::io_fiber_t > m_cp_io_fibers;
     iomgr::timer_handle_t m_cp_timer_hdl;
+    std::atomic< bool > m_shutdown_initiated{false};
```
yamingk, line 157: Probably moving this to homestore instance could be more reasonable?
And set it from homestore::shtudown?

src/lib/checkpoint/cp_mgr.cpp
sanebay: Actually its for the cp_mgr destructor to notify the cp_flush to not initiate back to back CP.

src/lib/checkpoint/cp_mgr.cpp
```diff
@@ -74,10 +74,17 @@ void CPManager::shutdown() {
...
 
-    auto cp = get_cur_cp();
-    delete (cp);
+    LOGINFO("Trigger cp flush");
+    auto fut = trigger_cp_flush(true /* force */);
+    auto success = std::move(fut).get();
```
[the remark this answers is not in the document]
sanebay: Ack.

src/include/homestore/checkpoint/cp_mgr.hpp
sanebay: Actually its for the cp_mgr destructor to notify the cp_flush to not initiate back to back CP.

yamingk: I understand that, but cp_flush can also get homestore inst to get shutdown initiated, right?

sanebay: As discussed renamed the cp_shutdown.

src/lib/checkpoint/cp_mgr.cpp
sanebay: Renamed the shutdown to cp_shutdown as its notify the cp flush that cp mgr is shutting down.

yamingk (dismissed):

hkadayam (approved):

PR #280, opened 2024-01-19
Add grpc stop for flip service in homestore.

sanebay:


src/tests/test_index_btree.cpp
```diff
@@ -121,7 +120,7 @@ struct BtreeTest : public BtreeTestHelper< TestType >, public ::testing::Test {
     }
 };
 
-using BtreeTypes = testing::Types< FixedLenBtree, VarKeySizeBtree, VarValueSizeBtree, VarObjSizeBtree >;
+using BtreeTypes = testing::Types< FixedLenBtree /* , VarKeySizeBtree, VarValueSizeBtree, VarObjSizeBtree */ >;
```
yamingk, line 123: Assume this expected to comment out these test cases for now right?

yamingk (approved):

src/tests/test_index_btree.cpp
sanebay: Yes temporarily disabled it.

PR #281, opened 2024-01-19
Remove logstore family

sanebay:
1. Earlier logstore services have two fixed logstore families, 1 logdev under each family and family managing multiple logstores . This PR removes the logstore families indirection. Now we remove the logstore family class and let logstore service handles multiple logdev's and logdevs handle multiple logstores under it. Clients can now create multiple log devices each having its own state. This PR depends on the dynamic journal vdev PR. Earlier there were fixed number of metablks. Now we can have more metablks, one per logdev.

```
Earlier
       LogStoreService 
             LogStoreFamily(CTRL and DATA)
                          LogStores  -> 1 LogDev
Now
        LogStoreService 
             LogDev(1... N)
                      LogStores(1..N). multiple logstores under 1 LogDev.
```

2. Removed the LOG_REPLICATED and LOG_LOCAL and make it single LOG service.
3. Move functionality of log store family of handling log stores to logdev class.
4. We have a single journal vdev. All logdevices uses journal descriptors to talk to a single journal vdev. Each descriptor has its own list of chunks and state of the offset from a single journal vdev.
5. Added api's to create/open logdev in logservice.

IMG_5425


src/include/homestore/logstore/log_store_internal.hpp
```diff
@@ -44,12 +44,14 @@ typedef std::function< void(logstore_req*, logdev_key) > log_req_comp_cb_t;
 typedef sisl::byte_view log_buffer;
 typedef uint32_t logstore_id_t;
 typedef uint8_t logstore_family_id_t;
```
hkadayam, line 46: Can you please remove logstore_family_id_t?

src/lib/logstore/log_dev.cpp
```diff
@@ -580,7 +583,242 @@ void LogDev::rollback(logstore_id_t store_id, logid_range_t id_range) {
...
+
+        // Also call the logstore to inform that start/replay is completed.
+        if (!format) {
+            for (auto& p : m_id_logstore_map) {
+                auto& lstore{p.second.m_log_store};
+                if (lstore && lstore->get_log_replay_done_cb()) {
+                    lstore->get_log_replay_done_cb()(lstore, lstore->seq_num() - 1);
+                    lstore->truncate(lstore->truncated_upto());
+                }
+            }
```
hkadayam, line 617: I don't think this code needs to be here. This method is about handling unopened_ids and we are doing things which are opened ids as well here right? Can we move this block to the caller. Yes it needs to take the write lock twice, which is ok

src/lib/logstore/log_dev.cpp
```diff
@@ -580,7 +583,242 @@ void LogDev::rollback(logstore_id_t store_id, logid_range_t id_range) {
...
+    return lstore;
+}
+
+void LogDev::open_log_store(logstore_id_t store_id, bool append_mode, const log_store_opened_cb_t& on_open_cb) {
+    folly::SharedMutexWritePriority::WriteHolder holder(m_store_map_mtx);
+    const auto it = m_id_logstore_map.find(store_id);
+    HS_REL_ASSERT((it == m_id_logstore_map.end()), "store_id {}-{} already exists", m_logdev_id, store_id);
+    LOGINFO("Opening log store id {}-{}", m_logdev_id, store_id);
+    m_id_logstore_map.insert(std::make_pair<>(store_id, logstore_info_t{nullptr, on_open_cb, append_mode}));
+}
```
hkadayam, line 643: I think depending on which PR goes first, there will some clash with my PR on Restartability of ReplDev and this one, since we made the open_log_store to return a future.

src/lib/logstore/log_dev.cpp
```diff
@@ -35,21 +35,18 @@ namespace homestore {
...
-    if (f_id == LogStoreService::DATA_LOG_FAMILY_IDX) {
-        m_flush_size_multiple = HS_DYNAMIC_CONFIG(logstore->flush_size_multiple_data_logdev);
-    } else if (f_id == LogStoreService::CTRL_LOG_FAMILY_IDX) {
-        m_flush_size_multiple = HS_DYNAMIC_CONFIG(logstore->flush_size_multiple_ctrl_logdev);
-    }
+LogDev::LogDev(const logdev_id_t id) : m_logdev_id{id} {
+    m_flush_size_multiple = HS_DYNAMIC_CONFIG(logstore->flush_size_multiple_logdev);
+    register_store_found_cb(bind_this(LogDev::on_log_store_found, 2));
+    register_append_cb(bind_this(LogDev::on_io_completion, 5));
+    register_logfound_cb(bind_this(LogDev::on_logfound, 6));
```
hkadayam, line 49: I think all these 3 callbacks can be removed. This was required because LogStoreFamily was separate class and thus register to LogDev. Now that is merged, it can simply be a function call and code will look lot simpler to follow.

src/lib/logstore/log_dev.cpp
```diff
@@ -580,7 +583,242 @@ void LogDev::rollback(logstore_id_t store_id, logid_range_t id_range) {
...
+    auto& l_info = const_cast< logstore_info_t& >(it->second);
+    l_info.m_log_store =
+        std::make_shared< HomeLogStore >(shared_from_this(), store_id, l_info.append_mode, sb.m_first_seq_num);
+    if (l_info.m_on_log_store_opened) l_info.m_on_log_store_opened(l_info.m_log_store);
+}
+
+static thread_local std::vector< std::shared_ptr< HomeLogStore > > s_cur_flush_batch_stores;
+
+void LogDev::on_io_completion(logstore_id_t id, logdev_key ld_key, logdev_key flush_ld_key,
+                              uint32_t nremaining_in_batch, void* ctx) {
```
hkadayam, line 699: Suggestion: In fact this whole function could be moved under on_flush_completion().

src/lib/logstore/log_dev.cpp
```diff
@@ -580,7 +583,242 @@ void LogDev::rollback(logstore_id_t store_id, logid_range_t id_range) {
...
+    for (auto& store_ptr : m_non_participating_stores) {
+        store_ptr->post_device_truncation(min_safe_ld_key);
+    }
+    m_min_trunc_stores.clear(); // Not clearing here, would cause a shared_ptr ref holding.
+    m_non_participating_stores.clear();
+
+    return min_safe_ld_key;
+}
+
+void LogDev::get_logdev_status(const int verbosity, nlohmann::json& js) const {
```
hkadayam, line 821: Is there a reason to change name from get_status to get_logdev_status?

hkadayam: We can certainly merge the current get_status() method with this one.

src/include/homestore/logstore_service.hpp
```diff
@@ -135,35 +144,35 @@ class LogStoreService {
...
 
-    folly::Future< std::error_code > create_vdev(uint64_t size, logstore_family_id_t family, uint32_t chunk_size);
-    shared< VirtualDev > open_vdev(const vdev_info& vinfo, logstore_family_id_t family, bool load_existing);
-    shared< JournalVirtualDev > get_vdev(logstore_family_id_t family) const {
-        return (family == DATA_LOG_FAMILY_IDX) ? m_data_logdev_vdev : m_ctrl_logdev_vdev;
-    }
+    folly::Future< std::error_code > create_vdev(uint64_t size, uint32_t chunk_size);
+    shared< VirtualDev > open_vdev(const vdev_info& vinfo, bool load_existing);
+    shared< JournalVirtualDev > get_vdev() const { return m_logdev_vdev; }
+    std::vector< std::shared_ptr< LogDev > > get_all_logdev();
```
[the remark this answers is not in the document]
sanebay: homestore am seeing mixed , so keeping the same as other in same file.

src/lib/logstore/log_dev.cpp
sanebay: Yes rebased to latest.

src/lib/logstore/log_dev.cpp
sanebay: register_append_cb and register_append_cb are used by  test_log_dev to override . So kept the same.

src/lib/logstore/log_dev.cpp
sanebay: The same callback is overridden by test code test_log_dev

src/lib/logstore/log_dev.cpp
hkadayam: I think we should adjust this - It was inevitable earlier, now it is avoidable, so I feel doing a callback which only calls its local function is too confusing. We have the oppurtunity to simplify. So if you feel strongly of not doing it now, may be in next PR we can take care? I am not sure why test case really cares about internal callback. Am I missing something?

sanebay: Removed the callbacks.

src/include/homestore/logstore_service.hpp
```diff
@@ -131,38 +142,37 @@ class LogStoreService {
      * Default to false
      * @param dry_run: If the truncate is a real one or just dry run to simulate the truncation
      */
-    void device_truncate(const device_truncate_cb_t& cb = nullptr, const bool wait_till_done = false,
-                         const bool dry_run = false);
+    void device_truncate(const device_truncate_cb_t& cb = nullptr, bool wait_till_done = false, bool dry_run = false);
```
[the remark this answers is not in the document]
sanebay: Device truncate is top level api which calls each logdev truncate.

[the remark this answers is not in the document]
xiaoxichen: but if we have N logdevs why it is mandatory to truncate them all together?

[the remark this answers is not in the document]
hkadayam: @sanebay So our idea is device_truncate() will be all log_devs and if one want to truncate each device they need to call log_dev->truncate()? If we don't provide the latter, then we need to add it, right?

src/tests/test_common/hs_repl_test_common.hpp
```diff
@@ -172,8 +172,7 @@ class HSReplTestHelper {
             name_ + std::to_string(replica_num_),
             {{HS_SERVICE::META, {.size_pct = 5.0}},
              {HS_SERVICE::REPLICATION, {.size_pct = 60.0, .repl_app = std::make_unique< TestReplApplication >(*this)}},
-             {HS_SERVICE::LOG_REPLICATED, {.size_pct = 20.0}},
-             {HS_SERVICE::LOG_LOCAL, {.size_pct = 2.0}}});
+             {HS_SERVICE::LOG, {.size_pct = 22.0}}});
```
hkadayam, line 175: Nit: We can still retain 20% :-). That 2% was because we can't give 0%.

hkadayam (approved):
Have some minor comments, but don't want to hold this PR merge. I am fine in merge this and we can adjust minor things in next checkin.

src/tests/test_common/hs_repl_test_common.hpp
sanebay: Yeah :) , just trying to avoid any regression.

src/include/homestore/logstore_service.hpp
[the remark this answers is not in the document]
sanebay: Yeah that is right. Logstoreservice has a method for it. LogStoreService::get_logdev(logdev_id_t id) and then do LogDev->truncate.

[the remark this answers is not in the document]
sanebay: > but if we have N logdevs why it is mandatory to truncate them all together?

Today log store service does for all families, so we keeping the same.. We might do at log device level later if needed.

shosseinimotlagh (approved):

PR #284, opened 2024-01-23
fix repl service

JacksonYao287:
`GenericReplService` is created in `HomeStore::start` and  will get the uuid through `get_my_repl_id()` of `ReplApplication` . 

when it is first-time-boot, `ReplApplication` does not get its real uuid in `HomeStore::start`. in HO case,  it will get the uuid by `discover_svcid` after `HomeStore::start` returns and before `HomeStore::format_and_start`. so , if we build `GenericReplService` in `HomeStore::start` for first-time-boot , incorrect uuid(default value) will be got.

this PR move the construction of `GenericReplService` to `HomeStore::do_start`, so that no matter it is the first time boot, we can make sure `GenericReplService` will get a correct uuid from `ReplApplication`.


xiaoxichen (commented):
We cannot explain why in @JacksonYao287 's environment the initial value of boost::uuid is not nil value..


I do think we need to have an assert in HO `our_uuid()` to gate the  `!_our_id.is_nil()` access to uuid before its properly initialized but based on what we see above for unknown reason it seems the default value can pass... 
```
    peer_id_t our_uuid() const final { return _our_id; }
```
===================
```
    peer_id_t test1;
    LOGI("Initializing HomeObject test1 uuid {}", to_string(test1));
    auto instance = std::make_shared< HSHomeObject >(std::move(application));
    LOGI("Initializing HomeObject uuid {}", to_string(instance->our_uuid()));
    peer_id_t test2;
    LOGI("Initializing HomeObject test2 uuid {}", to_string(test2));
```
log
```
[01/23/24 03:34:59.229] [storage_mgr] [info] [2815253] [hs_homeobject.cpp:35:init_homeobject] Initializing HomeObject test1 uuid d011982f-0100-0000-9c23-832f60550000
[01/23/24 03:34:59.229] [storage_mgr] [info] [2815253] [homeobject_impl.cpp:26:HomeObjectImpl] initialized with [executor=immediate]
[01/23/24 03:34:59.229] [storage_mgr] [info] [2815253] [hs_homeobject.cpp:37:init_homeobject] Initializing HomeObject uuid 6275672c-2062-6c6f-626d-67723d646562
[01/23/24 03:34:59.229] [storage_mgr] [info] [2815253] [hs_homeobject.cpp:40:init_homeobject] Initializing HomeObject test2 uuid 6275672c-2062-6c6f-626d-67723d646562
```

src/include/homestore/replication/repl_dev.h
```diff
@@ -101,8 +101,8 @@ class ReplDevListener {
 public:
     virtual ~ReplDevListener() = default;
 
-    void set_repl_dev(ReplDev* rdev) { m_repl_dev = rdev; }
-    virtual ReplDev* repl_dev() { return m_repl_dev; }
+    void set_repl_dev(shared< ReplDev > rdev) { m_repl_dev = rdev; }
+    virtual shared< ReplDev > repl_dev() { return m_repl_dev; }
```
szmyd, line 105: This `virtual` confuses me; is it meant to be called by a supposed child class? e.g.:
```c++
class MyRDL : public ReplDevListener {
   ...
   shared< ReplDev > repl_dev() { ...// do something; return ReplDevListener::repl_dev(); }
};
```
If this is the case I would suggest a different pattern...add a pure virtual that is used that someone _must_ implement and leave this non-virtual, or just remove the `virtual` altogether?

src/include/homestore/replication/repl_dev.h
```diff
@@ -170,7 +170,7 @@ class ReplDevListener {
     virtual void on_replica_stop() = 0;
 
 private:
-    ReplDev* m_repl_dev;
+    shared< ReplDev > m_repl_dev;
```
szmyd, line 173: Doesn't this add a circular dependency on ReplDev? It can never be destroyed now since it will never destroy the Listener...and so on.

src/include/homestore/replication/repl_dev.h
JacksonYao287: this is a default action for this virtual function. it means if the derived class does not implement this function explicitly, it will use this as the default.

src/include/homestore/replication/repl_dev.h
JacksonYao287: yes, it will cause circular dependency. i have rollback this change, will deal with this issue in a later PR

xiaoxichen (dismissed):

xiaoxichen (approved):
lgtm

src/include/homestore/replication/repl_dev.h
szmyd: Hmmm, if there is a use-case for it fine...but it adds a vtable lookup to call regardless if someone overrides it or not. Shouldn't set_repl_dev be virtual too if the getter is overrideable?

xiaoxichen (approved):
removed the virtual for now as no use case.

xiaoxichen (approved):

Issue #279, opened 2024-01-19
Timedout repl_req to be garbage collected when originator is crashed/rebooted


hkadayam:
Each replication request is tracked as repl_req structure. In followers they are created upon leader sending data either over raft or data channel (whichever comes first). There is a possibility of the request failed in leader or didn't receive quorum, resulting in these requests are not going to be further processed. It occupies system resources (including blks allocated). It needs to be garbage collected with timer.

Idea is to start a timer and walks through every ReplDev and look at the map structure and iterate over them based on time and free up the resources.

PR #285, opened 2024-01-23
Restartability of ReplDev

hkadayam:
Following changes are done as part of this commit

1. Because of ordering of services, restart of homestore doesn't recover the ReplDev does not work. Fixed the ordering and also made logstore and data service start/stop inside ReplService start/stop.

2. Error handling was not present, as a result tests failure were hard to comprehend. Added error handling of propose, commit, write failures

3. Aesthetical change to have open_log_store to return future< logstore > instead of callback

4. Added become_leader() call to ReplDev and fixed test cases to utilize that to ensure write happens/issued on correct leader.

5. Added testing to restart, write after restart, validate of ReplDev

### DIFF
--- a/src/lib/homestore.cpp
+++ b/src/lib/homestore.cpp
@@ -116,19 +116,23 @@ bool HomeStore::start(const hs_input_params& input, hs_before_services_starting_
     std::call_once(flag1, [this]() {
         m_periodic_logger =
             sisl::logging::CreateCustomLogger("homestore", "_periodic", false, true /* tee_to_stdout_stderr */);
+        sisl::logging::SetLogPattern("[%D %T.%f] [%^%L%$] [%t] %v", m_periodic_logger);
     });
-    sisl::logging::SetLogPattern("[%D %T.%f] [%^%L%$] [%t] %v", m_periodic_logger);
 
     HomeStoreDynamicConfig::init_settings_default();
 
     LOGINFO("Homestore is loading with following services: {}", m_services.list());
     if (has_meta_service()) { m_meta_service = std::make_unique< MetaBlkService >(); }
-    if (has_log_service()) { m_log_service = std::make_unique< LogStoreService >(); }
-    if (has_data_service()) { m_data_service = std::make_unique< BlkDataService >(std::move(s_custom_chunk_selector)); }
     if (has_index_service()) { m_index_service = std::make_unique< IndexService >(std::move(s_index_cbs)); }
     if (has_repl_data_service()) {
         m_repl_service = GenericReplService::create(std::move(s_repl_app));
+        m_log_service = std::make_unique< LogStoreService >();
         m_data_service = std::make_unique< BlkDataService >(std::move(s_custom_chunk_selector));
+    } else {
+        if (has_log_service()) { m_log_service = std::make_unique< LogStoreService >(); }
+        if (has_data_service()) {
+            m_data_service = std::make_unique< BlkDataService >(std::move(s_custom_chunk_selector));
+        }
     }
     m_cp_mgr = std::make_unique< CPManager >();
     m_dev_mgr = std::make_unique< DeviceManager >(input.devices, bind_this(HomeStore::create_vdev_cb, 2));
@@ -208,17 +212,18 @@ void HomeStore::do_start() {
 
     if (has_index_service()) { m_index_service->start(); }
 
-    if (has_data_service()) {
-        m_data_service->start();
-    } else if (has_repl_data_service()) {
-        m_data_service->start();
-        s_cast< GenericReplService* >(m_repl_service.get())->start();
+    if (has_repl_data_service()) {
+        s_cast< GenericReplService* >(m_repl_service.get())->start(); // Replservice starts logstore & data service
+    } else {
+        if (has_data_service()) { m_data_service->start(); }
+        if (has_log_service() && inp_params.auto_recovery) {
+            // In case of custom recovery, let consumer starts the recovery and it is consumer module's responsibilities
+            // to start log store
+            m_log_service->start(is_first_time_boot() /* format */);
+        }
     }
 
-    // In case of custom recovery, let consumer starts the recovery and it is consumer module's responsibilities
-    // to start log store
-    if (has_log_service() && inp_params.auto_recovery) { m_log_service->start(is_first_time_boot() /* format */); }
-
+    m_cp_mgr->start_timer();
     m_init_done = true;
 }
 
@@ -234,8 +239,17 @@ void HomeStore::shutdown() {
     m_cp_mgr.reset();
 
     if (has_repl_data_service()) {
+        // Log and Data services are stopped by repl service
         s_cast< GenericReplService* >(m_repl_service.get())->stop();
+        m_log_service.reset();
+        m_data_service.reset();
         m_repl_service.reset();
+    } else {
+        if (has_log_service()) {
+            m_log_service->stop();
+            m_log_service.reset();
+        }
+        if (has_data_service()) { m_data_service.reset(); }
     }
 
     if (has_index_service()) {
@@ -243,17 +257,10 @@ void HomeStore::shutdown() {
         // m_index_service.reset();
     }
 
-    if (has_log_service()) {
-        m_log_service->stop();
-        m_log_service.reset();
-    }
-
     if (has_meta_service()) {
         m_meta_service->stop();
         m_meta_service.reset();
     }
-
-    if (has_data_service()) { m_data_service.reset(); }
 
     m_dev_mgr->close_devices();
     m_dev_mgr.reset();
--- a/src/lib/logstore/log_store_family.cpp
+++ b/src/lib/logstore/log_store_family.cpp
@@ -63,7 +63,7 @@ void LogStoreFamily::start(bool format, JournalVirtualDev* blk_store) {
         // Also call the logstore to inform that start/replay is completed.
         if (!format) {
             for (auto& p : m_id_logstore_map) {
-                auto& lstore{p.second.m_log_store};
+                auto& lstore{p.second.log_store};
                 if (lstore && lstore->get_log_replay_done_cb()) {
                     lstore->get_log_replay_done_cb()(lstore, lstore->seq_num() - 1);
                     lstore->truncate(lstore->truncated_upto());
@@ -90,18 +90,25 @@ std::shared_ptr< HomeLogStore > LogStoreFamily::create_new_log_store(bool append
         folly::SharedMutexWritePriority::WriteHolder holder(m_store_map_mtx);
         const auto it = m_id_logstore_map.find(store_id);
         HS_REL_ASSERT((it == m_id_logstore_map.end()), "store_id {}-{} already exists", m_family_id, store_id);
-        m_id_logstore_map.insert(std::make_pair<>(store_id, logstore_info_t{lstore, nullptr, append_mode}));
+        m_id_logstore_map.insert(std::pair(store_id, logstore_info{.log_store = lstore, .append_mode = append_mode}));
     }
     LOGINFO("Created log store id {}-{}", m_family_id, store_id);
     return lstore;
 }
 
-void LogStoreFamily::open_log_store(logstore_id_t store_id, bool append_mode, const log_store_opened_cb_t& on_open_cb) {
+folly::Future< shared< HomeLogStore > > LogStoreFamily::open_log_store(logstore_id_t store_id, bool append_mode) {
     folly::SharedMutexWritePriority::WriteHolder holder(m_store_map_mtx);
-    const auto it = m_id_logstore_map.find(store_id);
-    HS_REL_ASSERT((it == m_id_logstore_map.end()), "store_id {}-{} already exists", m_family_id, store_id);
-    LOGINFO("Opening log store id {}-{}", m_family_id, store_id);
-    m_id_logstore_map.insert(std::make_pair<>(store_id, logstore_info_t{nullptr, on_open_cb, append_mode}));
+    auto it = m_id_logstore_map.find(store_id);
+    if (it == m_id_logstore_map.end()) {
+        bool happened;
+        std::tie(it, happened) = m_id_logstore_map.insert(std::pair(store_id,
+                                                                    logstore_info{
+                                                                        .log_store = nullptr,
+                                                                        .append_mode = append_mode,
+                                                                    }));
+        HS_REL_ASSERT_EQ(happened, true, "Unable to insert logstore into id_logstore_map");
+    }
+    return it->second.promise.getFuture();
 }
 
 void LogStoreFamily::remove_log_store(logstore_id_t store_id) {
@@ -139,9 +146,10 @@ void LogStoreFamily::device_truncate(const std::shared_ptr< truncate_req >& treq
 
 void LogStoreFamily::on_log_store_found(logstore_id_t store_id, const logstore_superblk& sb) {
     folly::SharedMutexWritePriority::ReadHolder holder(m_store_map_mtx);
-    const auto it = m_id_logstore_map.find(store_id);
+    auto it = m_id_logstore_map.find(store_id);
     if (it == m_id_logstore_map.end()) {
-        LOGERROR("Store Id {}-{} found but not opened yet.", m_family_id, store_id);
+        LOGERROR("Store Id {}-{} found but not opened yet, it will be discarded after logstore is started", m_family_id,
+                 store_id);
         m_unopened_store_id.insert(store_id);
         m_unopened_store_io.insert(std::make_pair<>(store_id, 0));
         return;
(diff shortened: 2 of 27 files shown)


src/lib/homestore.cpp
```diff
@@ -234,27 +239,29 @@ void HomeStore::shutdown() {
     m_cp_mgr.reset();
 
     if (has_repl_data_service()) {
+        // Log and Data services are stopped by repl service
         s_cast< GenericReplService* >(m_repl_service.get())->stop();
+        m_log_service.reset();
```
yamingk, line 244: since we are making log service and data service inside repl service start, should we also make each one of them stop inside m_repl_service->stop() as well?

src/lib/logstore/log_store_family.cpp
```diff
@@ -286,8 +294,8 @@ nlohmann::json LogStoreFamily::dump_log_store(const log_dump_req& dump_req) {
     if (dump_req.log_store == nullptr) {
         folly::SharedMutexWritePriority::ReadHolder holder(m_store_map_mtx);
         for (auto& id_logstore : m_id_logstore_map) {
-            auto store_ptr{id_logstore.second.m_log_store};
-            const std::string id{std::to_string(store_ptr->get_store_id())};
+            auto store_ptr = id_logstore.second.log_store;
```
yamingk, line 297: nit: auto const&

src/lib/logstore/log_store_family.cpp
```diff
@@ -286,8 +294,8 @@ nlohmann::json LogStoreFamily::dump_log_store(const log_dump_req& dump_req) {
     if (dump_req.log_store == nullptr) {
         folly::SharedMutexWritePriority::ReadHolder holder(m_store_map_mtx);
```
yamingk, line 295: nitpick: now that `mutable` is applied, probably we can make this dump_log_store as const function as well?
you can ignore this comment if it becomes messy or you want to keep as it is.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -38,11 +38,12 @@ RaftReplDev::RaftReplDev(RaftReplService& svc, superblk< raft_repl_dev_superblk
...
         if (m_rd_sb->is_timeline_consistent) {
-            logstore_service().open_log_store(LogStoreService::CTRL_LOG_FAMILY_IDX, m_rd_sb->free_blks_journal_id,
-                                              false, [this](shared< HomeLogStore > log_store) {
-                                                  m_free_blks_journal = std::move(log_store);
-                                                  m_rd_sb->free_blks_journal_id = m_free_blks_journal->get_store_id();
-                                              });
+            logstore_service()
+                .open_log_store(LogStoreService::CTRL_LOG_FAMILY_IDX, m_rd_sb->free_blks_journal_id, false)
+                .thenValue([this](auto log_store) {
+                    m_free_blks_journal = std::move(log_store);
```
yamingk, line 44: is `m_free_blks_journal` for future usage? Any free_blks (calling from HomeObject) will be journalized by this log store and truncated after cp was done upto certain point? I didn't see where we used it yet in homestore code yet, probably this part is going to be implemented?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -171,10 +240,14 @@ void RaftReplDev::on_push_data_received(intrusive< sisl::GenericRpcData >& rpc_d
     data_service()
```
yamingk, line 240: at line: 220, I don't quite understand while checking the DATA_RECEIVED, it will always be true and return, right?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -120,15 +144,60 @@ void RaftReplDev::push_data_to_all_followers(repl_req_ptr_t rreq) {
...
+               err);
+        return;
+    }
+
+    // Free the blks which is allocated already
+    RD_LOG(INFO, "Raft Channel: Error in processing rreq=[{}] error={}", rreq->to_compact_string(), err);
+    if (rreq->state.load() & uint32_cast(repl_req_state_t::BLK_ALLOCATED)) {
+        auto blkid = rreq->local_blkid;
+        data_service().async_free_blk(blkid).thenValue([blkid](auto&& err) {
+            HS_LOG_ASSERT(!err, "freeing blkid={} upon error failed, potential to cause blk leak", blkid.to_string());
```
yamingk, line 177: in free callback, we probably should also reset rreq->state to remove BLK_ALLOCATED and DATA_RECEIVED, right?

yamingk:
update conan version

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -85,18 +98,29 @@ void RaftReplDev::async_alloc_write(sisl::blob const& header, sisl::blob const&
         auto status = data_service().alloc_blks(uint32_cast(rreq->value.size),
                                                 m_listener->get_blk_alloc_hints(rreq->header, rreq->value.size),
                                                 rreq->local_blkid);
-        HS_REL_ASSERT_EQ(status, BlkAllocStatus::SUCCESS);
+        if (status != BlkAllocStatus::SUCCESS) {
```
JacksonYao287, line 101: if come here, should we send a message to drop the data pushed previously?

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -100,22 +101,68 @@ void RaftReplService::start() {
...
+    for (auto const& [buf, mblk] : m_config_sb_bufs) {
+        raft_group_config_found(buf, voidptr_cast(mblk));
+    }
+    m_config_sb_bufs.clear();
+
+    // Step 5: Start the data and logstore service now. This step is essential before we can ask Raft to join groups etc
+    hs()->data_service().start();
+    hs()->logstore_service().start(hs()->is_first_time_boot());
+
+    // Step 6: Iterate all the repl dev and ask each one of the join the raft group.
```
JacksonYao287, line 126: do we have any mechanism to check partial `raft_repl_dev`?
for example, when creating raft_repl_dev, the leader succeed adding the first follower to the group , but fail to add the second follower. 

now for this case , we just return an error to upper layer , not do any cleanup. i think it is the reponsibility of the raft repl severice to cleanup up all the partial raft group, even if restart, so that we can make sure no leak of raft_repl_dev

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -120,15 +144,60 @@ void RaftReplDev::push_data_to_all_followers(repl_req_ptr_t rreq) {
...
+
+    auto s = rreq->state.load();
+    if ((s & uint32_cast(repl_req_state_t::ERRORED)) ||
+        !(rreq->state.compare_exchange_strong(s, s | uint32_cast(repl_req_state_t::ERRORED)))) {
+        RD_LOG(INFO, "Raft Channel: Error in processing rreq=[{}] error={} already errored", rreq->to_compact_string(),
+               err);
+        return;
+    }
+
+    // Free the blks which is allocated already
```
yamingk, line 172: i believe in 'follower_create_req' function which calls 'do_blk_alloc', it also needs to check the status and call this handle_error, correct?

yamingk:
Looks good to me except one comment. Other comments are either minor or informational purpose.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -120,15 +144,60 @@ void RaftReplDev::push_data_to_all_followers(repl_req_ptr_t rreq) {
 
     group_msg_service()
         ->data_service_request_unidirectional(nuraft_mesg::role_regex::ALL, PUSH_DATA, rreq->pkts)
```
sanebay, line 146: whats the behaviour if the raft log append succeeded but the data push failed for one of the follower ?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -171,10 +240,14 @@ void RaftReplDev::on_push_data_received(intrusive< sisl::GenericRpcData >& rpc_d
     data_service()
         .async_write(r_cast< const char* >(data), push_req->data_size(), rreq->local_blkid)
         .thenValue([this, rreq](auto&& err) {
-            RD_REL_ASSERT(!err, "Error in writing data"); // TODO: Find a way to return error to the Listener
-            rreq->state.fetch_or(uint32_cast(repl_req_state_t::DATA_WRITTEN));
-            rreq->data_written_promise.setValue();
-            RD_LOG(INFO, "Data Channel: Data Write completed rreq=[{}]", rreq->to_compact_string());
+            if (err) {
+                RD_DBG_ASSERT(false, "Error in writing data");
+                handle_error(rreq, ReplServiceError::DRIVE_WRITE_ERROR);
```
sanebay, line 245: If there are any errors, we should free the blks allocated ?

src/lib/replication/service/generic_repl_svc.cpp
```diff
@@ -36,15 +37,20 @@ std::shared_ptr< GenericReplService > GenericReplService::create(cshared< ReplAp
 
 GenericReplService::GenericReplService(cshared< ReplApplication >& repl_app) :
         m_repl_app{repl_app}, m_my_uuid{repl_app->get_my_repl_id()} {
+    m_sb_bufs.reserve(100);
```
sanebay, line 40: some dynamic config MAX_NUM_PG_IN_SYSTEM ?

src/lib/homestore.cpp
```diff
@@ -116,19 +116,23 @@ bool HomeStore::start(const hs_input_params& input, hs_before_services_starting_
...
 
     HomeStoreDynamicConfig::init_settings_default();
 
     LOGINFO("Homestore is loading with following services: {}", m_services.list());
     if (has_meta_service()) { m_meta_service = std::make_unique< MetaBlkService >(); }
-    if (has_log_service()) { m_log_service = std::make_unique< LogStoreService >(); }
-    if (has_data_service()) { m_data_service = std::make_unique< BlkDataService >(std::move(s_custom_chunk_selector)); }
     if (has_index_service()) { m_index_service = std::make_unique< IndexService >(std::move(s_index_cbs)); }
     if (has_repl_data_service()) {
         m_repl_service = GenericReplService::create(std::move(s_repl_app));
```
sanebay, line 128: Repl and data services are not mutually exclusive now with_repl_data_service()

src/tests/test_raft_repl_dev.cpp
```diff
@@ -127,16 +131,23 @@ class TestReplicatedDB : public homestore::ReplDevListener {
         LOGINFO("[Replica={}] Received rollback on lsn={}", g_helper->replica_num(), lsn);
     }
 
+    void on_error(ReplServiceError error, const sisl::blob& header, const sisl::blob& key,
+                  cintrusive< repl_req_ctx >& ctx) override {
```
sanebay, line 135: assert() as we are not expecting any error.

src/lib/homestore.cpp
```diff
@@ -208,17 +212,18 @@ void HomeStore::do_start() {
...
+            // to start log store
+            m_log_service->start(is_first_time_boot() /* format */);
+        }
     }
 
-    // In case of custom recovery, let consumer starts the recovery and it is consumer module's responsibilities
-    // to start log store
-    if (has_log_service() && inp_params.auto_recovery) { m_log_service->start(is_first_time_boot() /* format */); }
-
+    m_cp_mgr->start_timer();
```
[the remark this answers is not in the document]
hkadayam: I don't think CP itself is providing any service as such, it is an internal requirement which upper layer can tag along. Not sure it will buy us much.

src/lib/homestore.cpp
hkadayam: I probably didn't quiet catch your comment. Can you please elaborate?

src/lib/logstore/log_store_family.cpp
hkadayam: My changes are only to replace { to = for better readability, but I will do as much as I can the one you pointed out in this code.

src/lib/replication/repl_dev/raft_repl_dev.cpp
hkadayam: m_free_blks_journal is needed if we set that we need `timeline_consistent`, which is not required for HomeObject, since there is no mutation for the key. It is needed for HomeBlocks.

src/lib/replication/repl_dev/raft_repl_dev.cpp
hkadayam: handle_error method does that.

src/tests/test_raft_repl_dev.cpp
hkadayam: Well, thought about that, but I wanted to extend this for negative check, where propose on leader and check error etc.. So didn't put it here. Later we will add counter and validate that.

src/lib/replication/service/raft_repl_service.cpp
hkadayam: We will be adding "remove_repl_dev" api as part of next work. There will be adding "destroy" the repl_dev functionality (which we can use for all error handling).

src/lib/replication/service/generic_repl_svc.cpp
```diff
@@ -120,7 +139,7 @@ void SoloReplService::load_repl_dev(sisl::byte_view const& buf, void* meta_cooki
     {
         std::unique_lock lg(m_rd_map_mtx);
         auto [_, happened] = m_rd_map.emplace(group_id, rdev);
-        (void) happened;
+        (void)happened;
```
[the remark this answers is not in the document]
hkadayam: I don't remember adding this line, but it is needed for optimized build to succeed (unused variable happened)

src/lib/replication/service/generic_repl_svc.cpp
hkadayam: This is just for reserve and not for any tunables? Do we need to clutter the dynamic config with this (also homestore doesn't know anything about PG).

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -141,14 +165,14 @@ repl_req_ptr_t RaftStateMachine::transform_journal_entry(nuraft::ptr< nuraft::lo
         auto new_buf = nuraft::buffer::expand(*rreq->raft_journal_buf(),
```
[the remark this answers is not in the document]
hkadayam: No, but in other use cases (not HomeObject), blocks allocated are not contiguous (so MultiBlkId could be more than one entry). It could be different on different replicas right? So one proposer it could be contiguous blocks and hence serialized of the BlkId is just 1 byte, whereas it could be multiple in one follower which could 4 blocks and hence 6 bytes. Here we replace them by allocating a fresh buffer and copy them (it should be rare but can happen). For HomeObject use case, this code block will not be excercised.

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -109,25 +117,41 @@ repl_req_ptr_t RaftStateMachine::transform_journal_entry(nuraft::ptr< nuraft::lo
...
+        repl_journal_entry* tmp_jentry = r_cast< repl_journal_entry* >(lentry->get_buf().data_begin());
+        RELEASE_ASSERT_EQ(tmp_jentry->major_version, repl_journal_entry::JOURNAL_ENTRY_MAJOR,
+                          "Mismatched version of journal entry received from RAFT peer");
+
+        RD_LOG(TRACE, "Received Raft log_entry=[term={}], journal_entry=[{}] ", lentry->get_term(),
+               tmp_jentry->to_string());
 
-    RD_LOG(TRACE, "Received Raft log_entry=[term={}], journal_entry=[{}] ", lentry->get_term(), jentry->to_string());
+        // For inline data we don't need to transform anything
+        if (tmp_jentry->code != journal_type_t::HS_LARGE_DATA) { return nullptr; }
```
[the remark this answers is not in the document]
hkadayam: Will do as part of my next PR, as I am changing some of these parameters for destroy_repl_dev feature.

src/lib/replication/repl_dev/raft_repl_dev.cpp
hkadayam: If data is not pushed to only one follower, leader doesn't get any error. The follower which has not received the data, will do a fetch (after a timeout) and get the data. In the meanwhile quorum would have reached for that log entry and committed (assuming there is one more healthy follower).

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -120,15 +144,60 @@ void RaftReplDev::push_data_to_all_followers(repl_req_ptr_t rreq) {
...
+                return;
+            }
             // Release the buffer which holds the packets
             RD_LOG(INFO, "Data Channel: Data push completed for rreq=[{}]", rreq->to_compact_string());
             rreq->fb_builder.Release();
             rreq->pkts.clear();
         });
 }
 
+void RaftReplDev::handle_error(repl_req_ptr_t const& rreq, ReplServiceError err) {
```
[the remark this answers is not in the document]
hkadayam: I thought about that, may be we need to, but right now I felt this scenario and many scenarios will be covered by ticket #279 which garbage collects all of them and independently call handle_error().

src/lib/replication/repl_dev/raft_repl_dev.cpp
hkadayam: Good point. Today that is not recoverable error and we crash. I am not sure we can handle the error. Perhaps we can do the discussion and handling as part of "Error recovery/validation" strategy.

src/lib/replication/repl_dev/raft_repl_dev.cpp
hkadayam: The idea of the state is to see what stages we have come across. Felt that it would be meaningful while debugging instead of erasing. Once errors are handled, there is not much benefit to these flags other than debugging purpose.

JacksonYao287 (approved):
LGTM, no more comments from my side, thx for the work @hkadayam

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: yes, sounds good

yamingk (approved):

sanebay (approved):

PR #287, opened 2024-01-24
Issue#257 Replication Fetch Remote Data

yamingk:
Changes made:
1. on follower, register FETCH_DATA to nuraft_mesg and add fetch remote data logic.
2. on leader side, handle fetch request and read data based on received remote blkid if this leader is the originator.
3. on follower, write the received data into drive and mark data written and trigger promise to wake up batch_append to moving forward.
4. introduce flip to simulate data not received from data channel (ingore data written) to trigger a fetch remote data.

Issues:
Start iomgr timer will yield a crash -- investigating.

Testing:
=======
1. unit test passed with a shutdown issue (being looked up). 
2. somehow with raft test application, after a test case complete, the next test case will stuck with ipc communication channel, still investigation why. By running the new test case alone is good.


src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -95,6 +95,7 @@ void RaftReplService::start() {
                         .with_stale_log_gap(HS_DYNAMIC_CONFIG(consensus.stale_log_gap_hi_threshold))
                         .with_fresh_log_gap(HS_DYNAMIC_CONFIG(consensus.stale_log_gap_lo_threshold))
                         .with_snapshot_enabled(HS_DYNAMIC_CONFIG(consensus.snapshot_freq_distance))
+                        .with_leadership_expiry(-1 /* never expires */) // >>> debug only
```
yamingk, line 98: will be updated

PR #294, opened 2024-01-30
Add get_leader_id/get_replication_status into replicationService

xiaoxichen:


src/include/homestore/replication_service.hpp
```diff
@@ -52,6 +61,14 @@ class ReplicationService {
...
 
     virtual meta_sub_type get_meta_blk_name() const = 0;
+
+    /// @brief get the leader replica_id of given group
+    virtual const replica_id_t get_leader_id(group_id_t group_id) const = 0;
+
+    /// @brief get replication status. If called on follower member
+    /// this API can return empty result.
+    virtual std::vector<peer_info> get_replication_status(group_id_t group_id) const = 0;
+
```
hkadayam, line 71: Is there a use case where we have this api in ReplicationService or can this be at every ReplDev itself, because we have ReplDev::is_leader() already and we could do ReplDev::get_leader() and get_replication_status() there right?

xiaoxichen: Yes there is no strong requirement to have that in ReplicationService as in ReplicationService it also get the ReplDev and operate based on that.. was trying to have a wrapper so that client(external HS) doesnt need to get ReplDev to get the stats when they want to report metrics.  However in HO we already keep the ReplDev ptr in HS_PG.

Will move those to ReplDev.

src/include/homestore/replication/repl_decls.h
```diff
@@ -65,6 +65,15 @@ using remote_blkid_list_t = folly::small_vector< RemoteBlkId, 4 >;
 using replica_id_t = uuid_t;
 using group_id_t = uuid_t;
 
+struct peer_info {
```
yamingk, line 68: is there a reason we can't use the same structure defined in nuraft_mes: https://github.com/eBay/nuraft_mesg/blob/main/include/nuraft_mesg/mesg_state_mgr.hpp#L31

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -12,6 +12,7 @@
 #include "replication/service/raft_repl_service.h"
 #include "replication/repl_dev/raft_repl_dev.h"
 #include "push_data_rpc_generated.h"
+#include "boost/lexical_cast.hpp"
```
hkadayam, line 15: nit: It is not mandatory, but common practice we decided to follow was to define the 
STL defs, followed by external libraries both indicate by <> (instead of "") and then followed by local repo includes.

src/include/homestore/replication/repl_decls.h
hkadayam: We are not exposing nuraft_mesg to the upper layer, so we probably have to redefine here.

hkadayam (approved):
One additional minor comment from me, otherwise look good to me.

src/include/homestore/replication/repl_decls.h
xiaoxichen: My thinking is 
1. the structure field (mostly `last_log_idx_`) is too implementation(raft) specific,  as a generic API for all repl-dev I am trying to use more neutral naming.
2. HS/HO we use replica_id_t,  translate it from string to replica_id_t make it more consistent from the view of HO.

JacksonYao287 (approved):
LGTM

PR #299, opened 2024-01-31
Fix log dev load issue.

sanebay:
Issue.
Getting asserts in logdev during recovery with mismatch in logdev idx and complaining header corrupted.

Cause
One of the logdev family 1 was not used for writes in UT and has no chunks. During recovery, when we read m_vdev_jd->sync_next_read(), we ignored the return value whether we read anything or not. On build machines, the return buffer contained garbage values, which later was considered as invalid header and end of log. log_stream_reader::next_group() line 58. For some reason on github actions, we get a proper loggroup header(from logdev family 0, because crc was matching with sequence number) as garbage value and causing it validate against a invalid header and doing a assert. This issue didnt happen before dynamic logdev changes because we had fixed number chunks for both logdev families and it read some junk and returned  error with no more logdev data. Now we have situation of 0 chunks and no garbage data.

Fix
handle the return code m_vdev_jd->sync_next_read() and handle case where then is no data needs to be read.

### DIFF
--- a/src/lib/device/virtual_dev.cpp
+++ b/src/lib/device/virtual_dev.cpp
@@ -125,8 +125,8 @@ void VirtualDev::add_chunk(cshared< Chunk >& chunk, bool is_fresh_chunk) {
 
 void VirtualDev::remove_chunk(cshared< Chunk >& chunk) {
     std::unique_lock lg{m_mgmt_mutex};
-    auto iter = std::remove_if(m_all_chunks.begin(), m_all_chunks.end(), [chunk](auto c) { return c == chunk; });
-    m_all_chunks.erase(iter, m_all_chunks.end());
+    m_all_chunks[chunk->chunk_id()].reset();
+    m_all_chunks[chunk->chunk_id()] = nullptr;
     m_chunk_selector->remove_chunk(chunk);
 }
 
--- a/src/lib/logstore/log_dev.cpp
+++ b/src/lib/logstore/log_dev.cpp
@@ -108,6 +108,7 @@ void LogDev::start(bool format, JournalVirtualDev* vdev) {
 }
 
 void LogDev::stop() {
+    THIS_LOGDEV_LOG(INFO, "Logdev stopping family {}", m_family_id);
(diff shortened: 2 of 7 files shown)


src/lib/logstore/log_dev.cpp
```diff
@@ -88,7 +88,7 @@ void LogDev::start(bool format, JournalVirtualDev* vdev) {
             m_store_found_cb(spair.first, spair.second);
         }
 
-        THIS_LOGDEV_LOG(INFO, "get start vdev offset during recovery {} log indx {} ",
+        THIS_LOGDEV_LOG(INFO, "family {} get start vdev offset during recovery {} log indx {} ", m_family_id,
```
hkadayam, line 91: You can change THIS_LOGDEV_LOG to add the family in everyone so we don't need to repeat it everywhere?

src/lib/device/virtual_dev.cpp
```diff
@@ -125,8 +125,8 @@ void VirtualDev::add_chunk(cshared< Chunk >& chunk, bool is_fresh_chunk) {
 
 void VirtualDev::remove_chunk(cshared< Chunk >& chunk) {
     std::unique_lock lg{m_mgmt_mutex};
-    auto iter = std::remove_if(m_all_chunks.begin(), m_all_chunks.end(), [chunk](auto c) { return c == chunk; });
-    m_all_chunks.erase(iter, m_all_chunks.end());
+    m_all_chunks[chunk->chunk_id()].reset();
+    m_all_chunks[chunk->chunk_id()] = nullptr;
```
hkadayam, line 129: Are we changing from sparse_vector to map (with rcu) in different PR?

src/lib/logstore/log_dev.cpp
sanebay: I checked it, it was already there, just added because couldnt see when testing.

src/lib/device/virtual_dev.cpp
sanebay: Yeah that will be done in a different PR.

hkadayam (approved):

shosseinimotlagh (approved):

PR #302, opened 2024-02-02
Enable all btree types and do clean restart

shosseinimotlagh:


sanebay (approved):

src/tests/CMakeLists.txt
```diff
@@ -72,9 +72,9 @@ if (${io_tests})
     set(TEST_INDEXBTREE_SOURCE_FILES test_index_btree.cpp)
     add_executable(test_index_btree ${TEST_INDEXBTREE_SOURCE_FILES})
     target_link_libraries(test_index_btree homestore ${COMMON_TEST_DEPS} GTest::gtest)
-    add_test(NAME IndexBtree COMMAND test_index_btree)
```
sanebay, line 75: Please add a todo comment to enable other btree types.

src/tests/test_index_btree.cpp
```diff
@@ -501,7 +501,7 @@ struct BtreeConcurrentTest : public BtreeTestHelper< TestType >, public ::testin
     }
 
 private:
-    const std::string m_shadow_filename = "shadow_map.txt";
+    const std::string m_shadow_filename = "/tmp/shadow_map.txt";
```
sanebay, line 504: This will conflict for different users using same build machine. Can you keep it same and delete it after UT runs ?

shosseinimotlagh: This file will be deleted after the last run of each type` function_normal(0, True, False, type) # cleanup after shutdown`

sanebay (approved):

sanebay (approved):

PR #310, opened 2024-02-08
Add lock is blk alloced to fix assert.

sanebay:
There seems a concurrency issue when is blk alloced fails in test_data_service.


yamingk:
can you bump con ver

yamingk (approved):

PR #311, opened 2024-02-08
issue 259/260: incremental resync

yamingk:
This PR relies on another PR in sisl change to build/work:  https://github.com/eBay/sisl/pull/216

1. enable non-recursive timer to be armed and triggered during fetch remote data (normal I/O path) and resync mode. 
2. Add error handling when originator is not there. e.g. orioginator crashed right after sending the appending log entries.
In this case, the member being resyned to is supposed to be receiving log entries from new leader with its corresponding originator id set in rreq, so that the next try of fetch remote will be directed to this new originator.
3. Add test case for incremental resync. Verified that it will batch of rreqs can be fetched fro originator verified the blocks written to disk.
4. staging handling, added dynamic config for maximum data size that can be fetched from remote. Add a test case with flip point to simulate staging to trigger the staging on sending side. Test case passed.

Other tests:
1. release build pass.
3. sanitize build pass.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -57,7 +57,7 @@ class HomestoreConan(ConanFile):
     def requirements(self):
         self.requires("iomgr/[~=11, include_prerelease=True]@oss/master")
         self.requires("sisl/[~=11, include_prerelease=True]@oss/master")
-        self.requires("nuraft_mesg/[~=2, include_prerelease=True]@oss/main")
+        self.requires("nuraft_mesg/[~=3, include_prerelease=True]@oss/main")
 
         self.requires("farmhash/cci.20190513@")
         if self.settings.arch in ['x86', 'x86_64']:
--- a/conanfile.py
+++ b/conanfile.py
@@ -5,7 +5,7 @@ from conans import CMake
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "5.1.3"
+    version = "5.1.4"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
@@ -56,8 +56,10 @@ class HomestoreConan(ConanFile):
 
     def requirements(self):
         self.requires("iomgr/[~=11, include_prerelease=True]@oss/master")
-        self.requires("sisl/[~=11, include_prerelease=True]@oss/master")
-        self.requires("nuraft_mesg/[~=2, include_prerelease=True]@oss/main")
+        #self.requires("sisl/[~=11, include_prerelease=True]@oss/master")
+        self.requires("sisl/11.1.1@arjun/oss")
+        #self.requires("nuraft_mesg/[~=2, include_prerelease=True]@oss/main")
+        self.requires("nuraft_mesg/3.0.1@arjun/oss")
 
         self.requires("farmhash/cci.20190513@")
         if self.settings.arch in ['x86', 'x86_64']:
--- a/conanfile.py
+++ b/conanfile.py
@@ -5,7 +5,7 @@ from conans import CMake
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "5.1.4"
+    version = "5.1.6"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/conanfile.py
+++ b/conanfile.py
@@ -56,10 +56,8 @@ class HomestoreConan(ConanFile):
 
     def requirements(self):
         self.requires("iomgr/[~=11, include_prerelease=True]@oss/master")
-        #self.requires("sisl/[~=11, include_prerelease=True]@oss/master")
-        self.requires("sisl/11.1.1@arjun/oss")
-        #self.requires("nuraft_mesg/[~=2, include_prerelease=True]@oss/main")
-        self.requires("nuraft_mesg/3.0.1@arjun/oss")
+        self.requires("sisl/[~=11, include_prerelease=True]@oss/master")
+        self.requires("nuraft_mesg/[~=2, include_prerelease=True]@oss/main")
 
         self.requires("farmhash/cci.20190513@")
         if self.settings.arch in ['x86', 'x86_64']:
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -188,7 +188,7 @@ void RaftReplDev::on_fetch_data_received(intrusive< sisl::GenericRpcData >& rpc_
         auto const& originator = req->blkid_originator();
         auto const& remote_blkid = req->remote_blkid();
 
-        RD_LOG(INFO, "Data Channel: FetchData received: lsn={}", lsn);
+        RD_LOG(DEBUG, "Data Channel: FetchData received: lsn={}", lsn);
 
         // release this assert if in the future we want to fetch from non-originator;
         RD_REL_ASSERT(originator == server_id(),
@@ -448,7 +448,7 @@ void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
     entries.reserve(rreqs.size());
 
     shared< flatbuffers::FlatBufferBuilder > builder = std::make_shared< flatbuffers::FlatBufferBuilder >();
-    RD_LOG(INFO, "Data Channel : FetchData from remote: rreq.size={}, my server_id={}", rreqs.size(), server_id());
+    RD_LOG(DEBUG, "Data Channel : FetchData from remote: rreq.size={}, my server_id={}", rreqs.size(), server_id());
     auto const& originator = rreqs.front()->remote_blkid.server_id;
 
     for (auto const& rreq : rreqs) {
@@ -496,7 +496,7 @@ void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
             auto raw_data = e.value().response_blob().cbytes();
             auto total_size = e.value().response_blob().size();
 
-            RD_DBG_ASSERT(total_size > 0, "Empty response from remote");
+            RD_DBG_ASSERT_GT(total_size, 0, "Empty response from remote");
             RD_DBG_ASSERT(raw_data, "Empty response from remote");
 
             RD_LOG(INFO, "Data Channel: FetchData completed for reques.size()={} ", rreqs.size());
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -184,15 +184,11 @@ void RaftReplDev::on_fetch_data_received(intrusive< sisl::GenericRpcData >& rpc_
     ctx->outstanding_read_cnt = fetch_req->request()->entries()->size();
 
     for (auto const& req : *(fetch_req->request()->entries())) {
-        RD_LOG(INFO, "Data Channel: FetchData received: lsn={}", req->lsn());
-
         auto const& lsn = req->lsn();
-        auto const& term = req->raft_term();
-        auto const& dsn = req->dsn();
-        auto const& header = req->user_header();
-        auto const& key = req->user_key();
         auto const& originator = req->blkid_originator();
         auto const& remote_blkid = req->remote_blkid();
+
+        RD_LOG(INFO, "Data Channel: FetchData received: lsn={}", lsn);
 
         // release this assert if in the future we want to fetch from non-originator;
         RD_REL_ASSERT(originator == server_id(),
@@ -224,11 +220,6 @@ void RaftReplDev::on_fetch_data_received(intrusive< sisl::GenericRpcData >& rpc_
                 }
                 ctx->cv.notify_one();
             });
-        } else {
-            // TODO: if we are not the originator, we need to fetch based on lsn;
-            // To be implemented;
-            RD_LOG(INFO, "I am not the originaltor for the requested blks, originaltor: {}, server_id: {}.", originator,
-                   server_id());
         }
     }
 
@@ -502,8 +493,8 @@ void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
                 return;
             }
 
-            auto raw_data = e.value()->response_blob().cbytes();
-            auto total_size = e.value()->response_blob().size();
+            auto raw_data = e.value().response_blob().cbytes();
+            auto total_size = e.value().response_blob().size();
 
             RD_DBG_ASSERT(total_size > 0, "Empty response from remote");
             RD_DBG_ASSERT(raw_data, "Empty response from remote");
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -148,7 +148,7 @@ void RaftReplDev::push_data_to_all_followers(repl_req_ptr_t rreq) {
            flatbuffers::FlatBufferToString(builder.GetBufferPointer() + sizeof(flatbuffers::uoffset_t),
                                            PushDataRequestTypeTable()));*/
 
-    RD_LOG(INFO, "Data Channel: Pushing data to all followers: rreq=[{}]", rreq->to_compact_string());
+    RD_LOG(DEBUG, "Data Channel: Pushing data to all followers: rreq=[{}]", rreq->to_compact_string());
 
     group_msg_service()
         ->data_service_request_unidirectional(nuraft_mesg::role_regex::ALL, PUSH_DATA, rreq->pkts)
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -399,6 +399,16 @@ repl_req_ptr_t RaftReplDev::follower_create_req(repl_key const& rkey, sisl::blob
     return rreq;
 }
 
+auto RaftReplDev::get_max_data_fetch_size() const {
+#ifdef _PRERELEASE
+    if (iomgr_flip::instance()->test_flip("simulate_staging_fetch_data")) {
+        LOGINFO("Flip simulate_staging_fetch_data is enabled, return max_data_fetch_size: 16K");
+        return 4 * 4096ull;
+    }
+#endif
+    return HS_DYNAMIC_CONFIG(consensus.data_fetch_max_size_mb) * 1024 * 1024ull;
+}
+
 void RaftReplDev::check_and_fetch_remote_data(std::vector< repl_req_ptr_t >* rreqs) {
     // Pop any entries that are already completed - from the entries list as well as from map
     rreqs->erase(std::remove_if(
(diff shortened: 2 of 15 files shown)


conanfile.py
```diff
@@ -56,8 +56,10 @@ def build_requirements(self):
 
     def requirements(self):
         self.requires("iomgr/[~=11, include_prerelease=True]@oss/master")
-        self.requires("sisl/[~=11, include_prerelease=True]@oss/master")
-        self.requires("nuraft_mesg/[~=2, include_prerelease=True]@oss/main")
+        #self.requires("sisl/[~=11, include_prerelease=True]@oss/master")
```
yamingk, line 59: please ignore this for now, it will be removed once the PR from nuraft_mesg and sisl have been merged.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -184,16 +184,16 @@ void RaftReplDev::on_fetch_data_received(intrusive< sisl::GenericRpcData >& rpc_
...
-
         auto const& lsn = req->lsn();
-        auto const& term = req->raft_term();
-        auto const& dsn = req->dsn();
-        auto const& header = req->user_header();
-        auto const& key = req->user_key();
         auto const& originator = req->blkid_originator();
         auto const& remote_blkid = req->remote_blkid();
 
+        RD_LOG(INFO, "Data Channel: FetchData received: lsn={}", lsn);
```
sanebay, line 191: TRACE ?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -564,17 +625,14 @@ AsyncNotify RaftReplDev::notify_after_data_written(std::vector< repl_req_ptr_t >
...
     } else {
-        check_and_fetch_remote_data(rreqs);
         // some data are not in completed state, let's schedule a timer to check it again;
         // we wait for data channel to fill in the data. Still if its not done we trigger a fetch from remote;
-#if 0
         m_wait_data_timer_hdl = iomanager.schedule_global_timer( // timer wakes up in current thread;
             HS_DYNAMIC_CONFIG(consensus.wait_data_write_timer_sec) * 1000 * 1000 * 1000, false /* recurring */,
             nullptr /* cookie */, iomgr::reactor_regex::all_worker, [this, rreqs](auto /*cookie*/) {
-                LOGINFO("Data Channel: Wait data write timer fired, checking if data is written");
+                RD_LOG(INFO, "Data Channel: Wait data write timer fired, checking if data is written");
```
sanebay, line 633: nit: debug

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -416,25 +421,51 @@ void RaftReplDev::check_and_fetch_remote_data(std::vector< repl_req_ptr_t >* rre
...
+        }
+
+        // check if there is any left over not processed;
+        if (next_batch_rreqs.size()) { fetch_data_from_remote(std::move(next_batch_rreqs)); }
     }
 }
 
-void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t >* rreqs) {
+void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
+    if (rreqs.size() == 0) { return; }
```
sanebay, line 445: assert ?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -416,25 +421,51 @@ void RaftReplDev::check_and_fetch_remote_data(std::vector< repl_req_ptr_t >* rre
...
+
     std::vector<::flatbuffers::Offset< RequestEntry > > entries;
-    entries.reserve(rreqs->size());
+    entries.reserve(rreqs.size());
 
     shared< flatbuffers::FlatBufferBuilder > builder = std::make_shared< flatbuffers::FlatBufferBuilder >();
+    RD_LOG(INFO, "Data Channel : FetchData from remote: rreq.size={}, my server_id={}", rreqs.size(), server_id());
```
sanebay, line 451: nit: trace/debug

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -416,25 +421,51 @@ void RaftReplDev::check_and_fetch_remote_data(std::vector< repl_req_ptr_t >* rre
...
+    for (auto const& rreq : rreqs) {
         entries.push_back(CreateRequestEntry(*builder, rreq->get_lsn(), rreq->term(), rreq->dsn(),
                                              builder->CreateVector(rreq->header.cbytes(), rreq->header.size()),
                                              builder->CreateVector(rreq->key.cbytes(), rreq->key.size()),
                                              rreq->remote_blkid.server_id /* blkid_originator */,
                                              builder->CreateVector(rreq->remote_blkid.blkid.serialize().cbytes(),
                                                                    rreq->remote_blkid.blkid.serialized_size())));
-        LOGINFO("Fetching data from remote: rreq=[{}], remote_blkid={}", rreq->to_compact_string(),
-                rreq->remote_blkid.blkid.to_string());
+        // releax this assert if there is a case in same batch originator can be different (can't think of one now)
```
sanebay, line 461: wont a leader change cause this ? some of entries have different originators.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -416,25 +421,51 @@ void RaftReplDev::check_and_fetch_remote_data(std::vector< repl_req_ptr_t >* rre
...
+void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
+    if (rreqs.size() == 0) { return; }
+
     std::vector<::flatbuffers::Offset< RequestEntry > > entries;
-    entries.reserve(rreqs->size());
+    entries.reserve(rreqs.size());
 
     shared< flatbuffers::FlatBufferBuilder > builder = std::make_shared< flatbuffers::FlatBufferBuilder >();
+    RD_LOG(INFO, "Data Channel : FetchData from remote: rreq.size={}, my server_id={}", rreqs.size(), server_id());
+    auto const& originator = rreqs.front()->remote_blkid.server_id;
```
sanebay, line 453: always reading from leader might be easy to do. this way any error we could independently retry.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -444,17 +475,36 @@ void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t >* rreqs) {
...
-            RD_REL_ASSERT(!!e, "Error in fetching data");
+            if (!e) {
+                // if we are here, it means the original who sent the log entries are down.
+                // we need to handle error and when the other member becomes leader, it will resend the log entries;
+                RD_LOG(INFO,
+                       "Not able to fetching data from originator={}, error={}, probably originator is down. Will "
+                       "retry when new leader start appending log entries",
+                       rreqs.front()->remote_blkid.server_id, e.error());
+                for (auto const& rreq : rreqs) {
+                    handle_error(rreq, RaftReplService::to_repl_error(e.error()));
```
sanebay, line 491: not sure if originator down, how we switch to current leader.

src/tests/test_raft_repl_dev.cpp
```diff
@@ -331,45 +331,132 @@ TEST_F(RaftReplDevTest, All_Append_Fetch_Remote_Data) {
...
-    } else {
-        ASSERT_TRUE(leader_str != my_id_str) << "I am a follower, Leader id " << leader_str.c_str()
-                                             << " should not equals to my ID " << my_id_str.c_str();
+        LOGINFO("Wait for grpc connection to replica-1 to expire and removed from raft-groups.");
+        std::this_thread::sleep_for(std::chrono::seconds{5});
+
+        g_helper->runner().set_num_tasks(SISL_OPTIONS["num_io"].as< uint64_t >());
+
+        // before replica-1 started, issue I/O so that replica-1 is lagging behind;
+        auto block_size = SISL_OPTIONS["block_size"].as< uint32_t >();
```
sanebay, line 377: Wont replica1 restart by the time we wait for 5seconds causing replica1 never lagging behind. Will shutdown without remove files and start homestore of replica1 simulate that ?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -148,7 +148,7 @@ void RaftReplDev::push_data_to_all_followers(repl_req_ptr_t rreq) {
            flatbuffers::FlatBufferToString(builder.GetBufferPointer() + sizeof(flatbuffers::uoffset_t),
                                            PushDataRequestTypeTable()));*/
 
-    LOGINFO("Data Channel: Pushing data to all followers: rreq=[{}]", rreq->to_compact_string());
+    RD_LOG(INFO, "Data Channel: Pushing data to all followers: rreq=[{}]", rreq->to_compact_string());
```
hkadayam, line 151: Please change from INFO to DEBUG

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -184,16 +184,16 @@ void RaftReplDev::on_fetch_data_received(intrusive< sisl::GenericRpcData >& rpc_
     ctx->outstanding_read_cnt = fetch_req->request()->entries()->size();
```
hkadayam, line 184: I don't think we need to do Context and read count etc, which is reimplementing future/promise right? Suggest to create promise and move promise to callback and do setValue there. We don't need mutex, cv and atomic count explicltly.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -395,6 +390,16 @@ repl_req_ptr_t RaftReplDev::follower_create_req(repl_key const& rkey, sisl::blob
...
 
+auto RaftReplDev::get_max_data_fetch_size() const {
+#ifdef _PRERELEASE
+    if (iomgr_flip::instance()->test_flip("simulate_staging_fetch_data")) {
+        LOGINFO("Flip simulate_staging_fetch_data is enabled, return max_data_fetch_size: 16K");
+        return 4 * 4096ull;
+    }
+#endif
+    return HS_DYNAMIC_CONFIG(consensus.data_fetch_max_size_mb) * 1024 * 1024ull;
+}
```
hkadayam, line 401: Nit: Do we need it member function or local static should be enough?

src/lib/replication/repl_dev/raft_repl_dev.cpp
hkadayam: @sanebay If originator is down, those requests are timed out. But later when a new leader is elected, that leader will send the new data and will create a new rreq and fetch data and commit.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -444,17 +475,36 @@ void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t >* rreqs) {
...
+                }
+                return;
+            }
 
             auto raw_data = e.value().response_blob().cbytes();
             auto total_size = e.value().response_blob().size();
 
-            for (auto const& rreq : *rreqs) {
+            RD_DBG_ASSERT(total_size > 0, "Empty response from remote");
+            RD_DBG_ASSERT(raw_data, "Empty response from remote");
```
hkadayam, line 500: Suggest to use RD_DBG_ASSERT_GT(total_size, 0) to see output in the log.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -416,25 +421,51 @@ void RaftReplDev::check_and_fetch_remote_data(std::vector< repl_req_ptr_t >* rre
...
+            next_batch_rreqs.emplace_back(rreq);
+        }
+
+        // check if there is any left over not processed;
+        if (next_batch_rreqs.size()) { fetch_data_from_remote(std::move(next_batch_rreqs)); }
     }
 }
 
-void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t >* rreqs) {
+void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
```
hkadayam, line 444: Nit: We were passing earlier as a vector pointer instead of vector by value is because, down the line we capture this vector in collectAllUnSafe().thenValue() and we do expensive multiple allocation. You might argue that this function is not in critical path and don't mind this additional capture. My only thinking is for any reason we use this function in future, may be it might be helpful, but its an nit optional comment.

src/tests/test_raft_repl_dev.cpp
```diff
@@ -331,45 +331,132 @@ TEST_F(RaftReplDevTest, All_Append_Fetch_Remote_Data) {
...
+    // step-3: on leader, wait for a while for replica-1 to finish shutdown so that it can be removed from raft-groups
+    // and following I/O issued by leader won't be pushed to relica-1;
     if (g_helper->replica_num() == 0) {
-        ASSERT_TRUE(leader_str == my_id_str)
-            << "Leader id " << leader_str.c_str() << " should  equals to my ID " << my_id_str.c_str();
-    } else {
-        ASSERT_TRUE(leader_str != my_id_str) << "I am a follower, Leader id " << leader_str.c_str()
-                                             << " should not equals to my ID " << my_id_str.c_str();
+        LOGINFO("Wait for grpc connection to replica-1 to expire and removed from raft-groups.");
+        std::this_thread::sleep_for(std::chrono::seconds{5});
```
hkadayam, line 372: I feel instead of 5 seconds sleep, we can query the repl_dev to get the number of alive members as 2 and then proceed?

src/tests/test_raft_repl_dev.cpp
hkadayam: I think Sanal is right, that we might restart replica-1, which might have resynced by the time the sleep is completed and thus would be a regular append. If so, we can improve the restart method to simulate a sleep between shutdown and restart in (restart_homestore method), which we can control to make sure replica-1 is down while this happens.

src/lib/replication/repl_dev/raft_repl_dev.cpp
sanebay: If the originator is removed and new leader is elected and its sending resync, wont we keep timing out for each entry because we dont who the leader is. We wont be able to move forward. One option is to first for check for raft who is the leader and if not found, ask all replica's who the leader is.

src/tests/test_raft_repl_dev.cpp
yamingk: The shutdown actually takes some time to cmoplete, that is why we sleep 5 seconds here. 
It can be proved everytime replica-1 will be lagging with more than 64 append entries. 
Without this 5 seconds, it needs to do more than 10K I/Os which is fine but unnecessary and verification will also take time to increase the test case run time.

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: It is explained at line:485. The new leader will re-send the append entries with originator set to this new leader. We just need to handle failure here if the originator goes down.

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: will change to debug

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: done

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: I moved it to a member function because of the flip which can be a little easier for reader outside of this function.

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: it should be fine for someone to call it with no an empty vector.

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: this is a very unlikely code path and I tend to keep the info log. If we see this log a lot, it mean something is wrong in our system, although I don't think it should happen, but having this info log allows us keep an eye on it.

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: all entries will have same originator. after leader change, leader will resend append entries with originator set to its own id since the remote_blkids only meaningful to the new leader.

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: done

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: With all the logging being DEBUG, this is the only one showing fetch remote happened and I tend to keep this since it is not likely to hit in normal I/O path. We can move it to debug once matrix are there for resync.

src/tests/test_raft_repl_dev.cpp
yamingk: the issue I dealt with is replica-1 is already gone, but shutdown takes longer time (take cp, etc.) and if we start I/O immediately after repilca-1 is gone, the 300 I/O finishes too fast before replica-1 is back. I want to simulate resync while I/Os are still on-going.

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: We discussed this, somehow I was hitting hang while using the future/promise by collectAll on fetch side (receiving side is already calling collectAllUnsafe). Need to dig into why.

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: We can still do std::move on a vector to avoid allocation, right? 

It is a vector now because of the staging handling, and we can't not do start_indx, end_index with a pointer to vector as it won't work for std::move for the whole vector.

sanebay (dismissed):

src/tests/test_raft_repl_dev.cpp
xiaoxichen: There is a 10s sleep between shutdown and start in `start_homestore` so there is no possibility the replica1 started before starting IO(i.e  not possible never lagging behind).

But seems we want to simulate resync while IO are still going on,  this is harder as we are expecting the 300 IOs span longer than (10-5) = 5s.   Better we can continue the IO by triggering one more batch all the way till we see replica1 had joined back, and optionally do one more round of IO. then goes to `this->wait_for_all_writes(exp_entries);`

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -416,25 +421,51 @@ void RaftReplDev::check_and_fetch_remote_data(std::vector< repl_req_ptr_t >* rre
...
     if (rreqs->size()) {
         // Some data not completed yet, let's fetch from remote;
-        fetch_data_from_remote(rreqs);
+        auto total_size_to_fetch = 0ul;
+        std::vector< repl_req_ptr_t > next_batch_rreqs;
+        const auto max_batch_size = get_max_data_fetch_size();
+        for (auto const& rreq : *rreqs) {
+            auto const& size = rreq->remote_blkid.blkid.blk_count() * get_blk_size();
+            if ((total_size_to_fetch + size) >= max_batch_size) {
+                fetch_data_from_remote(std::move(next_batch_rreqs));
```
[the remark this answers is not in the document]
yamingk: Normally I tend to update size *after* certain operation has been completed, for the ease of error handling, etc. 

But you have a good point here, if a single rreq (not necessarily the 1st one)'s remote_blkid addresses as large as its maximum blk count, which is 65535, it could be as large as 64MB in nuboject case (1K blk size) and could be 256MB for a 4KB blk size, and we have to handle this case. This will go away if we use GRPC streaming API which we probably also need for baseline resync. Will mark as TODO for now.

src/tests/test_raft_repl_dev.cpp
```diff
@@ -331,45 +331,132 @@ TEST_F(RaftReplDevTest, All_Append_Fetch_Remote_Data) {
...
 
-    auto leader = repl_dev->get_leader_id();
-    ASSERT_TRUE(leader != replica_id_t())
-        << "Error getting leader id for group_id=" << boost::uuids::to_string(group_id).c_str();
-    auto leader_str = boost::uuids::to_string(leader);
-    LOGINFO("Got raft leader {} for group {}", leader_str, group_id);
+    // step-0: do some IO before restart one member;
+    uint64_t exp_entries = 20;
+    if (g_helper->replica_num() == 0) {
+        g_helper->runner().set_num_tasks(20);
```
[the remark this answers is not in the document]
yamingk: The data channel and log channel use the same grpc connection actually and if there is any network issue, a new leader should be elected. For the network flip, I also discussed this with Sanel the other day, that we probably should do retry, this will make the logic complicated but if this network flip (network failure, but still leader doesn't change) is a real thing, we don't have choice but to handle it. I will mark it as a TODO also (to avoid comlicating things before we are 100% sure needed).

src/tests/test_raft_repl_dev.cpp
yamingk: I tested 3000 IOs, and we can actually do that by default, but I thought we don't want to increase the time for homestore build for everyone. I think we better do it in long duration test: https://github.com/eBay/HomeStore/issues/319

sanebay (approved):

xiaoxichen (approved):

PR #318, opened 2024-02-10
ThinProvision:  skip all zero buffers

shosseinimotlagh:


src/homeblks/volume/tests/vol_gtest.cpp
```diff
@@ -1673,9 +1680,11 @@ class IOTestJob : public TestJob {
         const uint32_t max_blks{
             static_cast< uint32_t >(tcfg.max_io_size / VolInterface::get_instance()->get_page_size(vinfo->vol))};
         // lba: [0, max_vol_blks - max_blks)
-        std::uniform_int_distribution< uint64_t > lba_random{0, vinfo->max_vol_blks - max_blks - 1};
+        //        std::uniform_int_distribution< uint64_t > lba_random{0, vinfo->max_vol_blks - max_blks - 1};
+        std::uniform_int_distribution< uint64_t > lba_random{0, 100};
```
shosseinimotlagh, line 1684: revert it back later

src/homeblks/volume/tests/vol_gtest.cpp
```diff
@@ -1673,9 +1680,11 @@ class IOTestJob : public TestJob {
...
         // nlbas: [1, max_blks]
-        std::uniform_int_distribution< uint32_t > nlbas_random{1, max_blks};
+        //        std::uniform_int_distribution< uint32_t > nlbas_random{1, max_blks};
+        std::uniform_int_distribution< uint32_t > nlbas_random{1, 100};
```
shosseinimotlagh, line 1687: revert back later

src/homeblks/volume/volume.cpp
```diff
@@ -334,8 +334,119 @@ indx_tbl* Volume::recover_indx_tbl(btree_super_block& sb, btree_cp_sb& cp_info)
...
+            }
+        }
+    }
+
+    if (start != -1) { result.emplace_back(start, empty_blocks.size() - start); }
+
+    return result;
+}
+
+static std::vector< std::pair< int, int > > compute_range_intervals(const uint8_t* buf, size_t page_size,
```
shosseinimotlagh, line 369: for later extension on subranges

src/.clang-format
```diff
@@ -18,7 +18,6 @@ AlignOperands:   false
 AlignTrailingComments: true
 AllowShortBlocksOnASingleLine: true
 AllowShortIfStatementsOnASingleLine: true
-AllowShortBlocksOnASingleLine: true
```
shosseinimotlagh, line 21: verbose (line 19)

raakella1 (changes requested):
Please open the PR against thin prov branch.

src/homeblks/volume/volume.cpp
```diff
@@ -334,8 +334,119 @@ indx_tbl* Volume::recover_indx_tbl(btree_super_block& sb, btree_cp_sb& cp_info)
                            SnapMgr::add_read_tracker, &cp_info);
     return static_cast< indx_tbl* >(tbl);
 }
+static std::vector< bool > find_non_zero_data(const uint8_t* buf, size_t size, uint32_t nlbas) {
+    std::vector< bool > empty_blocks;
+
+    auto is_buf_empty = [](const uint8_t* buf, size_t size) -> bool {
+        return buf[0] == 0 && !std::memcmp(buf, buf + 1, size - 1);
```
yamingk, line 341: as we discussed, you are going to change it with the intel crc algorithm to detect zero then do a bitwise compare, right?

shosseinimotlagh: affirmative. I will make this change and will update this pr

src/engine/common/homestore_config.fbs
```diff
@@ -142,6 +142,8 @@ table Generic {
     // percentage of cache used to create indx mempool. It should be more than 100 to 
     // take into account some floating buffers in writeback cache.
     indx_mempool_percent : uint32 = 110;
+
+    boot_thin_provisioning: bool = false;
```
yamingk, line 146: What is the deployment plan to enable this feature in production if this is defaulted to false?

src/homeblks/volume/volume.cpp
```diff
@@ -335,7 +335,105 @@ indx_tbl* Volume::recover_indx_tbl(btree_super_block& sb, btree_cp_sb& cp_info)
     return static_cast< indx_tbl* >(tbl);
 }
 
+static std::vector< std::pair< int, int > > get_true_intervals(const std::vector< bool >& empty_blocks) {
```
yamingk, line 338: I don't see any place using this api as well as copmute_range_intervals, can we remove them unless there is a plan to use them in the near future?

src/homeblks/volume/volume.cpp
```diff
@@ -335,7 +335,105 @@ indx_tbl* Volume::recover_indx_tbl(btree_super_block& sb, btree_cp_sb& cp_info)
...
+    return true;
+}
+static bool is_zero_request(const vol_interface_req_ptr& iface_req, uint32_t page_size) {
+    if (iface_req->iovecs.empty()) {
+        return is_buf_zero(static_cast< uint8_t* >(iface_req->buffer), iface_req->nlbas * page_size);
+    }
+    return is_iovec_zero(iface_req->iovecs);
+}
+
+static std::string print_ranges(lba_t start_lba, const std::vector< std::pair< int, int > >& intervals) {
```
yamingk, line 404: this one also can be removed if not going to be used

src/homeblks/volume/tests/vol_gtest.cpp
```diff
@@ -2758,6 +2795,9 @@ int main(int argc, char* argv[]) {
     gcfg.app_mem_size_in_gb = SISL_OPTIONS["app_mem_size_in_gb"].as< uint32_t >();
     gcfg.vol_copy_file_path = SISL_OPTIONS["vol_copy_file_path"].as< std::string >();
     const auto io_size_in_kb = SISL_OPTIONS["io_size"].as< uint32_t >();
+    gcfg.p_zero_buffer = SISL_OPTIONS["p_zero_buffer"].as< uint32_t >();
```
yamingk, line 2798: as we discussed, would suggest to add a basic zero write/read test case:
1. populate real data from LBA [0, 2MB]
2. write all zeros to [128K, 128K + 32K]
3. read LBA from [128K, 128K + 32K], verify all zeros are returned.

Another basic test:
Similar as above one, only skip the 1st step, e.g. write zeros to LBAs that has never been written before. 

This is to verify that new writes and/or overwrite with all zeros both works.

src/homeblks/volume/volume.cpp
yamingk: Can you put these three APIs that are currently not used together and make an #if 0 around them, so that it would be easier for others to read and understand this thin-provisioning code? Also put comment that these apis are for future usage if we support paritial zero detection.

shosseinimotlagh: sure

PR #323, opened 2024-02-14
Use new testing option.

szmyd:


raakella1:
Opening a different PR with compile errors fixed

PR #330, opened 2024-02-20
Possible fix for refresh node lock problem

shosseinimotlagh:
Test failure happens when a fiber is slow. This triggers `idx_node->m_last_mod_cp_id > cp_ctx->id()`, hence returns put/remove returns `btree_status_t::cp_mismatch`. One possible solution is that the caller performs retries. Having said that, in the UT, if it `cp_mismatch` happens, the caller retries until it happens.
###  Testing
In order to hit the problem, the number of fibers increases to 100 per thread and 10 threads (1000 fibers in total)
`bin/test_index_btree --gtest_filter=*/0.ConcurrentAllOps --preload_size=400000 --num_entries=800000 --num_iters=50000 --gtest_break_on_failure --num_fibers=100  --num_threads=10`

### DIFF
--- a/src/include/homestore/index/index_table.hpp
+++ b/src/include/homestore/index/index_table.hpp
@@ -75,16 +75,30 @@ public:
 
     template < typename ReqT >
     btree_status_t put(ReqT& put_req) {
-        auto cpg = hs()->cp_mgr().cp_guard();
-        put_req.m_op_context = (void*)cpg.context(cp_consumer_t::INDEX_SVC);
-        return Btree< K, V >::put(put_req);
+        auto ret = btree_status_t::success;
+        do {
+            auto cpg = hs()->cp_mgr().cp_guard();
+            put_req.m_op_context = (void*)cpg.context(cp_consumer_t::INDEX_SVC);
+            ret = Btree< K, V >::put(put_req);
+            if (ret == btree_status_t::cp_mismatch) {
+                LOGTRACEMOD(wbcache, "CP Mismatch, retrying put");
+            }
+        } while (ret == btree_status_t::cp_mismatch);
+        return ret;
     }
 
     template < typename ReqT >
     btree_status_t remove(ReqT& remove_req) {
-        auto cpg = hs()->cp_mgr().cp_guard();
-        remove_req.m_op_context = (void*)cpg.context(cp_consumer_t::INDEX_SVC);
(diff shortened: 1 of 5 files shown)


src/tests/btree_helpers/btree_test_helper.hpp
```diff
@@ -92,15 +92,44 @@ struct BtreeTestHelper {
...
-                    for (uint32_t i = start_range; i < end_range; i++) {
-                        put(i, btree_put_type::INSERT);
-                    }
-                    {
-                        std::unique_lock lg(m_test_done_mtx);
-                        if (--test_count == 0) { m_test_done_cv.notify_one(); }
-                    }
-                });
+                auto fiber_id = i;
+                iomanager.run_on_forget(
```
sanebay, line 96: Could you please move/copy this printing the progress to a function if possible and use it on concurrent test ops too. Today we dont have any visibility of whether its stuck or in progress.

src/include/homestore/index/index_table.hpp
```diff
@@ -75,16 +75,24 @@ class IndexTable : public IndexTableBase, public Btree< K, V > {
...
     btree_status_t put(ReqT& put_req) {
-        auto cpg = hs()->cp_mgr().cp_guard();
-        put_req.m_op_context = (void*)cpg.context(cp_consumer_t::INDEX_SVC);
-        return Btree< K, V >::put(put_req);
+        auto ret = btree_status_t::success;
+        do {
+            auto cpg = hs()->cp_mgr().cp_guard();
+            put_req.m_op_context = (void*)cpg.context(cp_consumer_t::INDEX_SVC);
+            ret = Btree< K, V >::put(put_req);
+        } while (ret == btree_status_t::cp_mismatch);
```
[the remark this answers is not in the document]
hkadayam: @sanebay I don't think we need to wait for timeout before trying cp_mismatch. It is just that entering CP is higher than whats modified. I think we should retry right away.

src/tests/btree_helpers/btree_test_helper.hpp
shosseinimotlagh: both of them have progress printing but they are slightly different (in terms of printing) because preload doesn't have deadline (time to stop).

sanebay (dismissed):

sanebay (approved):

PR #333, opened 2024-02-22
fix restartability bug for upper layer

JacksonYao287:
if it is not first-time-boot, the uuid of HO is saved in metaservice. so we need to get the uuid after metaservice is started, otherwise we will get a new default uuid

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -5,7 +5,7 @@ from conans import CMake
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "5.1.6"
+    version = "5.1.7"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/lib/homestore.cpp
+++ b/src/lib/homestore.cpp
@@ -126,6 +126,7 @@ bool HomeStore::start(const hs_input_params& input, hs_before_services_starting_
     if (has_repl_data_service()) {
         m_log_service = std::make_unique< LogStoreService >();
         m_data_service = std::make_unique< BlkDataService >(std::move(s_custom_chunk_selector));
+        m_repl_service = GenericReplService::create(std::move(s_repl_app));
     } else {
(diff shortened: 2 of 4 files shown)


xiaoxichen:
This suggests we are wrong in https://github.com/eBay/HomeStore/pull/284/files
for restart,  the `_our_id` in HO is set by `homestore_metablk_callback`,  which will be called after meta svc started in https://github.com/eBay/HomeStore/blob/master/src/lib/homestore.cpp#L210

In the #284 we move the initialization of  GenericReplService from `start` to `do_start`.

I think we either
a.  move https://github.com/eBay/HomeStore/blob/master/src/lib/homestore.cpp#L198 down below L210
or 
b. move L198 back to start() , remove wrong comments and  take the change in this PR to get uuid in GenericReplService::start() in L217

xiaoxichen (dismissed):
check back with #284

JacksonYao287:
@xiaoxichen 
for a, we can not make this change,  since the `register_handler` happens in the constructor of `RaftReplService`
https://github.com/eBay/HomeStore/blob/67cca17a715f08975ab7f28f211cdfe34963fbeb/src/lib/replication/service/raft_repl_service.cpp#L64
if we put L198 down below L210, then the handler can not be registered before metaservice#start, which will lead to a missing raft config.

well, b is acceptable, will do this!

xiaoxichen (approved):
lgtm

Though we had seen an initialized uuid default value for `m_repl_app->get_my_repl_id()` and do not know the reason, but maybe 
`        RELEASE_ASSERT(!m_my_uuid.value().is_nil(), "Uninitialized RaftReplService UUID");`

JacksonYao287:
> lgtm
> 
> Though we had seen an initialized uuid default value for `m_repl_app->get_my_repl_id()` and do not know the reason, but maybe ` RELEASE_ASSERT(!m_my_uuid.value().is_nil(), "Uninitialized RaftReplService UUID");`

ok, will add it in the later coming PR

PR #334, opened 2024-02-23
Fast forward API to truncate upto a log index

raakella1:


src/homelogstore/log_store.cpp
```diff
@@ -237,13 +237,54 @@ void HomeLogStore::truncate(const logstore_seq_num_t upto_seq_num, const bool in
     if (locked_now) { m_logdev.unlock_flush(); }
 }
 
+void HomeLogStore::truncate_sync(const logstore_seq_num_t upto_seq_num, const bool in_memory_truncate_only) {
```
hkadayam, line 240: I would rather do fast_forward api which does fill_gap and then call truncate this. Also we don't need in_memory_truncate_only for this use case.

src/homelogstore/log_store.cpp
```diff
@@ -237,13 +237,62 @@ void HomeLogStore::truncate(const logstore_seq_num_t upto_seq_num, const bool in
...
+    }
+}
+
 // NOTE: This method assumes the flush lock is already acquired by the caller
-void HomeLogStore::do_truncate(const logstore_seq_num_t upto_seq_num) {
+void HomeLogStore::do_truncate(const logstore_seq_num_t upto_seq_num, const bool persist_now) {
     m_records.truncate(upto_seq_num);
     m_safe_truncation_boundary.seq_num.store(upto_seq_num, std::memory_order_release);
 
     // Need to update the superblock with meta, we don't persist yet, will be done as part of log dev truncation
```
hkadayam, line 294: nit: suggest an statement 
// Need to update the superblock with meta, we don't persist yet, will be done as part of log dev truncation
to
// Need to update the superblock with meta, in case we don't persist yet, will be done as part of log dev truncation

hkadayam (approved):

PR #336, opened 2024-02-23
Create repl req details for all IO types.

hkadayam:
Following significant changes are made in this commit

* Till now repl req is not created for header_only (inline) request, which results in different handling of request nullptr everywhere resulted in some crash. Made this uniform behavior on every types.

* At present, we use replica set is_leader() to make decision on some commit, append entry code path. This could be potentially dangerous as at some inopprtunite time, if leader is switched and somehow the request survived, we might run into corruption cases. With this change, there is no special handling of leader. Only 2 roles, proposer and applier. In most use cases proposer is leader and applier is the follower, but that requirement is removed.

* Because of the above assumption, there was a bug where in leader non-app entries (conf entries) are not guaranteed to be flushed, before calling commit. Fixed that behavior as well.

* The HomeLogStore has an limitation (exposed when the above issue is fixed), whereby in proposer when the data write is completed and then propose to raft happens. RAFT calls end_of_append_batch in the same thread which does flush_sync. However, the data_write completion happens in IO thread and thus there is a potential for deadlock (since flush_sync takes giant lock). We can overcome this problem with fiber, but instead of forcing worker thread to have fiber, we isolate all logstore flush to be in separate flush thread always.

### DIFF
--- a/src/lib/logstore/log_dev.cpp
+++ b/src/lib/logstore/log_dev.cpp
@@ -87,13 +87,11 @@ void LogDev::start(bool format, JournalVirtualDev* vdev) {
         m_last_flush_idx = m_log_idx - 1;
     }
 
-    m_flush_timer_hdl = iomanager.schedule_global_timer(
-        HS_DYNAMIC_CONFIG(logstore.flush_timer_frequency_us) * 1000, true, nullptr /* cookie */,
-        iomgr::reactor_regex::all_worker,
-        [this](void*) {
-            if (m_pending_flush_size.load() && !m_is_flushing.load(std::memory_order_relaxed)) { flush_if_needed(); }
-        },
-        true /* wait_to_schedule */);
+    iomanager.run_on_wait(logstore_service().flush_thread(), [this]() {
+        m_flush_timer_hdl = iomanager.schedule_thread_timer(HS_DYNAMIC_CONFIG(logstore.flush_timer_frequency_us) * 1000,
+                                                            true /* recurring */, nullptr /* cookie */,
+                                                            [this](void*) { flush_if_needed(); });
+    });
 
     handle_unopened_log_stores(format);
 
@@ -133,7 +131,8 @@ void LogDev::stop() {
     }
 
     // cancel the timer
-    iomanager.cancel_timer(m_flush_timer_hdl, true);
+    iomanager.run_on_wait(logstore_service().flush_thread(),
+                          [this]() { iomanager.cancel_timer(m_flush_timer_hdl, true); });
 
     {
         folly::SharedMutexWritePriority::WriteHolder holder(m_store_map_mtx);
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -95,11 +95,15 @@ void RaftReplDev::async_alloc_write(sisl::blob const& header, sisl::blob const& 
     rreq->header = header;
     rreq->key = key;
     rreq->value = value;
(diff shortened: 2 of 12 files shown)


src/lib/logstore/log_dev.cpp
```diff
@@ -87,13 +87,11 @@ void LogDev::start(bool format, JournalVirtualDev* vdev) {
...
-        iomgr::reactor_regex::all_worker,
-        [this](void*) {
-            if (m_pending_flush_size.load() && !m_is_flushing.load(std::memory_order_relaxed)) { flush_if_needed(); }
-        },
-        true /* wait_to_schedule */);
+    iomanager.run_on_wait(logstore_service().flush_thread(), [this]() {
+        m_flush_timer_hdl = iomanager.schedule_thread_timer(HS_DYNAMIC_CONFIG(logstore.flush_timer_frequency_us) * 1000,
+                                                            true /* recurring */, nullptr /* cookie */,
+                                                            [this](void*) { flush_if_needed(); });
+    });
```
hkadayam, line 94: Set the flush thread timer only in the logstore flush thread and not global timer.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -599,22 +615,21 @@ AsyncNotify RaftReplDev::notify_after_data_written(std::vector< repl_req_ptr_t >
...
-                                    rreq->to_compact_string());
-                             return true; // Remove from the pending list
-                         } else {
-                             futs.emplace_back(rreq->data_written_promise.getSemiFuture());
-                             return false;
-                         }
-                     }),
+    rreqs->erase(std::remove_if(rreqs->begin(), rreqs->end(),
+                                [this, &futs](repl_req_ptr_t const& rreq) {
+                                    if ((rreq == nullptr) || (rreq->value_inlined)) { return true; }
```
hkadayam, line 620: Only change is this line 620, nothing else is changed, but this resulted in format change.

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -16,55 +16,22 @@ RaftStateMachine::RaftStateMachine(RaftReplDev& rd) : m_rd{rd} {
...
-    }
-    if (rreq->state.load() & uint32_cast(repl_req_state_t::DATA_WRITTEN)) {
-        m_lsn_req_map.erase(rreq->lsn);
-        m_rd.report_committed(rreq);
-    }
-    return m_success_ptr;
-}
-
-uint64_t RaftStateMachine::last_commit_index() { return uint64_cast(m_rd.get_last_commit_lsn()); }
-
```
hkadayam, line 62: Moved this precommit_ext and commit_ext to have linear flow in the source code, we first write into journal (and thus transform), then pre_commit and then commit.

src/tests/test_common/hs_repl_test_common.hpp
```diff
@@ -153,8 +153,13 @@ class HSReplTestHelper {
...
                 LOGINFO("Spawning Homestore replica={} instance", i);
-                boost::process::child c(argv_[0], "--log_mods", "replication:trace", "--replica_num", std::to_string(i),
-                                        proc_grp_);
+                std::string cmd_line;
+                fmt::format_to(std::back_inserter(cmd_line), "{} --replica_num {}", argv_[0], i);
+                for (int j{1}; j < argc_; ++j) {
+                    fmt::format_to(std::back_inserter(cmd_line), " {}", argv_[j]);
+                }
+                boost::process::child c(boost::process::cmd = cmd_line, proc_grp_);
+                // boost::process::child c(argv_[0], "--replica_num", std::to_string(i), proc_grp_);
```
hkadayam, line 162: With this, the main cmdline parameters are passed into replica process as well.

PR #337, opened 2024-02-28
issue 329: RafReplDev Metrics and Restart Leader Test

### LINKED ISSUE #329: Homestore Raft Repl Dev metrics | Leader Restart Test Case Implement
Some pointers on the metrics to be defined:

Homestore Per repl dev metrics
Error metrics, resync time. 
Replication lag 
Put, get latencies

Recovery time for each subsystem

yamingk:
1. add some metrics around read/write/fetch for error cases. 
2. add restart leader test case. 

Testing:
=======
1. Wrote a script to run restart leader for 100 times and passed. Verified leader changed to replica-1 after replica-0 (original leader) restarted with I/O running.

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -28,7 +28,8 @@ RaftReplDev::RaftReplDev(RaftReplService& svc, superblk< raft_repl_dev_superblk 
         m_group_id{rd_sb->group_id},
         m_my_repl_id{svc.get_my_repl_uuid()},
         m_raft_server_id{nuraft_mesg::to_server_id(m_my_repl_id)},
-        m_rd_sb{std::move(rd_sb)} {
+        m_rd_sb{std::move(rd_sb)},
+        m_metrics{fmt::format("{}_{}", group_id_str(), m_raft_server_id).c_str()} {
     m_state_machine = std::make_shared< RaftStateMachine >(*this);
 
     if (load_existing) {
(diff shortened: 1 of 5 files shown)


raakella1 (dismissed):
LG!

xiaoxichen (dismissed):

src/tests/test_raft_repl_dev.cpp
```diff
@@ -461,6 +466,65 @@ TEST_F(RaftReplDevTest, All_restart_one_follower_inc_resync_with_staging) {
...
+        LOGINFO("Run on worker threads to schedule append on repldev for {} Bytes.", block_size);
+        g_helper->runner().set_task([this, block_size]() {
+            this->generate_writes(block_size /* data_size */, block_size /* max_size_per_iov */);
+        });
+        g_helper->runner().execute().get();
+    }
+
+    this->wait_for_all_writes(exp_entries);
+
+    if (g_helper->replica_num() != 0) { std::this_thread::sleep_for(std::chrono::seconds{10}); }
```
yamingk, line 520: this line actually should not be needed since we have sync_for_verify_start(). I will remove it in my other PR.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -487,6 +493,8 @@ void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
     builder->FinishSizePrefixed(
         CreateFetchData(*builder, CreateFetchDataRequest(*builder, builder->CreateVector(entries))));
 
+    COUNTER_INCREMENT(m_metrics, fetch_rreq_cnt, 1);
```
hkadayam, line 496: would suggest add a metrics on 2 more things:
1. Number of entries we are requesting
2. Size of total blocks requested

yamingk: Thought about these two things, didn't do that considering the space. 

These two things will be histogram which could take relatively larger space, we can have them if we are fine.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -512,6 +522,9 @@ void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
             auto raw_data = e.value().response_blob().cbytes();
             auto total_size = e.value().response_blob().size();
 
+            COUNTER_INCREMENT(m_metrics, fetch_rreq_cnt, 1);
```
hkadayam, line 525: Aren't we double incrementing the fetch_rreq_cnt, one while requesting and other after response?

raakella1 (approved):
LG!

PR #343, opened 2024-03-01
Ability to ssh to github actions runnner.

sanebay:
ssh via tmate will close the VM, if UT's fail. so put a sleep instead and build manually.

### DIFF
--- a/.github/workflows/build_dependencies.yml
+++ b/.github/workflows/build_dependencies.yml
@@ -83,6 +83,7 @@ on:
 jobs:
   BuildHomestoreDeps:
     runs-on: ${{ inputs.platform }}
+    timeout-minutes: 1440
     steps:
     - name: Retrieve Code
       uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/build_dependencies.yml
```diff
@@ -83,6 +83,7 @@ on:
 jobs:
   BuildHomestoreDeps:
     runs-on: ${{ inputs.platform }}
+    timeout-minutes: 1440
```
sanebay, line 86: From  net, to add this to keep the VM for longer time.

.github/workflows/build_dependencies.yml
```diff
@@ -206,10 +207,18 @@ jobs:
...
+      uses: mxschmitt/action-tmate@v3
+      with:
+        limit-access-to-actor: true
+        detached: true
+
     - name: Create and Test Package
       run: |
         sanitize=$([[ "${{ inputs.tooling }}" == "Sanitize" ]] && echo "True" || echo "False")
         pre=$([[ "${{ inputs.build-type }}" != "Debug" ]] && echo "-o sisl:prerelease=${{ inputs.prerelease }}" || echo "")
+        sudo rm -rf /usr/local/lib/android/
```
sanebay, line 221: For sanitize build, VM is too low on space. By removing this unwanted folder, we get some 10G extra.

szmyd (approved):
makes sense; how is it used?

sanebay:
> makes sense; how is it used?
In actions, there is a tab for Setup tmate session which gives details about the ssh connection.

image

PR #347, opened 2024-03-08
Remove std::move() in async_alloc_write.

xiaoxichen:
Previously we do move() for rreq which
1. has no effect on write_with_data path as rreq is captured by value which is a const.
2. make rreq (intrusive_ptr) to NULL after move and causing panic in handle_error(rreq) later.

Removing the move() to fix the bug, though with a cost of extra atomic_inc/dec

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -121,7 +121,7 @@ void RaftReplDev::async_alloc_write(sisl::blob const& header, sisl::blob const& 
         // Write the data
         data_service().async_write(rreq->value, rreq->local_blkid).thenValue([this, rreq](auto&& err) {
             if (!err) {
-                auto raft_status = m_state_machine->propose_to_raft(std::move(rreq));
+                auto raft_status = m_state_machine->propose_to_raft(rreq);
                 if (raft_status != ReplServiceError::OK) { handle_error(rreq, raft_status); }
             } else {
                 HS_DBG_ASSERT(false, "Error in writing data");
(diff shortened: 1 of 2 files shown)


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -121,7 +121,7 @@ void RaftReplDev::async_alloc_write(sisl::blob const& header, sisl::blob const&
         // Write the data
         data_service().async_write(rreq->value, rreq->local_blkid).thenValue([this, rreq](auto&& err) {
             if (!err) {
-                auto raft_status = m_state_machine->propose_to_raft(std::move(rreq));
```
xiaoxichen, line 124: the rreq is a const so actually the move is not happening

hkadayam (approved):

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -121,7 +121,7 @@ void RaftReplDev::async_alloc_write(sisl::blob const& header, sisl::blob const&
...
+                auto raft_status = m_state_machine->propose_to_raft(rreq);
```
yamingk, line 124: to overcome the inc/dec, is there a reason why this lambda can't be declared as `mutable` so that rreq can be non-const?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -131,7 +131,7 @@ void RaftReplDev::async_alloc_write(sisl::blob const& header, sisl::blob const&
     } else {
         rreq->value_inlined = true;
         RD_LOG(DEBUG, "Skipping data channel send since value size is 0");
-        auto raft_status = m_state_machine->propose_to_raft(std::move(rreq));
+        auto raft_status = m_state_machine->propose_to_raft(rreq);
```
yamingk, line 134: I probably didn't get the reason for this change, since rreq is not const here, right?

xiaoxichen: yes so that will make rreq into NULL and `handle_error` will panic when accessing this.

yamingk: ahh..., right.

PR #355, opened 2024-03-19
Allowing consumer of HS specified device type for each service

xiaoxichen:
This is the change we need for production ,  for more background we recorded here https://github.com/eBay/HomeObject/issues/110

Consumer setting the device type to auto will let HS choose based on what it has.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -5,7 +5,7 @@ from conans import CMake
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.1.1"
+    version = "6.2.1"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/conanfile.py
+++ b/conanfile.py
@@ -5,7 +5,7 @@ from conans import CMake
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.0.1"
+    version = "6.1.1"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/include/homestore/homestore_decl.hpp
+++ b/src/include/homestore/homestore_decl.hpp
@@ -87,7 +87,7 @@ constexpr uint16_t MAX_UUID_LEN{128};
 static constexpr hs_uuid_t INVALID_SYSTEM_UUID{0};
 
 ///////////// All Enums //////////////////////////
-ENUM(HSDevType, uint8_t, Data, Fast);
+ENUM(HSDevType, uint8_t, Auto, Data, Fast);
 ENUM(Op_type, uint8_t, READ, WRITE, UNMAP);
 VENUM(PhysicalDevGroup, uint8_t, DATA = 0, FAST = 1, META = 2);
 ENUM(io_flag, uint8_t,
@@ -148,7 +148,8 @@ static std::string in_bytes(uint64_t sz) {
 }
 
 struct hs_format_params {
-    float size_pct;
+    HSDevType dev_type{HSDevType::Data};
+    float size_pct; // size pct to that type
     uint32_t num_chunks{1};
     uint64_t chunk_size{0};
     uint32_t block_size{0};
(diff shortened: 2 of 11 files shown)


src/lib/homestore.cpp
```diff
@@ -365,7 +425,7 @@ shared< VirtualDev > HomeStore::create_vdev_cb(const vdev_info& vinfo, bool load
 }
 
 uint64_t HomeStore::pct_to_size(float pct, HSDevType dev_type) const {
-    uint64_t sz = uint64_cast((pct * static_cast< double >(m_dev_mgr->total_capacity())) / 100);
+    uint64_t sz = uint64_cast((pct * static_cast< double >(m_dev_mgr->total_capacity(dev_type))) / 100);
```
xiaoxichen, line 428: This is the most significant behavior change in the case of hybrid storage. 

Previously we calculate the size based on PCT * (HDD+SSD) ,  now it is PCT * size_of_target_device_type

conanfile.py
```diff
@@ -5,7 +5,7 @@
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.0.1"
+    version = "6.1.1"
```
szmyd, line 8: Technically a major (non-backwards compat) change. So if not going to bump Major, patch is good enough?

src/include/homestore/homestore_decl.hpp
```diff
@@ -87,7 +87,7 @@ constexpr uint16_t MAX_UUID_LEN{128};
 static constexpr hs_uuid_t INVALID_SYSTEM_UUID{0};
 
 ///////////// All Enums //////////////////////////
-ENUM(HSDevType, uint8_t, Data, Fast);
+ENUM(HSDevType, uint8_t, Auto, Data, Fast);
```
yamingk, line 90: I am not clear on how HomeObj will specify the percentage for each service with this Auto devType in place.
How can HomeObj know whether it is mixed dev type, so that META+LOG+INDEX can accumulate upto 100% of the SSD drive, and the replication could consume 100% of the other devType drives, and if it is not mixed, say all HDD, or all NVMe, then META+LOG+INDEX+REPLICATION should accumulate upto 100%. 

Can you share how is HomeObject going to use this?

src/lib/homestore.cpp
```diff
@@ -147,37 +151,93 @@ bool HomeStore::start(const hs_input_params& input, hs_before_services_starting_
...
+            }
+        } else {
+            return HSDevType::Data;
+        }
+    };
+    std::map< HSDevType, float > total_pct_by_type = {{HSDevType::Fast, 0.0f}, {HSDevType::Data, 0.0f}};
+    // applying optimal placement for auto
+    for (auto& [svc_type, fparams] : format_opts) {
+        if ((svc_type & HS_SERVICE::META) && has_meta_service()) {
+            if (fparams.dev_type == HSDevType::Auto) { fparams.dev_type = get_optimal_dev_type(svc_type); }
```
yamingk, line 171: To simplify the logic a bit, I am thinking can we relax this Auto check, and call get_optimal_dev_type directly for every services, the logic is consumer can specify FAST, but if fast is not available, we fall back to DATA. 

This is the logic we apply for 1.3 HomeStore also. Does it make sense?

yamingk:
LGTM, can you also run `./bin/test_raft_repl_dev` on your local machine (i don't think this PR will break anything) which currently doesn't run in conan build as well as github CI, just to make sure everything is right.

yamingk (approved):

xiaoxichen:
Merge as test_raft_repl_dev passsed

```
[       OK ] RaftReplDevTest.Leader_Restart (28386 ms)
[       OK ] RaftReplDevTest.Leader_Restart (28386 ms)
[       OK ] RaftReplDevTest.Leader_Restart (28386 ms)
[----------] 6 tests from RaftReplDevTest (115097 ms total)

[----------] 6 tests from RaftReplDevTest (115090 ms total)

[----------] 6 tests from RaftReplDevTest (115278 ms total)

[----------] Global test environment tear-down
[----------] Global test environment tear-down
[----------] Global test environment tear-down
[==========] 6 tests from 1 test suite ran. (115097 ms total)
[==========] 6 tests from 1 test suite ran. (115090 ms total)
[==========] 6 tests from 1 test suite ran. (115278 ms total)
[  PASSED  ] 6 tests.
[  PASSED  ] 6 tests.
[  PASSED  ] 6 tests.
```

PR #358, opened 2024-03-20
Add multiple chunk test for journal vdev.

sanebay:


src/tests/test_journal_vdev.cpp
```diff
@@ -470,6 +503,112 @@ TEST_F(VDevJournalIOTest, Recovery) {
...
+    auto log_dev_jd = test.vdev_jd();
+
+    auto restart_restore = [&]() {
+        test.save();
+        restart_homestore();
+        test.restore();
+        log_dev_jd = test.vdev_jd();
+    };
+
+#define verify_jd(...)                                                                                                 \
```
yamingk, line 521: Is there any specific reason it can not be a static function?

sanebay: To keep all the struct members in one line which is easy to read. clang-format splits to multiple lines.

yamingk:
Looks like logstore ut is failing in CI.

src/tests/test_journal_vdev.cpp
yamingk: c++20 allows that too, right?

sanebay: if we put a function {.a=10,.b=20} , clang-format will split into multiple  5 or 6 lines. Its too fool compiler to keep it one or two line.

yamingk: If clang-format is the concern, we can use below to instruct clang-format not to do that.
// clang-format off
...
// clang-format on

sanebay: Done.

src/lib/device/journal_vdev.cpp
```diff
@@ -472,11 +472,19 @@ off_t JournalVirtualDev::Descriptor::dev_offset(off_t nbytes) const {
...
-    RELEASE_ASSERT_EQ(m_end_offset - data_start_offset_aligned, m_total_size, "offset size mismatch {}", to_string());
+    if (!m_journal_chunks.empty()) {
+        m_data_start_offset = offset;
+        auto data_start_offset_aligned = sisl::round_down(m_data_start_offset, m_vdev.info().chunk_size);
+        m_end_offset = data_start_offset_aligned + m_journal_chunks.size() * m_vdev.info().chunk_size;
+        LOGINFOMOD(journalvdev, "Updated data start offset off 0x{} {}", to_hex(offset), to_string());
+        RELEASE_ASSERT_EQ(m_end_offset - data_start_offset_aligned, m_total_size, "offset size mismatch {}",
+                          to_string());
+    } else {
+        // If there are no chunks, we round up to the next chunk size.
```
yamingk, line 483: Can you explain a bit when this `else` branch will be hit?

src/lib/device/journal_vdev.hpp
```diff
@@ -63,7 +63,7 @@ class JournalVirtualDev : public VirtualDev {
 
         off_t m_data_start_offset{0};                   // Start offset of where actual data begin for this vdev
         std::atomic< uint64_t > m_write_sz_in_total{0}; //  Size will be decreased by truncate and increased by append;
-        bool m_truncate_done{true};
+        bool m_truncate_done{false};
         uint64_t m_reserved_sz{0};                       // write size within chunk, used to check chunk boundary;
         std::vector< shared< Chunk > > m_journal_chunks; // Chunks part of this journal in order.
         uint64_t m_total_size{0};                        // Total size of all chunks.
```
yamingk, line 69: at line: 71, m_end_offset is actually being updated after init, right? If so, can you update the comment?

src/tests/test_journal_vdev.cpp
```diff
@@ -470,6 +504,107 @@ TEST_F(VDevJournalIOTest, Recovery) {
     }
 }
 
+TEST_F(VDevJournalIOTest, MultipleChunkTest) {
```
yamingk, line 507: General comment for this test case: if this is a basic test for multiple chunks, fixed chunk-size, data-size is okay. 

However from this PR description, this test case is also going to be used for long hour running, I think we should make chunk-size and data_size configurable and choose a different size every run, so that we can achieve some randomized tests. 

Also we should this test itself seems only did a few insertions each for loop. We should also randomize the number of insertions along with truncations. 

Another category this test didn't cover is the crash recovery, which I think is a must have for journal vdev to make this feature stable.

src/lib/device/journal_vdev.cpp
sanebay: As discussed we round up.

src/lib/device/journal_vdev.hpp
sanebay: Yes its updated in init and during recovery time.

src/tests/test_journal_vdev.cpp
sanebay: Added two issues to track this.
https://github.com/eBay/HomeStore/issues/379
https://github.com/eBay/HomeStore/issues/378

src/lib/device/journal_vdev.hpp
yamingk: can you update the comment accordingly at line 71?

yamingk (approved):

PR #360, opened 2024-03-22
bump up dependency

xiaoxichen:


xiaoxichen:
@hkadayam  this is needed to allow us pining HS on 5.X in HO and unblock SM.

JacksonYao287 (approved):
since after  @hkadayam `s PR is merged , we need also do some change in homeobject to adopt the latest homestore changes, and before these homeobject changes are merged , homeobject and uppper layer will block on the building

so for the successfully building of current homeobject to enable nuobject 2.0 integration test, we can do this change for now.

PR #361, opened 2024-03-24
Error handling on pre-append et al. 

hkadayam:
Major changes in this PR includes:

* At present, there is no provision for consumer of replication to prevent the follower
from appending the logs. In case of linked data, its possible for followers to not have
enough space among other reasons to reject. This PR introduces pre-append and allow
consumers to return back error, which then again retried later. Not just allocation of
blocks for indirect data, but also this PR ensures that data is pulled, before accepting
pre-append on the follower that, there is no network related failures before writing the
log entry.

* In case of log entry has arrived first in raft channel, before data channel, at present
we start the timer then to wait. Instead this PR starts a global reaper thread, which wakes
up often to look for any pending fetches and thus data path, just put the fetch request to
the queue. This avoids expensive timer creation in data path.

* Added several robustness, error injection test cases to ensure stability under all conditions.

* These tests, exposes quiet a few errors in-terms of deadlocks between fetch, wait and grpc client
threads. Fixed all of them.

* Streamlined the test_raft_repl_dev testing, to be able to write the tests seamlessly.

### DIFF
--- a/src/lib/common/homestore_config.fbs
+++ b/src/lib/common/homestore_config.fbs
@@ -220,14 +220,14 @@ table Consensus {
     // Minimum log gap a replica has to be from leader before joining the replica set.
     min_log_gap_to_join: int32 = 30;
     
-    // amount of time in seconds to wait on data write before fetch data from remote;
-    wait_data_write_timer_sec: uint32 =  30 (hotswap);
+    // amount of time in millis to wait on data write before fetch data from remote;
+    wait_data_write_timer_ms: uint64 = 1500 (hotswap);
     
-    // Leadership expiry 120 seconds
-    leadership_expiry_ms: uint32 = 120000;
+    // Leadership expiry (=0 indicates 20 times heartbeat period), set -1 to never expire
+    leadership_expiry_ms: int32 = 0;
 
-    // data fetch max size limit in MB
-    data_fetch_max_size_mb: uint32 = 2;
+    // data fetch max size limit in KB (2MB by default)
+    data_fetch_max_size_kb: uint32 = 2048;
 }
 
 table HomeStoreSettings {
--- a/src/lib/common/homestore_config.fbs
+++ b/src/lib/common/homestore_config.fbs
@@ -228,6 +228,9 @@ table Consensus {
 
     // data fetch max size limit in KB (2MB by default)
     data_fetch_max_size_kb: uint32 = 2048;
+
+    // Timeout for data to be received after raft entry after which raft entry is rejected.
+    data_receive_timeout_ms: uint64 = 10000;
 }
 
 table HomeStoreSettings {
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -12,7 +12,6 @@
 #include <homestore/blkdata_service.hpp>
 #include <homestore/logstore_service.hpp>
 #include <homestore/superblk_handler.hpp>
-#include <homestore/crc.h>
 
 #include "common/homestore_assert.hpp"
 #include "common/homestore_config.hpp"
@@ -260,10 +259,6 @@ void RaftReplDev::on_fetch_data_received(intrusive< sisl::GenericRpcData >& rpc_
                 pkts.insert(pkts.end(), ret.begin(), ret.end());
             }
 
-            // TODO: Temporary code
-            crc32_t const crc = crc32_ieee(init_crc32, pkts.front().cbytes(), pkts.front().size());
-            RD_LOG(DEBUG, "Data Channel: FetchData data read completed for {} buffers, crc={}", sgs_vec.size(), crc);
-
             rpc_data->set_comp_cb([sgs_vec = std::move(sgs_vec)](boost::intrusive_ptr< sisl::GenericRpcData >&) {
                 for (auto const& sgs : sgs_vec) {
                     for (auto const& iov : sgs.iovs) {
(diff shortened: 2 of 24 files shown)


src/lib/common/homestore_config.fbs
```diff
@@ -220,14 +220,17 @@ table Consensus {
...
     
-    // amount of time in seconds to wait on data write before fetch data from remote;
-    wait_data_write_timer_sec: uint32 =  30 (hotswap);
+    // amount of time in millis to wait on data write before fetch data from remote;
+    wait_data_write_timer_ms: uint64 = 1500 (hotswap);
     
-    // Leadership expiry 120 seconds
-    leadership_expiry_ms: uint32 = 120000;
+    // Leadership expiry (=0 indicates 20 times heartbeat period), set -1 to never expire
+    leadership_expiry_ms: int32 = 0;
```
xiaoxichen, line 227: it is a bit tricky as heartbeat period is not expose.
Suggesting lets expose 
`heart_beat_interval_` in HS configuration then `election_timeout_upper_bound_`/`election_timeout_lower_bound_` and `leadership_expiry_ms` configuration as multiple of `heart_beat_interval_`

src/tests/test_raft_repl_dev.cpp
```diff
@@ -251,279 +263,290 @@ class RaftReplDevTest : public testing::Test {
...
-        std::this_thread::sleep_for(std::chrono::seconds{5});
+    g_helper->sync_for_verify_start();
+    LOGINFO("Validate all data written so far by reading them");
+    this->validate_data();
+    g_helper->sync_for_cleanup_start();
+}
 
-        LOGINFO("Switch the leader to replica_num = 1");
-        switch_all_db_leader();
+#if 0
```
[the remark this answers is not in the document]
hkadayam: Its not a typo, but for future expansion of test as a placeholder. This needs a lot of flips at tricky points and thus didn't write that yet.

src/lib/common/homestore_config.fbs
hkadayam: Not sure I got your comment, we expose heartbeat_period_ms as HS config and also elect_timeout_low and high. So leadership_expiry by default is 20 times heartbeat_period unless someone explicitly sets an absolute value.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -405,64 +423,83 @@ repl_req_ptr_t RaftReplDev::applier_create_req(repl_key const& rkey, sisl::blob
     // thing. So take state mutex and allocate the blk
     std::unique_lock< std::mutex > lg(rreq->state_mtx);
     if (rreq->state.load() & uint32_cast(repl_req_state_t::BLK_ALLOCATED)) { return rreq; }
+
+    auto hints_result = m_listener->get_blk_alloc_hints(user_header, data_size);
```
xiaoxichen, line 427: are we using the `get_blk_alloc_hints` as pre-append?

This might not enough as we might want to reject based on business logical.  say if we get a putblob but the shard doesnt exists in the replica etc.

src/tests/test_raft_repl_dev.cpp
```diff
@@ -202,7 +206,10 @@ class TestReplicatedDB : public homestore::ReplDevListener {
         g_helper->runner().execute().get();
     }
 
-    uint64_t db_num_writes() const { return m_num_commits.load(std::memory_order_relaxed); }
+    uint64_t db_commit_count() const {
+        std::shared_lock lk(db_mtx_);
+        return commit_count_;
```
yamingk, line 211: Is there any difference between `commit_count_` and `m_num_commits`?

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: yes，what`s more, the pre-append should also be called in leader, since leader might reject some request base on some logic(business or not)

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -122,12 +143,12 @@ void RaftReplDev::async_alloc_write(sisl::blob const& header, sisl::blob const&
...
+                handle_error(rreq, ReplServiceError::DRIVE_WRITE_ERROR);
+            } else {
                 auto raft_status = m_state_machine->propose_to_raft(rreq);
                 if (raft_status != ReplServiceError::OK) { handle_error(rreq, raft_status); }
-            } else {
-                HS_DBG_ASSERT(false, "Error in writing data");
-                handle_error(rreq, ReplServiceError::DRIVE_WRITE_ERROR);
             }
         });
     } else {
```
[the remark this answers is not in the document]
hkadayam: More we inline, the less journal space we might have. Since our object size is 1K min, I am not sure if we inline anything in HO (may be 1K object)

src/lib/replication/repl_dev/raft_repl_dev.cpp
hkadayam: @xiaoxichen put_blob before shard create is already covered in the follower, because all of them goes through this one call `get_blk_alloc_hints` even for put_blob. We could change the api name to something with more meaning and put more info on it. Lets keep it open and evolve that API.

@JacksonYao287 As long as leader is the always the proposer, I am not sure if we have to do anything for the leader. After all why would leader propose something it needs to veto on?

src/tests/test_raft_repl_dev.cpp
hkadayam: You are right, it is redundant, will remove that.

xiaoxichen (dismissed):

PR #364, opened 2024-03-28
Replication Snapshot and Compact 

yamingk:
Changes:
=======
1.  add snapshot creation to listener (HomeObject PR is in progress) and compact handling (called by raft and in repl dev it will persist compact_lsn as a barrier for truncation).
3.  add home raft log store truncation.
4. add resource manager resource audit routine to trigger raft log store truncation as needed.
5. Add JournalVDev critical high watermark, in its callback it will trigger raft log store truncation immeidately so that it doesn't need to want for next timer audit.

Reviewers please focus on resource mgr, home raft repl log, raft repl dev and JournalVDev. 
Not much logic change in other layer/files. 

Testing:
=======
1. snapshot creation and truncation can be triggered from the logs with below command line.
./bin/test_raft_repl_dev --gtest_filter=*Snapshot_and_Compact* --log_mods replication:debug --num_io=999999 --snapshot_distance=200 --num_raft_logs_resv=20000 --res_mgr_audit_timer_ms=120000
2. it also needs to be added to replication long duration test.

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -40,6 +40,8 @@ RaftReplDev::RaftReplDev(RaftReplService& svc, superblk< raft_repl_dev_superblk 
         m_next_dsn = m_rd_sb->last_applied_dsn + 1;
         m_commit_upto_lsn = m_rd_sb->commit_lsn;
         m_last_flushed_commit_lsn = m_commit_upto_lsn;
+        m_compact_lsn = m_rd_sb->compact_lsn;
+
         m_rdev_name = fmt::format("rdev{}", m_rd_sb->group_ordinal);
 
         // Its ok not to do compare exchange, because loading is always single threaded as of now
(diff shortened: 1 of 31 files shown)


src/lib/replication/repl_dev/raft_repl_dev.h
```diff
@@ -86,6 +89,8 @@ class RaftReplDev : public ReplDev,
 
     RaftReplDevMetrics m_metrics;
 
+    nuraft::ptr< nuraft::snapshot > m_last_snapshot{nullptr};
```
hkadayam, line 92: Why do homestore need to maintain the last snapshot. Isn't upper layer going to maintain snapshot. It could potentially store multiple snapshots?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -108,6 +110,18 @@ bool RaftReplDev::join_group() {
 
 void RaftReplDev::use_config(json_superblk raft_config_sb) { m_raft_config_sb = std::move(raft_config_sb); }
 
+void RaftReplDev::on_create_snapshot(nuraft::snapshot& s, nuraft::async_result< bool >::handler_type& when_done) {
+    RD_LOG(DEBUG, "create_snapshot last_idx={}/term={}", s.get_last_log_idx(), s.get_last_log_term());
+    repl_snapshot snapshot{.last_log_idx_ = s.get_last_log_idx(), .last_log_term_ = s.get_last_log_term()};
+    auto result = m_listener->on_create_snapshot(snapshot).get();
```
hkadayam, line 116: I think we need to ask upper layer to create the snapshot and require upper layer snapshot to be derived from repl_snapshot and store required parameters like last_term and last_log_idx. This way how snapshot created and maintained is completely controlled by upper layer.

src/lib/replication/repl_dev/raft_repl_dev.h
yamingk: You point is valid. I also thought about it, I did this only because raft is asking for a nuraft::snapshot and we don't want to expose to upper layer, and in this snapshot, there is only log_idx and log_term which is not mutable to upper layer (right?), so I didn't ask for upper layer to get the last snapshot.

Issue #377, opened 2024-04-11
cannot round up disk size


xiaoxichen:
2024-04-11 15:49:47.000	
[04/11/24 00:49:46.390] [storage_mgr] [info] [11] [device_manager.cpp:311:create_vdev] Virtal Dev=index of size=511.56mb successfully created
2024-04-11 15:49:47.000	
[04/11/24 00:49:46.390] [storage_mgr] [info] [11] [device_manager.cpp:222:create_vdev] blkdata Virtual device is attempted to be created with size=7.89gb, it needs to be rounded to new_size=7.93gb
2024-04-11 15:49:47.000	
[04/11/24 00:49:46.390] [storage_mgr] [info] [11] [device_manager.cpp:230:create_vdev] blkdata Virtual device is attempted to be created with num_chunks=63.48kb, it needs to be adjust to new_num_chunks=640

xiaoxichen:
```
            vparam.vdev_size = sisl::round_up(vparam.vdev_size, vparam.num_chunks * vparam.blk_size);
            if (input_vdev_size != vparam.vdev_size) {
                LOGINFO(
                    "{} Virtual device is attempted to be created with size={}, it needs to be rounded to new_size={}",
                    vparam.vdev_name, in_bytes(input_vdev_size), in_bytes(vparam.vdev_size));
            }
```
this is wrong as we should not round to input num_chunks

xiaoxichen:
```
    // Adjust the maximum number chunks requested before round up vdev size.
    uint32_t max_num_chunks = 0;
    for (const auto& d : m_dev_infos) {
        max_num_chunks += hs_super_blk::max_chunks_in_pdev(d);
    }

```
this is wrong as well , we sum up the max_chunks across all pdevs.  It should be vdevSize / min_chunk_size.

xiaoxichen:
we didnt round the chunk boundary to optimal_page_size

PR #381, opened 2024-04-15
Bug fixes for create_vdev

xiaoxichen:
1. max_num_chunks should based on vdev_size/min_chunk_size, previous implementation wrongly include other pdevs that doesnt belongs to this vdev(e.g different type).

2. we should never round up vdev size as this might causing vdev size exceed underlaying pdev size.

3. a few other fixes.

Add a set of sanity checks as well as UT to ensure chunks are align to pdev->align_size().

Fixes: https://github.com/eBay/HomeStore/issues/377

### DIFF
--- a/src/lib/device/device_manager.cpp
+++ b/src/lib/device/device_manager.cpp
@@ -192,6 +192,7 @@ shared< VirtualDev > DeviceManager::create_vdev(vdev_parameters&& vparam) {
     std::vector< PhysicalDev* > pdevs = pdevs_by_type_internal(vparam.dev_type);
     RELEASE_ASSERT_GT(pdevs.size(), 0, "Unable to find any pdevs for given vdev type, can't create vdev");
     RELEASE_ASSERT(vparam.blk_size % pdevs[0]->align_size() == 0, "blk_size should be multiple of pdev align_size");
+    RELEASE_ASSERT(vparam.blk_size % 512 == 0, "blk_size should be multiple of 512bytes to be DMA aligned");
     // Identify the number of chunks
     if (vparam.multi_pdev_opts == vdev_multi_pdev_opts_t::ALL_PDEV_STRIPED) {
         auto total_streams = std::accumulate(pdevs.begin(), pdevs.end(), 0u,
@@ -208,9 +209,8 @@ shared< VirtualDev > DeviceManager::create_vdev(vdev_parameters&& vparam) {
     // Based on the min chunk size, we calculate the max number of chunks that can be created in each target pdev
     uint32_t min_chunk_size = hs_super_blk::min_chunk_size(vparam.dev_type);
     // FIXME: it is possible that each vdev is less than max_num_chunks, but total is more than MAX_CHUNKS_IN_SYSTEM.
-    uint32_t max_num_chunks = vparam.vdev_size / min_chunk_size > hs_super_blk::MAX_CHUNKS_IN_SYSTEM
-        ? hs_super_blk::MAX_CHUNKS_IN_SYSTEM
-        : vparam.vdev_size / min_chunk_size;
+    // uint32 convert is safe as it only overflow when vdev size > 64PB with 16MB min_chunk_size.
+    uint32_t max_num_chunks = std::min(uint32_t(vparam.vdev_size / min_chunk_size), hs_super_blk::MAX_CHUNKS_IN_SYSTEM);
 
     auto input_vdev_size = vparam.vdev_size;
     if (vparam.size_type == vdev_size_type_t::VDEV_SIZE_STATIC) {
--- a/src/lib/device/device_manager.cpp
+++ b/src/lib/device/device_manager.cpp
@@ -188,11 +188,9 @@ shared< VirtualDev > DeviceManager::create_vdev(vdev_parameters&& vparam) {
     if (vdev_id == sisl::Bitset::npos) { throw std::out_of_range("System has no room for additional vdev"); }
     m_vdev_id_bm.set_bit(vdev_id);
 
-    // Determine if we have a devices available on requested dev_tier. If so use them, else fallback to data tier
     std::vector< PhysicalDev* > pdevs = pdevs_by_type_internal(vparam.dev_type);
     RELEASE_ASSERT_GT(pdevs.size(), 0, "Unable to find any pdevs for given vdev type, can't create vdev");
     RELEASE_ASSERT(vparam.blk_size % pdevs[0]->align_size() == 0, "blk_size should be multiple of pdev align_size");
-    RELEASE_ASSERT(vparam.blk_size % 512 == 0, "blk_size should be multiple of 512bytes to be DMA aligned");
     // Identify the number of chunks
     if (vparam.multi_pdev_opts == vdev_multi_pdev_opts_t::ALL_PDEV_STRIPED) {
         auto total_streams = std::accumulate(pdevs.begin(), pdevs.end(), 0u,
--- a/src/lib/device/device_manager.cpp
+++ b/src/lib/device/device_manager.cpp
@@ -23,6 +23,7 @@
 #include <homestore/homestore_decl.hpp>
 #include "device/chunk.h"
 #include "device/device.h"
+#include "device/hs_super_blk.h"
 #include "device/physical_dev.hpp"
 #include "device/virtual_dev.hpp"
 #include "common/homestore_utils.hpp"
@@ -189,8 +190,8 @@ shared< VirtualDev > DeviceManager::create_vdev(vdev_parameters&& vparam) {
 
     // Determine if we have a devices available on requested dev_tier. If so use them, else fallback to data tier
     std::vector< PhysicalDev* > pdevs = pdevs_by_type_internal(vparam.dev_type);
-    RELEASE_ASSERT_GT(pdevs.size(), 0, "Unable to find any pdevs for even data tier, can't create vdev");
-
+    RELEASE_ASSERT_GT(pdevs.size(), 0, "Unable to find any pdevs for given vdev type, can't create vdev");
+    RELEASE_ASSERT(vparam.blk_size % pdevs[0]->align_size() == 0, "blk_size should be multiple of pdev align_size");
     // Identify the number of chunks
     if (vparam.multi_pdev_opts == vdev_multi_pdev_opts_t::ALL_PDEV_STRIPED) {
         auto total_streams = std::accumulate(pdevs.begin(), pdevs.end(), 0u,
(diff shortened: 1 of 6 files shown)


src/lib/device/device_manager.cpp
```diff
@@ -204,11 +205,12 @@ shared< VirtualDev > DeviceManager::create_vdev(vdev_parameters&& vparam) {
...
 
-    // Adjust the maximum number chunks requested before round up vdev size.
-    uint32_t max_num_chunks = 0;
-    for (const auto& d : m_dev_infos) {
-        max_num_chunks += hs_super_blk::max_chunks_in_pdev(d);
-    }
+    // Based on the min chunk size, we calculate the max number of chunks that can be created in each target pdev
+    uint32_t min_chunk_size = hs_super_blk::min_chunk_size(vparam.dev_type);
+    // FIXME: it is possible that each vdev is less than max_num_chunks, but total is more than MAX_CHUNKS_IN_SYSTEM.
+    uint32_t max_num_chunks = vparam.vdev_size / min_chunk_size > hs_super_blk::MAX_CHUNKS_IN_SYSTEM
```
yamingk, line 211: How about: `uint32_t max_num_chunks = std::min(vparam.vdev_size / min_chunk_size, hs_super_blk::MAX_CHUNKS_IN_SYSTEM);`

src/lib/device/device_manager.cpp
```diff
@@ -204,11 +205,12 @@ shared< VirtualDev > DeviceManager::create_vdev(vdev_parameters&& vparam) {
...
     }
 
-    // Adjust the maximum number chunks requested before round up vdev size.
-    uint32_t max_num_chunks = 0;
-    for (const auto& d : m_dev_infos) {
-        max_num_chunks += hs_super_blk::max_chunks_in_pdev(d);
-    }
+    // Based on the min chunk size, we calculate the max number of chunks that can be created in each target pdev
+    uint32_t min_chunk_size = hs_super_blk::min_chunk_size(vparam.dev_type);
+    // FIXME: it is possible that each vdev is less than max_num_chunks, but total is more than MAX_CHUNKS_IN_SYSTEM.
```
yamingk, line 210: This is a true statement, but it won't cause any problem, right?
`max_num_chunks` only garunteens chunk_size won't be smaller than the `min_chunk_size`. 
The actual `vparam.num_chunks` is a std::min of itself and this `max_num_chunks`, line: 223. 
And the input vparam.num_chunks is provided by HS consumer and this consumer can control all the services requested in HS won't exceed MAX_CHUNK_IN_SYSTEM, right?

src/lib/device/device_manager.cpp
```diff
@@ -256,7 +262,29 @@ shared< VirtualDev > DeviceManager::create_vdev(vdev_parameters&& vparam) {
...
+
+        vparam.vdev_size = sisl::round_down(vparam.vdev_size, vparam.chunk_size);
+        if (input_vdev_size != vparam.vdev_size) {
+            LOGINFO(
+                "{} Virtual device is attempted to be created with size={}, it needs to be rounded to new_size={}",
+                vparam.vdev_name, in_bytes(input_vdev_size), in_bytes(vparam.vdev_size));
+        }
     }
+    // sanity checks
+    RELEASE_ASSERT(vparam.vdev_size % vparam.chunk_size == 0, "vdev_size should be multiple of chunk_size");
```
yamingk, line 282: good set of asserts I was about to suggest.

src/lib/device/device_manager.cpp
```diff
@@ -217,31 +219,35 @@ shared< VirtualDev > DeviceManager::create_vdev(vdev_parameters&& vparam) {
...
             auto input_num_chunks = vparam.num_chunks;
             vparam.num_chunks = std::min(vparam.num_chunks, max_num_chunks);
             if (input_num_chunks != vparam.num_chunks) {
                 LOGINFO("{} Virtual device is attempted to be created with num_chunks={}, it needs to be adjust to "
                         "new_num_chunks={}",
                         vparam.vdev_name, in_bytes(input_num_chunks), in_bytes(vparam.num_chunks));
             }
+
+            // this ensure chunk_size % vparam.blk_size == 0
+            vparam.vdev_size = sisl::round_down(vparam.vdev_size, vparam.num_chunks * vparam.blk_size);
```
yamingk, line 231: I understand vparam.blk_size are provided by HS services which is either atomic_page_size or phy_page_size which are all retrieved from iomgr, as device_manager layer self constraints, should we also make an assert that this blk_size needs to be DMA (512) aligned?

src/tests/test_device_manager.cpp
```diff
@@ -174,6 +175,29 @@ TEST_F(DeviceMgrTest, StripedVDevCreation) {
     this->validate_striped_vdevs();
 }
 
+TEST_F(DeviceMgrTest, SmallStripedVDevCreation) {
```
yamingk, line 178: Not suggesting it needs to be included in this PR. We can open tickets to add: 
- Mixed drive types with very large drive size (simulated). 
- Pure HDD with very large drive size (32TB). 
- Pure Fast drive with very large drive size (32TB).

src/lib/device/device_manager.cpp
```diff
@@ -256,7 +262,29 @@ shared< VirtualDev > DeviceManager::create_vdev(vdev_parameters&& vparam) {
...
+        auto input_chunk_size = vparam.chunk_size;
+        vparam.chunk_size = std::max(vparam.chunk_size, min_chunk_size);
+        vparam.chunk_size = sisl::round_up(vparam.chunk_size, vparam.blk_size);
+        if (input_chunk_size != vparam.chunk_size) {
+            LOGINFO("{} Virtual device is attempted to be created with chunk_size={}, it needs to be adjust to "
+                    "new_chunk_size={}",
+                    vparam.vdev_name, in_bytes(input_chunk_size), in_bytes(vparam.chunk_size));
+        }
+
+        vparam.vdev_size = sisl::round_down(vparam.vdev_size, vparam.chunk_size);
```
yamingk, line 274: shouldn't we also calculate vparam.num_chunks for this else branch?

src/lib/device/device_manager.cpp
xiaoxichen: the vparam.vdev_size / min_chunk_size is a uint64 so I tried to avoid the complexity of force type convert ---  i.e `vparam.vdev_size / min_chunk_size` can overflow uint32

src/lib/device/device_manager.cpp
xiaoxichen: Yes it will not causing issue ,just fail the create vdev request.
I was thinking to tune the num_chunks based on the available chunk count. i.e if we have 1000 chunks available and the input requested 2000, we can tune that down to 1000... this is just a minor improvement .

src/lib/device/device_manager.cpp
xiaoxichen: oh it is the same, sorry will fix that

xiaoxichen: it is theoretically possible but not practically possible  as `min_chunk_size = 16MB * max_uint_32` = 2^56 = 64PB...

We will not have that big device. Will change to use force type convert.

src/lib/device/device_manager.cpp
yamingk: It should not fail the create vdev request, if the consumer of HS set the correct num_chunks for each services during hs.format_and_start, right? 

Your thought is also valid, that consumer just supply a total num_chunks consumer want to create and let HS do whatever trick to consume available chunks.

yamingk (approved):

src/lib/device/device_manager.cpp
xiaoxichen: Thinking about the case, we have 64K chunk limit across all vdevs. This 64K is a magic number that the client do not know. 
Even client knows like homeobject,  several service created with 
```
            HomeStore::instance()->format_and_start({
                {HS_SERVICE::META, hs_format_params{.dev_type = run_on_type, .size_pct = 5.0}},
                {HS_SERVICE::LOG, hs_format_params{.dev_type = run_on_type, .size_pct = 10.0, .chunk_size = 32 * Mi}},
                {HS_SERVICE::INDEX, hs_format_params{.dev_type = run_on_type, .size_pct = 5.0}},
                {HS_SERVICE::REPLICATION,
                 hs_format_params{.dev_type = run_on_type,
                                  .size_pct = 79.0,
                                  .num_chunks = 63000,
                                  .block_size = _data_block_size,
                                  .alloc_type = blk_allocator_type_t::append,
                                  .chunk_sel_type = chunk_selector_type_t::CUSTOM}},
            });
```

It is very hard to know if we are exceeding 64K,  as LOG device based on chunk_size, num_chunks is in-direct.  META/INDEX doesnt specified num_chunk but based on the default value of `hs_format_params`....


Later if LOG moved to dynamic size it is more tricky as the max size of the log device can be limited by the available chunk count left by other svc. e.g if data take 65500 then only 34 chunks left for log dev .

src/lib/device/device_manager.cpp
```diff
@@ -204,11 +206,11 @@ shared< VirtualDev > DeviceManager::create_vdev(vdev_parameters&& vparam) {
...
 
-    // Adjust the maximum number chunks requested before round up vdev size.
-    uint32_t max_num_chunks = 0;
-    for (const auto& d : m_dev_infos) {
-        max_num_chunks += hs_super_blk::max_chunks_in_pdev(d);
-    }
+    // Based on the min chunk size, we calculate the max number of chunks that can be created in each target pdev
+    uint32_t min_chunk_size = hs_super_blk::min_chunk_size(vparam.dev_type);
+    // FIXME: it is possible that each vdev is less than max_num_chunks, but total is more than MAX_CHUNKS_IN_SYSTEM.
+    // uint32 convert is safe as it only overflow when vdev size > 64PB with 16MB min_chunk_size.
```
JacksonYao287, line 212: should we add an assert to make sure vdev size is smaller than or eaqual to 64 PB?

xiaoxichen: I dont think we can have a vdev > 64PB in any near future.

src/lib/device/device_manager.cpp
yamingk: I agree right now it is a magic number. We can get it through a static API to get the max_num_chunks from homestore (it can optimally to return the "real" max number, already taking min_chunk_size into account). 

I was thinking consumer need to specify num_chunks for all the services. In your case above, if we want to support only one service specify num_chunks, and let other services to land based on vdev size, then yes we also need to play some tricks in homestore.

xiaoxichen: Yes, maybe lets ignore the well known number part for now, this is a easy fix issue whenever we want to fix.

>I was thinking consumer need to specify num_chunks for all the requested services. 
I feel very confused regarding the path  of dynamic size vdev introduced by LOG....  by dynamic sizing it naturally cannot provide num_chunks,  but we need to `reserve`  N  chunks for it to be dynamic within [0, N * CHUNK_SIZE)

yamingk (approved):

PR #382, opened 2024-04-17
prevent chunk_size overflow.

xiaoxichen:
The default num_chunks is 1, if upper layer doesnt set num_chunk properly,  the
chunk_size(uint32) = device_size(uint64) / num_chunks(uint32)

can be overlow.


JacksonYao287 (approved):
LGTM

PR #384, opened 2024-04-17
Fix varsizeBlkAllocator panic when initilized with small size.

xiaoxichen:
When the allocator size is small, it is possible we have a size 0 level in slab, which will crash in constructing SlabCacheQueue


hkadayam (approved):
Looks good to me

raakella1 (dismissed):
LG!

raakella1 (approved):
LG!

PR #387, opened 2024-04-18
Add baseline resync with read and write support

sanebay:
Add generic snapshot context and data structure.
Add callbacks for listener as discussed.
Use nuraft read and write logical snapshot callbacks to transfer the data.
Do test on both read and write for the baseline resync. 
Key and values were pushed by leader and written by follower and its tested.  
Currently verification of the data in the test are disabled but have been locally tested.

### DIFF
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -59,9 +59,60 @@ struct repl_key {
     std::string to_string() const { return fmt::format("server={}, term={}, dsn={}", server_id, term, dsn); }
 };
 
-struct repl_snapshot {
-    uint64_t last_log_idx_{0};
-    uint64_t last_log_term_{0};
+using repl_snapshot = nuraft::snapshot;
+using repl_snapshot_ptr = nuraft::ptr< nuraft::snapshot >;
+
+// Consumers of the ReplDevListener dont have to know what underlying
+// snapshot implementation is used. Consumers can export and save the state
+// of the snapshot using serialize and load the state using deserialize.
+class snapshot_context {
+public:
+    snapshot_context(int64_t lsn) : lsn_(lsn) {}
+    virtual ~snapshot_context() = default;
+    virtual void deserialize(const sisl::io_blob_safe& snp_ctx) = 0;
+    virtual sisl::io_blob_safe serialize() = 0;
+    int64_t get_lsn() { return lsn_; }
+
+protected:
+    int64_t lsn_;
+};
+
+class nuraft_snapshot_context : public snapshot_context {
+public:
+    nuraft_snapshot_context(nuraft::snapshot& snp) : snapshot_context(snp.get_last_log_idx()) {
+        auto snp_buf = snp.serialize();
+        snapshot_ = nuraft::snapshot::deserialize(*snp_buf);
+    }
+
+    void deserialize(const sisl::io_blob_safe& snp_ctx) override {
+        // Load the context from the io blob to nuraft buffer.
+        auto snp_buf = nuraft::buffer::alloc(snp_ctx.size());
+        nuraft::buffer_serializer bs(snp_buf);
+        bs.put_raw(snp_ctx.cbytes(), snp_ctx.size());
+        snapshot_ = nuraft::snapshot::deserialize(bs);
+        lsn_ = snapshot_->get_last_log_idx();
+    }
+
+    sisl::io_blob_safe serialize() override {
+        // Dump the context from nuraft buffer to the io blob.
+        auto snp_buf = snapshot_->serialize();
+        sisl::io_blob_safe blob{s_cast< size_t >(snp_buf->size())};
+        std::memcpy(blob.bytes(), snp_buf->data_begin(), snp_buf->size());
+        return blob;
+    }
+
+    nuraft::ptr< nuraft::snapshot > nuraft_snapshot() { return snapshot_; }
+
+private:
+    nuraft::ptr< nuraft::snapshot > snapshot_;
+};
+
+struct snapshot_data {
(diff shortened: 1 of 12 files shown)


src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -293,7 +295,13 @@ bool HomeRaftLogStore::compact(ulong compact_lsn) {
     // we rely on resrouce mgr timer to trigger truncate for all log stores in system;
     // this will be friendly for multiple logstore on same logdev;
 
-    // m_log_store->truncate(to_store_lsn(compact_lsn));
+#ifdef _PRERELEASE
+    if (iomgr_flip::instance()->test_flip("force_home_raft_log_truncate")) {
```
hkadayam, line 299: Suggest a parameterized flip, so you can force raft log truncate on specific logstore. This will trigger on every logstore truncate always.

src/lib/replication/log_store/repl_log_store.cpp
```diff
@@ -9,7 +9,7 @@ namespace homestore {
 uint64_t ReplLogStore::append(nuraft::ptr< nuraft::log_entry >& entry) {
     // We don't want to transform anything that is not an app log
     if (entry->get_val_type() != nuraft::log_val_type::app_log) { return HomeRaftLogStore::append(entry); }
-
+    if (entry->get_buf_ptr()->size() == 0) { return HomeRaftLogStore::append(entry); }
```
hkadayam, line 12: nit: isn't easier to put || with above condition?

src/include/homestore/replication/repl_dev.h
```diff
@@ -59,9 +59,60 @@ struct repl_key {
...
+
+protected:
+    int64_t lsn_;
+};
+
+class nuraft_snapshot_context : public snapshot_context {
+public:
+    nuraft_snapshot_context(nuraft::snapshot& snp) : snapshot_context(snp.get_last_log_idx()) {
+        auto snp_buf = snp.serialize();
+        snapshot_ = nuraft::snapshot::deserialize(*snp_buf);
```
szmyd, line 84: This seems strange; seems like we make two copies? Can we not just copy `snp` into a allocated prt<snapshot>?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1031,6 +1030,7 @@ std::pair< bool, nuraft::cb_func::ReturnCode > RaftReplDev::handle_raft_event(nu
             auto reqs = sisl::VectorPool< repl_req_ptr_t >::alloc();
             for (auto& entry : entries) {
                 if (entry->get_val_type() != nuraft::log_val_type::app_log) { continue; }
+                if (entry->get_buf_ptr()->size() == 0) { continue; }
```
szmyd, line 1033: What's the case where we have to handle zero sized buffers now?

src/lib/replication/repl_dev/raft_state_machine.h
```diff
@@ -111,9 +111,12 @@ class RaftStateMachine : public nuraft::state_machine {
...
 
-    bool apply_snapshot(nuraft::snapshot&) override { return false; }
-
     void create_snapshot(nuraft::snapshot& s, nuraft::async_result< bool >::handler_type& when_done) override;
+    int read_logical_snp_obj(nuraft::snapshot& s, void*& user_ctx, ulong obj_id, raft_buf_ptr_t& data_out,
+                             bool& is_last_obj) override;
+    void save_logical_snp_obj(nuraft::snapshot& s, ulong& obj_id, nuraft::buffer& data, bool is_first_obj,
+                              bool is_last_obj) override;
+    bool apply_snapshot(nuraft::snapshot& s) override;
     nuraft::ptr< nuraft::snapshot > last_snapshot() override;
```
szmyd, line 120: Should we implement: 
```void nuraft::state_machine::free_user_snp_ctx(void*& user_snp_ctx);```

**Review by szmyd — COMMENTED** [review]
Some comments; still reviewing...

File: src/include/homestore/replication/repl_dev.h (hunk 3 above)
Comment (sanebay) [reply src/include/homestore/replication/repl_dev.h:84]: nuraft::snapshot doesnt have a copy constructor to use the argument snp. So had to do this way.

File: src/lib/replication/repl_dev/raft_repl_dev.cpp (hunk 4 above)
Comment (sanebay) [reply src/lib/replication/repl_dev/raft_repl_dev.cpp:1033]: In raft repl test case, we can also send data size as 0.

File: src/lib/replication/repl_dev/raft_state_machine.h (hunk 5 above)
Comment (sanebay) [reply src/lib/replication/repl_dev/raft_state_machine.h:120]: Yeah right,  we should.

File: src/tests/test_raft_repl_dev.cpp [hunk 6]
```diff
@@ -704,6 +853,66 @@ TEST_F(RaftReplDevTest, GCReplReqs) {
...
+        this->wait_for_all_commits();
+        LOGINFO("Got all commits for replica 0 and 2");
+    }
+
+    if (g_helper->replica_num() == 0) {
+        // Leader does manual snapshot and truncate
+        LOGINFO("Leader create snapshot and truncate");
+        this->create_snapshot();
+        this->truncate(0);
+    }
```
szmyd, line 897: Should we just write the `snapshot_freq_distance` number of logs to cause an internal snapshot like we'll experience in a real setup? We never call `snapshot` directly from above the library from what I know.

sanebay: We do set snapshot_freq_distance at line 946.

szmyd (approved):
lgtm

PR #392, opened 2024-04-25
Added remove repl_dev and made repl_req much more contained

hkadayam:


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -108,6 +108,23 @@ bool RaftReplDev::join_group() {
...
+
+    auto err = m_state_machine->propose_to_raft(std::move(rreq));
+    if (err != ReplServiceError::OK) {
+        m_stage.update([](auto* stage) { *stage = repl_dev_stage_t::ACTIVE; });
+        return folly::makeSemiFuture< ReplServiceError >(std::move(err));
+    }
+
+    return m_destroy_promise.getSemiFuture();
+}
+
```
hkadayam, line 127: This change is for RemoveReplDev feature.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -122,41 +139,41 @@ void RaftReplDev::on_create_snapshot(nuraft::snapshot& s, nuraft::async_result<
...
-    rreq->rkey = repl_key{.server_id = server_id(), .term = raft_server()->get_term(), .dsn = m_next_dsn.fetch_add(1)};
+
+    auto const guard = m_stage.access();
+    if (auto const stage = *guard.get(); stage != repl_dev_stage_t::ACTIVE) {
+        RD_LOGW("Raft channel: Not ready to accept writes, stage={}", enum_name(stage));
+        handle_error(rreq,
+                     (stage == repl_dev_stage_t::INIT) ? ReplServiceError::SERVER_IS_JOINING
+                                                       : ReplServiceError::SERVER_IS_LEAVING);
+        return;
+    }
```
hkadayam, line 153: This too is RemoveReplDev feature

src/include/homestore/replication/repl_dev.h
```diff
@@ -61,51 +69,89 @@ struct repl_req_ctx : public boost::intrusive_ref_counter< repl_req_ctx, boost::
...
+    bool is_localize_pending() const { return m_is_jentry_localize_pending; }
+    bool is_data_inlined() const { return (m_op_code == journal_type_t::HS_DATA_INLINED); }
+    bool has_linked_data() const { return (m_op_code == journal_type_t::HS_DATA_LINKED); }
 
-    uint64_t dsn() const { return rkey.dsn; }
-    uint64_t term() const { return rkey.term; }
-    void alloc_journal_entry(uint32_t size, bool is_raft_buf);
     raft_buf_ptr_t& raft_journal_buf();
     uint8_t* raw_journal_buf();
+    flatbuffers::FlatBufferBuilder& create_fb_builder() { return m_fb_builder; }
```
yamingk, line 100: `get_fb_builder`?

src/include/homestore/replication/repl_dev.h
```diff
@@ -61,51 +69,89 @@ struct repl_req_ctx : public boost::intrusive_ref_counter< repl_req_ctx, boost::
...
+    void release_fb_builder() { m_fb_builder.Release(); }
```
yamingk, line 101: move to modifiers part?

src/include/homestore/replication/repl_dev.h
```diff
@@ -61,51 +69,89 @@ struct repl_req_ctx : public boost::intrusive_ref_counter< repl_req_ctx, boost::
...
 
+    /////////////////////// Non modifiers methods //////////////////
     std::string to_string() const;
     std::string to_compact_string() const;
-    Clock::time_point created_time() const { return start_time; }
+    Clock::time_point created_time() const { return m_start_time; }
+
+    /////////////////////// All Modifiers methods //////////////////
+    ReplServiceError alloc_local_blks(cshared< ReplDevListener >& listener, uint32_t data_size);
```
yamingk, line 109: Can we put api documents here for some modifiers for readability for whom that are not familiar with this code?

src/include/homestore/replication/repl_dev.h
```diff
@@ -61,51 +69,89 @@ struct repl_req_ctx : public boost::intrusive_ref_counter< repl_req_ctx, boost::
     friend class SoloReplDev;
```
yamingk, line 69: Can we also put a important comment on top of `repl_req_ctx` that we should avoid putting members in public area because we are trying to achieve safe access for this structure so that engineers from other area can also achieve this goal in the future.

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -335,24 +380,23 @@ void RaftReplService::fetch_pending_data() {
...
 }
 
-#if 0
 void RaftReplService::gc_repl_devs() {
     std::unique_lock lg(m_rd_map_mtx);
-    for (it = m_rd_map.begin(); it != m_rd_map.end();) {
+    for (auto it = m_rd_map.begin(); it != m_rd_map.end();) {
         auto rdev = std::dynamic_pointer_cast< RaftReplDev >(it->second);
         if (rdev->is_destroy_pending() &&
             (get_elapsed_time_sec(rdev->destroyed_time()) >=
```
yamingk, line 388: is there a reason why we need this check given this function is only called in a configured timer (`generic.repl_dev_cleanup_interval_sec`) already?

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -247,6 +255,40 @@ AsyncReplResult< shared< ReplDev > > RaftReplService::create_repl_dev(group_id_t
                   : make_async_error< shared< ReplDev > >(ReplServiceError::SERVER_NOT_FOUND);
 }
 
+// Remove repl_dev for raft is a 3 step process.
+// Step 1: The leader of the repl_dev proposes all members an entry to delete the entire group. As part of commit of
+// this entry, all members call leave() method, which does the following:
+//      a) Mark in cache that the group is destroyed, so no subsequent write operations are permitted (relevant to
+//      leader only)
+//      b) Call the upper_layer on_destroy(), so that consumers persistent meta information can be deleted.
```
yamingk, line 263: it requires consumer to do its on_destroy in a sync way, right?

src/lib/replication/repl_dev/raft_repl_dev.h
```diff
@@ -22,14 +23,17 @@ struct raft_repl_dev_superblk : public repl_dev_superblk {
     uint32_t raft_sb_version{RAFT_REPL_DEV_SB_VERSION};
     logstore_id_t free_blks_journal_id; // Logstore id for storing free blkid records
     uint8_t is_timeline_consistent; // Flag to indicate whether the recovery of followers need to be timeline consistent
-    uint64_t last_applied_dsn;      // Last applied data sequence Number
+    uint64_t last_applied_dsn;      // Last applied data sequence number
+    uint8_t destroy_pending;        // Flag to indicate whether the group is in destroy pending state
```
yamingk, line 27: Print it on RaftReplDev::RaftReplDev(...) where we print other info as well.

yamingk, line 27: I didn't see any place init this field. Shouldn't we initialize it to false, or is the whole superblk structure already zeroed somewhere before passed to RaftReplDev constructor? A specific init won't might be better here.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -166,176 +183,57 @@ void RaftReplDev::async_alloc_write(sisl::blob const& header, sisl::blob const&
             }
         });
     } else {
-        rreq->value_inlined = true;
-        RD_LOG(DEBUG, "Skipping data channel send since value size is 0");
+        RD_LOGD("Skipping data channel send since value size is 0");
+        rreq->add_state(repl_req_state_t::DATA_WRITTEN);
```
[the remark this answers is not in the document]
yamingk: it is explained here: https://github.com/eBay/HomeStore/blob/master/src/lib/replication/log_store/repl_log_store.cpp#L44

leader synchronously flush the data so doesn't need wait for data written flag.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -166,176 +183,57 @@ void RaftReplDev::async_alloc_write(sisl::blob const& header, sisl::blob const&
...
         auto raft_status = m_state_machine->propose_to_raft(rreq);
         if (raft_status != ReplServiceError::OK) { handle_error(rreq, raft_status); }
     }
 }
 
-void RaftReplDev::push_data_to_all_followers(repl_req_ptr_t rreq) {
-    auto& builder = rreq->fb_builder;
+void RaftReplDev::push_data_to_all_followers(repl_req_ptr_t rreq, sisl::sg_list const& data) {
```
yamingk, line 193: are there any logical change for all these functions we should pay attention in this file besides the rreq safety access?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -356,154 +255,205 @@ void RaftReplDev::on_push_data_received(intrusive< sisl::GenericRpcData >& rpc_d
...
+}
+
+folly::Future< folly::Unit > RaftReplDev::notify_after_data_written(std::vector< repl_req_ptr_t >* rreqs) {
+    std::vector< folly::Future< folly::Unit > > futs;
+    futs.reserve(rreqs->size());
+    std::vector< repl_req_ptr_t > unreceived_data_reqs;
+
+    // Walk through the list of requests and wait for the data to be received and written
+    for (auto const& rreq : *rreqs) {
+        if ((rreq == nullptr) || (!rreq->has_linked_data())) { continue; }
```
yamingk, line 343: When will rreq be nullptr? it is retrieved from `ReplLogStore::end_of_append_batch` via `m_sm.lsn_to_req`, so after a log entry is appended and it can happen that some lsn doesn't map to any rreq, is it a reasonable case we should accept?

src/include/homestore/replication/repl_dev.h
hkadayam: I thought about that, but wanted to have a contrary to release and we wanted to use builder once. That is why let it be create_fb_builder.

src/include/homestore/replication/repl_dev.h
hkadayam: Done

src/include/homestore/replication/repl_dev.h
hkadayam: sure will do.

src/include/homestore/replication/repl_dev.h
hkadayam: Done.

src/lib/replication/service/raft_repl_service.cpp
hkadayam: The timer could wake up more frequently and it is a common timer for all repl_devs. Also each repl_dev would like to want to wait more time, especially in cases of migrate PG. So adding additional check to be more precise.

src/lib/replication/service/raft_repl_service.cpp
hkadayam: Consumer can do on_destroy in async way as long as they persist that information in their metablk.

src/lib/replication/repl_dev/raft_repl_dev.cpp
hkadayam: There is no logic change. Changes which is important is save_pushed_data and save_fetched_data, but there is no logic change there as well.

src/tests/test_raft_repl_dev.cpp
```diff
@@ -577,6 +628,38 @@ TEST_F(RaftReplDevTest, Snapshot_and_Compact) {
...
+    // Step 3: Shutdown one of the follower and remove another repl_dev, once the follower is up, it should remove the
+    // repl_dev and proceed
+    LOGINFO("Shutdown one of the followers (replica=1) and then remove dbs on other members. Expect replica=1 to "
+            "remove after it is up");
+    this->restart_replica(1, 15 /* shutdown_delay_sec */);
+    this->remove_db(dbs_.back(), true /* wait_for_removal */);
+
+    // TODO: Once generic crash flip/test_infra is available, use flip to crash during removal and restart them to see
+    // if records are being removed
+    g_helper->sync_for_cleanup_start();
```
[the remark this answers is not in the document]
hkadayam: The tests wait for all the followers to finish remove repl_dev. Method `remove_db()` has `wait_for_removal` which waits for the replica to remove. Then sync_for_cleanup_start() will not start on all replicas until all replicas finished previous step and previous step will not finish in the followers until we remove the repl_dev. So it is tested.

src/lib/replication/repl_dev/raft_repl_dev.h
hkadayam: It is initialized as part of the metablk created time, in RaftReplDev::RaftReplDev() and they are written whenever the cp is flushed.

src/include/homestore/replication_service.hpp
```diff
@@ -35,6 +35,12 @@ class ReplicationService {
     virtual AsyncReplResult< shared< ReplDev > > create_repl_dev(group_id_t group_id,
                                                                  std::set< replica_id_t > const& members) = 0;
 
+    /// @brief Removes the entire Repl Device and completely the replica device. While the underlying replica group is
```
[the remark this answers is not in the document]
hkadayam: Done

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -264,6 +306,11 @@ void RaftReplService::load_repl_dev(sisl::byte_view const& buf, void* meta_cooki
         return;
     }
```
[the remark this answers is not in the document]
hkadayam: Do you mean use `repl_dev->is_destroy_pending()` instead of the check `destroy_pending == 0x1`. That is not possible because repl_dev is not created yet, we need to create repl_dev based on rd_sb value.

src/lib/replication/repl_dev/raft_repl_dev.cpp
hkadayam: rreq can be nullptr on cases where this is an internal raft entry. end_of_append_batch contains lsns which are user and raft config as well. Having said that I changed that in end_of_append_batch to not put that requests into vector, so notify_after_data_written will not contain null rreq_ptrs.

src/lib/replication/service/raft_repl_service.cpp
yamingk: Can we put a comment above the if by saying `// Although we have a global timer, we want precise control on each individual rdev. `, just for readability?

src/lib/replication/service/raft_repl_service.cpp
yamingk: as long as they persist that information in their metablk "before returning on_destroy()", correct? 
We better make return type as future and homestore shouldn't proceed without a promise is made. Otherwise, there will be resource leak. If we leave it as return void, it has to to written in API contract that all destroy has to be completed before returning to homestore.

src/lib/replication/repl_dev/raft_repl_dev.h
yamingk: I mean print `destroy_pending`

yamingk: In RaftReplDev constructor it is a std::move, right? https://github.com/eBay/HomeStore/blob/master/src/lib/replication/repl_dev/raft_repl_dev.cpp#L33, I don't see any zero in the superblk class `create` api, have I missed anything?

yamingk (approved):

PR #398, opened 2024-04-30
Remove logdev which are unopened during recovery.

sanebay:
Removing repl dev will remove the logdev. If logdev or log store are not opened during, they are marked as ununsed and deleted. Remove chunks from the journal descriptor. Remove the logdev metablk.
1. Fix a regression in device manager where vdev size was not zero. Use after move.
2. Add UT to delete the logdev and delete it if its unopened during recovery.
3. logdev can be called during delete of repldev. If repl dev is marked for delete, repldev wont open logdev's during recovery and those are deleted. We first release all chunks and then  metablks related to logdev.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -5,7 +5,7 @@ from conans import CMake
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.4.1"
+    version = "6.4.2"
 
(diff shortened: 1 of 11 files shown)


hkadayam (dismissed):
Looks good to me

shosseinimotlagh (approved):

PR #403, opened 2024-05-03
Crash recovery of BlkAllocator fixes

hkadayam:
* AppendBlkAllocator to implement alloc_on_cache and alloc_on_disk

* FixedBlkAllocator didn't implement pre-alloc during recovery, implemented that


src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -102,86 +80,68 @@ BlkAllocStatus AppendBlkAllocator::alloc(blk_count_t nblks, const blk_alloc_hint
...
-    out_bid = BlkId{m_last_append_offset, nblks, m_chunk_id};
-
-    auto cur_cp = hs()->cp_mgr().cp_guard();
-    m_last_append_offset += nblks;
-    m_freeable_nblks -= nblks;
-
-    // it is guaranteed that dirty buffer always contains updates of current_cp or next_cp, it will
-    // never get dirty buffer from across updates;
-    set_dirty_offset(cur_cp->id() % MAX_CP_COUNT);
+    out_bid = BlkId{m_last_append_offset.fetch_add(nblks), nblks, m_chunk_id};
```
JacksonYao287, line 83: should we mark `m_is_dirty` to true here?

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -102,86 +80,68 @@ BlkAllocStatus AppendBlkAllocator::alloc(blk_count_t nblks, const blk_alloc_hint
...
+}
 
-bool AppendBlkAllocator::is_blk_alloced_on_disk(BlkId const&, bool) const { return false; }
+BlkAllocStatus AppendBlkAllocator::alloc_on_cache(BlkId const& blkid) {
+    auto new_offset = blkid.blk_num() + blkid.blk_count();
+    auto cur_offset = m_last_append_offset.load();
+    do {
+        if (cur_offset >= new_offset) { break; } // Already allocated
+    } while (!m_last_append_offset.compare_exchange_weak(cur_offset, new_offset));
+    return BlkAllocStatus::SUCCESS;
```
JacksonYao287, line 101: 1 m_is_dirty.store(true)？ like my above comment

2 if a write request comes when replaying log (which is carried out in a background thread and will commit the blk), this chunk might be selected for this write . then  a new blk will be allocated from this chunk(`AppendBlkAllocator::alloc` is called.) 
if `AppendBlkAllocator::alloc` is called between line 97 and 98 , `m_last_append_offset` will be a new one, and thus `m_last_append_offset.compare_exchange_weak(cur_offset, new_offset)` will fail for ever, right?

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -190,7 +150,13 @@ blk_num_t AppendBlkAllocator::get_used_blks() const { return m_last_append_offse
 
 blk_num_t AppendBlkAllocator::get_freeable_nblks() const { return m_freeable_nblks; }
 
-blk_num_t AppendBlkAllocator::get_defrag_nblks() const { return get_freeable_nblks() - available_blks(); }
+blk_num_t AppendBlkAllocator::get_defrag_nblks() const { return get_freeable_nblks() + available_blks(); }
```
JacksonYao287, line 153: if I understand correctly, `get_defrag_nblks()` will return how many blks can be GC, so in your PR,  should just return `m_freeable_nblks`, right?

src/lib/blkalloc/append_blk_allocator.cpp
hkadayam: It is not dirtied yet, dirty is relevant only when we commit the blk. In other words, dirty phase is between committing the blk and persisting the blk in disk. So alloc_on_disk() has to be done dirtying it.

src/lib/blkalloc/append_blk_allocator.cpp
hkadayam: 1. As responded above.
2. `m_last_append_offset.compare_exchange_weak(cur_offset, new_offset)` will not forever fail, because compare_exchange_weak will set the cur_offset with whatever the m_last_append_offset value. This is done by the std::atomic::compare_exchange_weak to avoid additional atomic operation, reference: https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange

Atomically compares the [object representation](https://en.cppreference.com/w/cpp/language/object)(until C++20)[value representation](https://en.cppreference.com/w/cpp/language/object)(since C++20) of *this with that of expected. If those are bitwise-equal, replaces the former with desired (performs read-modify-write operation). Otherwise, loads the actual value stored in *this into expected (performs load operation).

src/lib/blkalloc/append_blk_allocator.cpp
hkadayam: I am not sure why we need get_freeable_blks() and get_defrag_nblks() return the same thing. I would expect get_defrag_blks() indicate how much I get after I defragment them, right?

yamingk: No, we shouldn't make this change. defrag_nblks should exclude available blocks, otherwise, if it is a chunk with nothing written on it yet, it will also be selected for GC, which is wrong. 

Defrag_nblks should only be the number of blocks, that by compacting the shard to a new place, that how many nblks could be freed, by nature it shouldn't include available blocks on this chunk. This is the reason why m_freeable_nblks is init to all available blks so that it can keep track of the defrag nblks.

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -30,7 +30,7 @@ AppendBlkAllocator::AppendBlkAllocator(const BlkAllocConfig& cfg, bool need_form
         nullptr);
 
     if (need_format) {
-        m_freeable_nblks = available_blks();
+        m_freeable_nblks = 0;
```
yamingk, line 33: see my comment below.

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -102,86 +80,68 @@ BlkAllocStatus AppendBlkAllocator::alloc(blk_count_t nblks, const blk_alloc_hint
...
+    // Freeing something in the middle, increment the count
+    m_freeable_nblks.fetch_add(nblks);
 }
 
+void AppendBlkAllocator::free_on_disk(BlkId const&) { m_is_dirty.store(true); }
+
 bool AppendBlkAllocator::is_blk_alloced(const BlkId& in_bid, bool) const {
     // blk_num starts from 0;
-    return in_bid.blk_num() < get_used_blks();
+    return in_bid.blk_num() < m_last_append_offset;
```
yamingk, line 134: Why not access from existing api (`get_used_nblks()`)? Isn't it better to avoid understanding the real meaning of `m_last_append_offset` that it actually means nblks, so that if in the future this field changes, only needs to update that get_used_blks(), not every where this fields is being access directly?

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -102,86 +80,68 @@ BlkAllocStatus AppendBlkAllocator::alloc(blk_count_t nblks, const blk_alloc_hint
...
+    auto const nblks = bid.blk_count();
+    auto exp_last_offset = bid.blk_num() + nblks;
+    auto const new_offset = m_last_append_offset - nblks;
+    if (m_last_append_offset.compare_exchange_strong(exp_last_offset, new_offset)) { return; }
+
+    // Freeing something in the middle, increment the count
+    m_freeable_nblks.fetch_add(nblks);
 }
 
+void AppendBlkAllocator::free_on_disk(BlkId const&) { m_is_dirty.store(true); }
```
yamingk, line 130: it will cause unnecessary dirty of the sb, right? It is not dirtying anything actually.

src/lib/blkalloc/append_blk_allocator.h
```diff
@@ -84,7 +78,7 @@ class AppendBlkAllocator : public BlkAllocator {
     BlkAllocStatus alloc(blk_count_t nblks, blk_alloc_hints const& hints, BlkId& out_blkid) override;
     void free(BlkId const& b) override;
     BlkAllocStatus alloc_on_disk(BlkId const& in_bid) override;
-    BlkAllocStatus mark_blk_allocated(BlkId const& b) override;
+    BlkAllocStatus alloc_on_cache(BlkId const& b) override;
```
yamingk, line 81: It looks like with this PR, alloc_on_cache and alloc is the same.

src/lib/blkalloc/append_blk_allocator.cpp
hkadayam: I generally prefer to use member variables instead of methods to make code easier to read and provide apis only for external consumption. If you want me to revert that, I will do that.

src/lib/blkalloc/append_blk_allocator.h
hkadayam: alloc_on_cache is only during recovery, because alloc() is done during regular operation, which maintains information in the cache, but during recovery, whatever was committed needs to be put back in cache, at that time, alloc_on_cache() is called internally by homestore.

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -102,86 +80,68 @@ BlkAllocStatus AppendBlkAllocator::alloc(blk_count_t nblks, const blk_alloc_hint
...
-// cp_flush should not block alloc/free;
-//
 void AppendBlkAllocator::cp_flush(CP* cp) {
-    const auto idx = cp->id() % MAX_CP_COUNT;
     // check if current cp's context has dirty buffer already
-    if (m_dirty_sb[idx].is_dirty) {
-        m_sb->last_append_offset = m_dirty_sb[idx].last_append_offset;
-        m_sb->freeable_nblks = m_dirty_sb[idx].freeable_nblks;
+    if (m_is_dirty.exchange(false)) {
+        m_sb->last_append_offset = m_last_append_offset.load();
```
yamingk, line 107: It is okay to do so by removing the m_dirty_sb, but it will take the latest m_last_append_offset that belongs to next CP, which is fine, that we need to do some check on alloc_on_cache during recovery (pls see my other comment).

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -102,86 +80,68 @@ BlkAllocStatus AppendBlkAllocator::alloc(blk_count_t nblks, const blk_alloc_hint
...
+    return BlkAllocStatus::SUCCESS;
+}
 
-bool AppendBlkAllocator::is_blk_alloced_on_disk(BlkId const&, bool) const { return false; }
+BlkAllocStatus AppendBlkAllocator::alloc_on_cache(BlkId const& blkid) {
+    auto new_offset = blkid.blk_num() + blkid.blk_count();
+    auto cur_offset = m_last_append_offset.load();
+    do {
+        if (cur_offset >= new_offset) { break; } // Already allocated
+    } while (!m_last_append_offset.compare_exchange_weak(cur_offset, new_offset));
```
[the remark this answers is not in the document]
hkadayam: I think there is some confusion with the name alloc_on_cache. It will be called only during recovery, along with other comments, let me rename it to something more meaningful something like `alloc_on_recovery()`

src/lib/blkalloc/append_blk_allocator.cpp
JacksonYao287: I understand your point. but I thinks that is only right to raft_repl_dev. for raft_repl_dev, when applying log in background thread, `on_commit` will be called and thus `data_service().commit_blk` and `alloc_on_disk()`, this is  the scenario you mentioned here.

but , data_service can be used independently , if we use data_service with append_block_allocator(not in raft_repl_dev like nublox), then we will only call `async_allocate_write` , and `commit_blk` will not be called in this case, right? so the append_block_allocator will not be marked as dirty and cp flush will do nothing , which will lead to data loss.

src/lib/blkalloc/append_blk_allocator.cpp
JacksonYao287: thanks for the explanation of `compare_exchange_weak`, I do misunderstand it, my bad.
but this risk I mentioned above is still possible. I believe  we do not have any mechanism to keep `append_blk_allocator` from handling `allocate_blk` reqeust before all the logs are replayed. right?

for example , if we get homestore restarted and replaying log , when reaching line 97, we get an m_last_append_offset of 100. if an alloc request comes at this moment, maybe m_last_append_offset will be set to 101. but 101 might be already allocated to other blk before restarting. that`s the case i am saying.

so does homestore start handling write request only until all the logs are replyed? this is my point

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -102,86 +80,68 @@ BlkAllocStatus AppendBlkAllocator::alloc(blk_count_t nblks, const blk_alloc_hint
...
-        m_last_append_offset -= n;
-    }
-    set_dirty_offset(cur_cp->id() % MAX_CP_COUNT);
+    auto const nblks = bid.blk_count();
+    auto exp_last_offset = bid.blk_num() + nblks;
+    auto const new_offset = m_last_append_offset - nblks;
+    if (m_last_append_offset.compare_exchange_strong(exp_last_offset, new_offset)) { return; }
+
+    // Freeing something in the middle, increment the count
+    m_freeable_nblks.fetch_add(nblks);
```
sanebay, line 127: irrespective of its last or not, shouldnt we increase the count.

src/lib/blkalloc/fixed_blk_allocator.cpp
```diff
@@ -55,27 +54,54 @@ blk_num_t FixedBlkAllocator::init_portion(BlkAllocPortion& portion, blk_num_t st
 bool FixedBlkAllocator::is_blk_alloced(BlkId const& b, bool use_lock) const { return true; }
 
 BlkAllocStatus FixedBlkAllocator::alloc([[maybe_unused]] blk_count_t nblks, blk_alloc_hints const&, BlkId& out_blkid) {
-    HS_DBG_ASSERT_EQ(nblks, 1, "FixedBlkAllocator does not support multiple blk allocation yet");
-    return alloc_contiguous(r_cast< BlkId& >(out_blkid));
-}
+    if (m_state == state_t::RECOVERING) {
```
sanebay, line 57: nit: could we move this to init_portion during recovery ?

src/lib/blkalloc/append_blk_allocator.cpp
hkadayam: The data service expects one to use journal service or maintain their own journal. There is no way for data service can keep track of pending commits. If one were to use Data service alone without other service, then it can be done with no crash protection.

src/lib/blkalloc/append_blk_allocator.cpp
hkadayam: Yes homestore start will not be available to accept writes or follower commits until we finish replay.

src/lib/blkalloc/append_blk_allocator.cpp
JacksonYao287: ok , if this is the design goal and the contract between homestore and the related user, it makes sense. got it ! thanks

src/lib/blkalloc/append_blk_allocator.cpp
JacksonYao287: how to achieve this goal?
if that , reft_repl_service#start() should return only until all the repl_dev has replayed all the logs to the latest commit index. this way, it can block sm from accepting new writes.

can you point out where this part of code is?

hkadayam: In RaftReplService::start, we load all repl_devs here
https://github.com/eBay/HomeStore/blob/master/src/lib/replication/service/raft_repl_service.cpp#L106

Then we start logstore_service.start() which replays all logs (which at this point repl_dev would replay till checkpoint lsn)

Then only we join the raft_group. After raft_group has been joined raft will commit things in lsn order only. So there is never a chance of out-of-order commit or action.

src/lib/blkalloc/append_blk_allocator.cpp
hkadayam: Yes that is correct.

src/lib/blkalloc/fixed_blk_allocator.cpp
hkadayam: Problem is there is no way to know when the recovery is completed, that is why we are doing on the first alloc.

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -102,86 +80,68 @@ BlkAllocStatus AppendBlkAllocator::alloc(blk_count_t nblks, const blk_alloc_hint
...
-    m_freeable_nblks += n;
-    if (bid.blk_num() + n == m_last_append_offset) {
-        // we are freeing the the last blk id, let's rewind.
-        m_last_append_offset -= n;
-    }
-    set_dirty_offset(cur_cp->id() % MAX_CP_COUNT);
+    auto const nblks = bid.blk_count();
+    auto exp_last_offset = bid.blk_num() + nblks;
+    auto const new_offset = m_last_append_offset - nblks;
+    if (m_last_append_offset.compare_exchange_strong(exp_last_offset, new_offset)) { return; }
```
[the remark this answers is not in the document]
hkadayam: First off @xiaoxichen this is a super catch. Intuitively felt something wrong with that code because of anti-pattern, but went ahead with it, without giving too much thought.

We have 2 choices:
Choice 1: We can fix this with following code:
```
// If we are freeing the last block, just move the offset back
blk_num_t cur_last_offset = m_last_append_offset.load();
auto const input_last_offset = bid.blk_num() + bid.blk_count();
blk_num_t new_offset;
bool freeing_in_middle{false};
do {
    if (input_last_offset == cur_last_offset) {
        new_offset = input_last_offset;
         freeing_in_middle = false;
    } else {
         new_offset = cur_last_offset;
         freeing_in_middle = true;    
    }
} while (!m_last_append_offset.compare_exchange_weak(cur_last_offset, new_offset));

if (freeing_in_middle) {
     // Freeing something in the middle, increment the count
    m_freeable_nblks.fetch_add(bid.blk_count());
}
```

This code piece ensures all conditions for example: 
```say m_last_append_offset = 10 
Thread 1 frees, blkid=8, count=1, while inside the do while, 
Thread 2 frees  blkid=9, count=1, and moved the m_last_append_offset = 9
Now thread 1 can go back and move the m_last_append_offset = 8
```
which is the correct thing to do, except there is always one additional atomic operation.

Choice 2:
Just treat any free as holes and never try to bring back m_last_append_offset.

Let me update the PR with choice 1 and we shall discuss this with implications a bit and can always remove this code.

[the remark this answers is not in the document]
xiaoxichen: will 
`auto const new_offset = bid.blk_num()` works?  it avoid the pattern of read + modify an atomic

src/lib/blkalloc/append_blk_allocator.cpp
hkadayam: Not really, our goal is to find the fragmented blks. We will have another api to get_available_blks() or something to get sum of this.

src/lib/blkalloc/append_blk_allocator.cpp
hkadayam: Agreed and updated with latest changes, please rereview them.

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -102,95 +79,111 @@ BlkAllocStatus AppendBlkAllocator::alloc(blk_count_t nblks, const blk_alloc_hint
...
 void AppendBlkAllocator::free(const BlkId& bid) {
-    std::unique_lock lk(m_mtx);
-    auto cur_cp = hs()->cp_mgr().cp_guard();
-    const auto n = bid.blk_count();
-    m_freeable_nblks += n;
-    if (bid.blk_num() + n == m_last_append_offset) {
-        // we are freeing the the last blk id, let's rewind.
-        m_last_append_offset -= n;
+    // If we are freeing the last block, just move the offset back
+    // If we are freeing the last block, just move the offset back
```
yamingk, line 136: duplicate comment, I guess you mean if we are freeing in the middle, update m_freeable_nblks

src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -102,95 +79,111 @@ BlkAllocStatus AppendBlkAllocator::alloc(blk_count_t nblks, const blk_alloc_hint
...
 }
 
-BlkAllocStatus AppendBlkAllocator::alloc_on_disk(BlkId const&) { return BlkAllocStatus::SUCCESS; }
-
-BlkAllocStatus AppendBlkAllocator::mark_blk_allocated(BlkId const&) { return BlkAllocStatus::SUCCESS; }
-
-void AppendBlkAllocator::free_on_disk(BlkId const&) {}
+// Reserve on disk will update the commit_offset with the new_offset, if its above the current commit_offset.
+BlkAllocStatus AppendBlkAllocator::reserve_on_disk(BlkId const& blkid) {
+    auto cpg = hs()->cp_mgr().cp_guard();
```
yamingk, line 91: we need an release assert commit_offset should not across m_last_append_offset. 
This is for self-defense that caller is messing up with blkallocator, e.g. only called reserver_on_disk, not calling reserve_on_cache

src/lib/blkalloc/append_blk_allocator.h
```diff
@@ -137,12 +118,12 @@ class AppendBlkAllocator : public BlkAllocator {
     void on_meta_blk_found(const sisl::byte_view& buf, void* meta_cookie);
 
 private:
-    std::mutex m_mtx;                  // thread_safe, TODO: open option for consumer to choose to go lockless;
-    blk_num_t m_last_append_offset{0}; // last appended offset in blocks;
-    blk_num_t m_freeable_nblks{0};
+    std::atomic< blk_num_t > m_last_append_offset{0}; // last appended offset in blocks;
```
yamingk, line 121: Maybe put some comment say, 
`m_last_append_offset` is in memory only.
`m_commit_offset` is the persistence version (going to be persisted) in next cp.

yamingk (approved):

src/lib/blkalloc/append_blk_allocator.cpp
hkadayam: Done

src/lib/blkalloc/append_blk_allocator.h
hkadayam: Done

src/lib/blkalloc/append_blk_allocator.cpp
JacksonYao287: sorry , i do not get your point here. logstore_service.start() will only open devs , right?  log replay should involve the statemachine of repl_dev, which will apply the change in log entry.

so can you point out the code where is this part ? i am a little confused here

Issue #400, opened 2024-05-02
handle clean degrade-mode


JacksonYao287:
Testing and bug-fixing the degrade-mode(clean, bad drive removed) ---- running with N-1 data drive.   This is a big task as I believe we will have a lot of assert/panic here and there, like index has the blobid->pba mapping but the drive is gone.

xiaoxichen:
please do not close this till this get verify end to end. Better we can test this via UT

PR #405, opened 2024-05-06
handle missing pdev when reading

JacksonYao287:
#400 
in some scenarios, homestore might be restarted with missing pdev(for example, breakfix), thus all the chunks in that pdev will  not be loaded. so  we need to deal with this case.

1 for read from client, we need to check whether the chunk in the provided blkid is loaded and proceed accordingly.
2 for write ,  we will almost not encounter this case since we will never allocate blk for a missing pdev. but for the sake of completeness, check is also added .

### DIFF
--- a/src/lib/device/device.h
+++ b/src/lib/device/device.h
@@ -159,15 +159,15 @@ public:
 
     const Chunk* get_chunk(uint32_t chunk_id) const {
         std::unique_lock lg{m_vdev_mutex};
-        // if a pdev is misssing when restart, chunk_id from cleint might be larger than m_chunks.size()
-        if (chunk_id == INVALID_CHUNK_ID || chunk_id >= m_chunks.size()) return nullptr;
+        // if a pdev is misssing when restart, chunk_id from client might be larger than m_chunks.size()
+        if (!m_chunks.index_exists(chunk_id)) return nullptr;
         return m_chunks[chunk_id].get();
     }
 
     Chunk* get_chunk_mutable(uint32_t chunk_id) {
         std::unique_lock lg{m_vdev_mutex};
         // if a pdev is misssing when restart, chunk_id from client might be larger than m_chunks.size()
-        if (chunk_id == INVALID_CHUNK_ID || chunk_id >= m_chunks.size()) return nullptr;
+        if (!m_chunks.index_exists(chunk_id)) return nullptr;
         return m_chunks[chunk_id].get();
     }
 
--- a/src/lib/device/device.h
+++ b/src/lib/device/device.h
@@ -159,12 +159,16 @@ public:
 
     const Chunk* get_chunk(uint32_t chunk_id) const {
         std::unique_lock lg{m_vdev_mutex};
-        return (chunk_id == INVALID_CHUNK_ID) ? nullptr : m_chunks[chunk_id].get();
+        // if a pdev is misssing when restart, chunk_id from cleint might be larger than m_chunks.size()
+        if (chunk_id == INVALID_CHUNK_ID || chunk_id >= m_chunks.size()) return nullptr;
+        return m_chunks[chunk_id].get();
     }
 
     Chunk* get_chunk_mutable(uint32_t chunk_id) {
         std::unique_lock lg{m_vdev_mutex};
-        return (chunk_id == INVALID_CHUNK_ID) ? nullptr : m_chunks[chunk_id].get();
+        // if a pdev is misssing when restart, chunk_id from client might be larger than m_chunks.size()
+        if (chunk_id == INVALID_CHUNK_ID || chunk_id >= m_chunks.size()) return nullptr;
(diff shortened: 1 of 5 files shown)


src/lib/device/device.h
```diff
@@ -159,12 +159,16 @@ class DeviceManager {
 
     const Chunk* get_chunk(uint32_t chunk_id) const {
         std::unique_lock lg{m_vdev_mutex};
-        return (chunk_id == INVALID_CHUNK_ID) ? nullptr : m_chunks[chunk_id].get();
+        // if a pdev is misssing when restart, chunk_id from cleint might be larger than m_chunks.size()
+        if (chunk_id == INVALID_CHUNK_ID || chunk_id >= m_chunks.size()) return nullptr;
```
xiaoxichen, line 163: we probabaly need to change the `m_chunks` to a map, or we need to insert dummy entries.

As the missing chunks can be in the middle , not the tail.  E.G.  chunk 0/1 from disk 1, 2/3 from disk 2, 4/5 from disk 3, 6/7 from disk 4

we might lost disk 2,  in your current implementation ,    we probably get 4/5 when requesting 2/3

JacksonYao287: m_chunks is actually sparse vector, which will insert dummy element up to the index. it means if we have a sparse vector with a size of 10  and we use sparse_vector[20] to insert a new element, it will fill up [10-19] with elements by the default constructor.
```
sisl::sparse_vector< shared< Chunk > > m_chunks; 
```
pls refer to https://github.com/eBay/sisl/blob/8808be84afd8a967f83feb437e3652a87bf13b46/include/sisl/fds/sparse_vector.hpp#L62

src/lib/device/device.h
```diff
@@ -159,12 +159,16 @@ class DeviceManager {
...
+        return m_chunks[chunk_id].get();
```
xiaoxichen, line 164: can we just `return get_chunk_mutable(chunk_id);`" for better readability ?

src/lib/device/device.h
```diff
@@ -159,12 +159,16 @@ class DeviceManager {
...
     }
 
     Chunk* get_chunk_mutable(uint32_t chunk_id) {
         std::unique_lock lg{m_vdev_mutex};
-        return (chunk_id == INVALID_CHUNK_ID) ? nullptr : m_chunks[chunk_id].get();
+        // if a pdev is misssing when restart, chunk_id from client might be larger than m_chunks.size()
+        if (chunk_id == INVALID_CHUNK_ID || chunk_id >= m_chunks.size()) return nullptr;
```
xiaoxichen, line 170: maybe `m_chunks.index_exists(chunk_id)`?

src/lib/device/virtual_dev.hpp
```diff
@@ -299,6 +299,7 @@ class VirtualDev {
 
 private:
     uint64_t to_dev_offset(BlkId const& b, Chunk** chunk) const;
+    bool is_chunk_loaded(cshared< Chunk >& chunk) const;
```
xiaoxichen, line 302: maybe is_chunk_exists or is_chunk_available is a better naming?  loaded sounds like if not you can load again:)

xiaoxichen (approved):
lgtm, a few nit FYI

src/lib/device/virtual_dev.cpp
```diff
@@ -600,8 +648,12 @@ int VirtualDev::cp_progress_percent() { return 100; }
 ///////////////////////// VirtualDev Private Methods /////////////////////////////
 uint64_t VirtualDev::to_dev_offset(BlkId const& b, Chunk** chunk) const {
     *chunk = m_dmgr.get_chunk_mutable(b.chunk_num());
-    RELEASE_ASSERT(*chunk, "Chunk got null {}", b.chunk_num());
+    if (!(*chunk)) return INVALID_DEV_OFFSET;
     return uint64_cast(b.blk_num()) * block_size() + uint64_cast((*chunk)->start_offset());
 }
 
+bool VirtualDev::is_chunk_loaded(cshared< Chunk >& chunk) const {
```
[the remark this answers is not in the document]
JacksonYao287: i wanted to set the parameter to only chunk_id. but if that, we need to get the chunk_id before 
```
if (sisl_unlikely(!is_chunk_available(chunk))) 
```
so , puttingthe action of getting chunk id inside `is_chunk_available` itself  seems more clean.
since it generates no impact to the logic, we can keep it like this and change it if necessary in the future.

src/lib/device/device.h
JacksonYao287: we can not call `get_chunk_mutable(chunk_id)` inside `get_chunk` since `get_chunk` is a `const func() const` but `get_chunk_mutable` is non-const

JacksonYao287:
@xiaoxichen I address your comment, PTAL

xiaoxichen (approved):

PR #413, opened 2024-05-10
fix logstore recovery issue

JacksonYao287:
the recovery issue is caused by the logdev metablk recovery callback `rollback_super_blk_found`, it will put all the log dev to `m_unopened_logdev` , which will be destroyed when logservice start.

in this pr :

1 change the logic about `m_unopened_logdev` to make sure only those log devs which are not opened before logservice#start will be destroyed.

2 change the metablk dependency , so that `rollback_super_blk_found` will be called after all the log dev metablk are recovered

3 add logic to make sure the logstore of a repldev is ready before it is going to join raft group. now , we use a future to open logstore, so there might be a small case that logstore is not ready when repldev is joining raft group

### DIFF
--- a/src/lib/logstore/log_store_service.cpp
+++ b/src/lib/logstore/log_store_service.cpp
@@ -237,13 +237,9 @@ void LogStoreService::rollback_super_blk_found(const sisl::byte_view& buf, void*
         auto id = rollback_sb->logdev_id;
         LOGDEBUGMOD(logstore, "Log dev rollback superblk found logdev={}", id);
         const auto it = m_id_logdev_map.find(id);
-        if (it != m_id_logdev_map.end()) {
-            logdev = it->second;
-        } else {
-            logdev = std::make_shared< LogDev >(id, m_logdev_vdev.get());
-            m_id_logdev_map.emplace(id, logdev);
-        }
-
+        HS_REL_ASSERT((it != m_id_logdev_map.end()),
+                      "found a rollback_super_blk of logdev id {}, but the logdev with id {} doesnt exist", id);
+        logdev = it->second;
         logdev->log_dev_meta().rollback_super_blk_found(buf, meta_cookie);
     }
 }
--- a/src/lib/logstore/log_store_service.cpp
+++ b/src/lib/logstore/log_store_service.cpp
@@ -51,7 +51,7 @@ LogStoreService::LogStoreService() {
         [this](meta_blk* mblk, sisl::byte_view buf, size_t size) {
             rollback_super_blk_found(std::move(buf), voidptr_cast(mblk));
         },
-        nullptr);
+        nullptr, true, std::optional< meta_subtype_vec_t >({logdev_sb_meta_name}));
 }
 
 folly::Future< std::error_code > LogStoreService::create_vdev(uint64_t size, HSDevType devType, uint32_t chunk_size) {
@@ -175,7 +175,9 @@ void LogStoreService::open_logdev(logdev_id_t logdev_id) {
         m_id_reserver->reserve(logdev_id);
         auto logdev = std::make_shared< LogDev >(logdev_id, m_logdev_vdev.get());
         m_id_logdev_map.emplace(logdev_id, logdev);
+        LOGDEBUGMOD(logstore, "log_dev={} does not exist, created!", logdev_id);
     }
+    m_unopened_logdev.erase(logdev_id);
     LOGDEBUGMOD(logstore, "Opened log_dev={}", logdev_id);
 }
 
@@ -206,17 +208,17 @@ void LogStoreService::logdev_super_blk_found(const sisl::byte_view& buf, void* m
         auto id = sb->logdev_id;
         LOGDEBUGMOD(logstore, "Log dev superblk found logdev={}", id);
         const auto it = m_id_logdev_map.find(id);
-        if (it == m_id_logdev_map.end()) {
-            LOGERROR("logdev={} found but not opened yet, it will be discarded after logstore is started", id);
-            m_unopened_logdev.insert(id);
-        }
-
         // We could update the logdev map either with logdev or rollback superblks found callbacks.
         if (it != m_id_logdev_map.end()) {
             logdev = it->second;
         } else {
             logdev = std::make_shared< LogDev >(id, m_logdev_vdev.get());
             m_id_logdev_map.emplace(id, logdev);
+            // when recover logdev meta blk, we get all the logdevs from the superblk. we put them in m_unopened_logdev
+            // too. after logdev meta blks are all recovered, when a client opens a logdev, we remove it from
+            // m_unopened_logdev. so that when we start log service, all the left items in m_unopened_logdev are those
+            // not open, which can be destroyed
+            m_unopened_logdev.insert(id);
         }
 
         logdev->log_dev_meta().logdev_super_blk_found(buf, meta_cookie);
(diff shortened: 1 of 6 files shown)


src/lib/logstore/log_store_service.cpp
```diff
@@ -235,11 +237,6 @@ void LogStoreService::rollback_super_blk_found(const sisl::byte_view& buf, void*
...
         LOGDEBUGMOD(logstore, "Log dev rollback superblk found logdev={}", id);
         const auto it = m_id_logdev_map.find(id);
-        if (it == m_id_logdev_map.end()) {
-            LOGERROR("logdev={} found but not opened yet, it will be discarded after logstore is started", id);
-            m_unopened_logdev.insert(id);
-        }
-
         if (it != m_id_logdev_map.end()) {
             logdev = it->second;
         } else {
```
xiaoxichen, line 242: I am not sure whether it is valid to have a `rollback_superblk` but do not have a `logdev_superblock`  ?  
If not, maybe we dont need the L239-L245?
If yes,  what is the case?   do we need to add the logdev into `m_unopened_logdev` as well? 

cc @sanebay

xiaoxichen (commented):
Do we know why this UT doesnt catch the issue?  

https://github.com/eBay/HomeStore/pull/398/files#diff-ffe5ad33fc8c3ce2cb52620e0eaa2fa4f6243a12f476ced23a85f4d752adf2a6R432-R437

src/lib/logstore/log_store_service.cpp
```diff
@@ -175,7 +175,9 @@ void LogStoreService::open_logdev(logdev_id_t logdev_id) {
         m_id_reserver->reserve(logdev_id);
         auto logdev = std::make_shared< LogDev >(logdev_id, m_logdev_vdev.get());
         m_id_logdev_map.emplace(logdev_id, logdev);
+        LOGDEBUGMOD(logstore, "log_dev={} does not exist, created!", logdev_id);
```
sanebay, line 178: We already clear it in line 160

src/lib/logstore/log_store_service.cpp
sanebay: Yes dont understand the changes done in the logdev. If we find ether logdev or rollback superblk, we create logdev structure. If its nt opened, we delete them, thats a feature.

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -104,11 +104,13 @@ HomeRaftLogStore::HomeRaftLogStore(logdev_id_t logdev_id, logstore_id_t logstore
         m_logstore_id = logstore_id;
         LOGDEBUGMOD(replication, "Opening existing home log_dev={} log_store={}", m_logdev_id, logstore_id);
         logstore_service().open_logdev(m_logdev_id);
-        logstore_service().open_log_store(m_logdev_id, logstore_id, true).thenValue([this](auto log_store) {
-            m_log_store = std::move(log_store);
-            DEBUG_ASSERT_EQ(m_logstore_id, m_log_store->get_store_id(), "Mismatch in passed and create logstore id");
-            REPL_STORE_LOG(DEBUG, "Home Log store created/opened successfully");
-        });
+        m_log_store_future =
```
sanebay, line 107: this part LGTM.

sanebay (changes requested):

src/lib/logstore/log_store_service.cpp
xiaoxichen: @sanebay can you point out where we remove the logdev from `m_unopened_logdev` if it was opened?

src/lib/logstore/log_store_service.cpp
JacksonYao287: yes, but that is the moment we start logservice, right? 
we open the logdev before we start logservice, so according to the logic here, it will crerate a new one, right?

src/lib/logstore/log_store_service.cpp
JacksonYao287: @sanebay 
if i understand correctly,  a rollback superblk aims to  record the rollback info of a logdev, right? a rollback superblk in only meaningful when it has an corresponding log dev instance.
so if we find a rollback superblk without the corresponding log dev, that should be an error.

in fact, i wanted to add and assert in line 246 to confirm each rollback superblk has a an corresponding log dev.

JacksonYao287: > @sanebay can you point out where we remove the logdev from `m_unopened_logdev` if it was opened?
https://github.com/eBay/HomeStore/blob/78aead7920e60beaef29fa5f049177b3f03dadd9/src/lib/logstore/log_store_service.cpp#L100
PTAL!

xiaoxichen: @JacksonYao287  I am asking, in previous code, where we remove OPENED logdev from `m_unopened_logdev`...

JacksonYao287: I believe the position i mentioned above is the only place to remove element from m_unopened_logdev.

src/lib/logstore/log_store_service.cpp
sanebay: I really dont get the issue you are tyring to fix. Wont the changes in home_raft_log_store.cpp fix the actual issue ?

JacksonYao287: no, the changes in home_raft_log_store.cpp is to prevent the repl_dev from joining raft group before the repl_log_store is ready. this is one changes.

the direct cause of the recovery issue is we do not remove open logdevs from m_unopened_logdev correctly. other changes in the PR is to resolve this.

src/lib/logstore/log_store_service.cpp
```diff
@@ -206,17 +208,17 @@ void LogStoreService::logdev_super_blk_found(const sisl::byte_view& buf, void* m
...
         if (it != m_id_logdev_map.end()) {
             logdev = it->second;
         } else {
             logdev = std::make_shared< LogDev >(id, m_logdev_vdev.get());
             m_id_logdev_map.emplace(id, logdev);
+            // when recover logdev meta blk, we get all the logdevs from the superblk. we put them in m_unopened_logdev
+            // too. after logdev meta blks are all recovered, when a client opens a logdev, we remove it from
+            // m_unopened_logdev. so that when we start log service, all the left items in m_unopened_logdev are those
+            // not open, which can be destroyed
+            m_unopened_logdev.insert(id);
```
sanebay, line 221: You will have to add the same in rollback_super_blk_found too.

JacksonYao287: @sanebay 
pls take a look at line 54 in this file
```
    meta_service().register_handler(
        logdev_rollback_sb_meta_name,
        [this](meta_blk* mblk, sisl::byte_view buf, size_t size) {
            rollback_super_blk_found(std::move(buf), voidptr_cast(mblk));
        },
        nullptr, true, std::optional< meta_subtype_vec_t >({logdev_sb_meta_name}));
```

this will make sure all the `rollback_super_blk_found` will be called after all the `logdev_super_blk_found` is called . order is guaranteed here.

sanebay (commented):
LGTM

src/lib/logstore/log_store_service.cpp
xiaoxichen: with the ordering we enforced, do we still have the `else` clause ? 

maybe it is an assert?

JacksonYao287: yes, that what I wanted, I believe we need assert here. pls take a look at my above comment.
>in fact, i wanted to add an assert in line 246 to confirm each rollback superblk has a an corresponding log dev.

xiaoxichen (approved):
LGTM.

If you can add some comment before the asset in L240 as well as in L54 to explain how we ensure the ordering, it would be even better.

Issue #401, opened 2024-05-02
handle unclean degrade-mode


JacksonYao287:
Testing and bug-fixing the degrade-mode (unclean, bad drive still present to SM but stucking/returning eio).  we might need to have ut to test that with flip.

JacksonYao287:
#416

PR #416, opened 2024-05-13
inject flip into io path to simulate io error and stuck io

JacksonYao287:
this PR inject flip point to simulate io_error and stuck io in read/write path , which will be used for later long running test
#401

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -5,7 +5,7 @@ from conans import CMake
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.4.6"
+    version = "6.4.7"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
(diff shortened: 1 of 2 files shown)


xiaoxichen (commented):
can we move the 
```
    if (iomgr_flip::instance()->test_flip("blk_write_failure")) {
        return folly::makeFuture< std::error_code >(std::make_error_code(std::errc::io_error));
    }
    if (iomgr_flip::instance()->test_flip("blk_write_stuck")) {
        std::this_thread::sleep_for(std::chrono::hours(24 * 365 * 100)); // Sleep for 100 years
        return folly::makeFuture< std::error_code >(std::make_error_code(std::errc::resource_unavailable_try_again));
    }
```
into  a function ?

JacksonYao287:
> can we move the
> 
> ```
>     if (iomgr_flip::instance()->test_flip("blk_write_failure")) {
>         return folly::makeFuture< std::error_code >(std::make_error_code(std::errc::io_error));
>     }
>     if (iomgr_flip::instance()->test_flip("blk_write_stuck")) {
>         std::this_thread::sleep_for(std::chrono::hours(24 * 365 * 100)); // Sleep for 100 years
>         return folly::makeFuture< std::error_code >(std::make_error_code(std::errc::resource_unavailable_try_again));
>     }
> ```
> 
> into a function ?

will wrap them into a macro

xiaoxichen (commented):
my personal taste prefer inline compared to macro, but sort of preference's territory. 

LGTM

JacksonYao287:
> my personal taste prefer inline compared to macro, but sort of preference's territory.
> 
> LGTM

i think it does not matter to use macro or inline for the case here, since it will not bring too much impact. so can you give an approval?

xiaoxichen (approved):

Issue #379, opened 2024-04-11
Crash recovery for logstore and journal vdev


sanebay:
1. Find the area's in journal vdev where we need crash recovery points.
2. Implmenet those flip points.

1. Chunk get from and release from the pool. Multiple head chunks.
2. Remove log store, log dev. 
3. Flush of log group 
4. LogDevMeta resever_store_id, truncate() just before persist.
5. During truncate, release of chunks. 
6. We also need to add flip point when we add new chunks to the journal_vdev, in which we do persistence of the private data and think it through what will happen on reboot and verify with a test case.

PR #419, opened 2024-05-13
Add long running log store tests.

sanebay:
Changes.
1. Fix the truncation in corner cases.
2. Add long running test for logstore. Stripped down version of test_log_store. Add burst of requests to all logstores, truncate, restart, add/remove logstores logdev's, ad rollbacks etc.
3. Enable logstore test except the parallel truncation UT.

More details.
Journalvdev maintains a list of chunks to store all the log entries. All append log entries are appended to the last chunk in the list(right side/tail offset) and truncate is applied to the head of the chunk list(left part/ data start offset). Whenever we append log entries, if we dont have enough space, we create a new chunk and append to the list. So log groups(batch of log entries) dont go across chunks. So there will be holes in these chunks at the end which are marked by end_of_chunk in chunk private data. The hole lies between end_of_chunk and (chunk_start + chunk_size) . When we read and we reach this hole, there is no data so we skip and move to the next chunk. Similarly if truncate happens to be in that hole also, we release the whole chunk and move to the next chunk. Also we set the data_start_offset to the start of the next chunk.


yamingk:
I understand we have went through the truncate corner cases during a small group of meeting, please also describe what is the corner case in truncate that is being fixed in this PR, it will help others also to review and understand this change if they want to.

src/lib/device/journal_vdev.cpp
```diff
@@ -559,54 +574,108 @@ void JournalVirtualDev::Descriptor::truncate(off_t truncate_offset) {
...
+    // timestamp and cleanup the other. We check if the truncation offset happens to be
+    // between the end_of_chunk mark and actual chunk end. In that case there is not data
+    // in that chunk and we release this chunk and select the next chunk.
+    bool update_truncate_offset = false;
+    auto [new_head_chunk, index, offset_in_chunk] = offset_to_chunk(truncate_offset);
+    auto end_new_head_chunk = m_vdev.get_end_of_chunk(new_head_chunk);
+    if (offset_in_chunk >= static_cast< int64_t >(end_new_head_chunk)) {
+        // If the truncation offset is same as end_of_chunk, we dont have any more
+        // data in this chunk to read. Select the next chunk.
+        if (++index <= (m_journal_chunks.size() - 1)) {
```
yamingk, line 589: what will be consequence for the `else` part of this if?

src/lib/device/journal_vdev.cpp
```diff
@@ -45,15 +45,15 @@ JournalVirtualDev::JournalVirtualDev(DeviceManager& dmgr, const vdev_info& vinfo
...
-            [this]() {
-                m_init_private_data->created_at = get_time_since_epoch_ms();
-                m_init_private_data->end_of_chunk = m_vdev_info.chunk_size;
-                sisl::blob private_blob{r_cast< uint8_t* >(m_init_private_data.get()), sizeof(JournalChunkPrivate)};
-                return private_blob;
-            },
-            m_vdev_info.hs_dev_type, m_vdev_info.vdev_id, m_vdev_info.chunk_size});
+        ChunkPool::Params{HS_DYNAMIC_CONFIG(generic.journal_chunk_pool_capacity),
+                          [this]() {
+                              m_init_private_data->created_at = get_time_since_epoch_ms();
```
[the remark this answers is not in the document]
sanebay: No change here. It was just formatting issue. We use created_at during recovery to find out which head chunk is the latest. We use the latest head chunk.

src/lib/device/journal_vdev.cpp
```diff
@@ -559,54 +574,108 @@ void JournalVirtualDev::Descriptor::truncate(off_t truncate_offset) {
...
+    auto end_new_head_chunk = m_vdev.get_end_of_chunk(new_head_chunk);
+    if (offset_in_chunk >= static_cast< int64_t >(end_new_head_chunk)) {
+        // If the truncation offset is same as end_of_chunk, we dont have any more
+        // data in this chunk to read. Select the next chunk.
+        if (++index <= (m_journal_chunks.size() - 1)) {
+            // Go to the next chunk to make it the new head chunk.
+            new_head_chunk = m_journal_chunks[index];
+            update_truncate_offset = true;
+            size_to_truncate += (new_head_chunk->size() - end_new_head_chunk);
+        }
```
yamingk, line 594: better to put a comment here
// else fall through which means all chunks no chunks left in in this journal and all chunks will be released.

src/lib/device/journal_vdev.cpp
```diff
@@ -559,54 +574,108 @@ void JournalVirtualDev::Descriptor::truncate(off_t truncate_offset) {
...
-    // Align the data_start_offset to the chunk_size as we deleting chunks and
-    // all chunks are same size in a journal vdev.
-    uint32_t start = sisl::round_down(ds_off, m_vdev.info().chunk_size);
+    // All the chunks which are covered inside the truncate offset are released.
+    // cover_offset is a logical offset.
+    index = 0;
+    off_t tail_off =
+        static_cast< off_t >(data_start_offset() + m_write_sz_in_total.load(std::memory_order_relaxed)) + m_reserved_sz;
+    auto chunk_size = m_vdev.info().chunk_size;
+    LOGINFOMOD(journalvdev, "Truncate begin truncate {} desc {}", to_hex(truncate_offset), to_string());
```
yamingk, line 610: I think we need to log it as periodic log and make it info level, in production. We can do every_10th of logging if afraid to much logs.

src/lib/device/journal_vdev.cpp
yamingk: in else branch, we didn't set update_truncate_offset to true, what will be the truncate_offset in else branch? it looks like we rely on `update_truncate_offset` to update data start offset as return value, is it expected the we release all the chunks without updating data start offset?

src/lib/device/journal_vdev.cpp
```diff
@@ -559,54 +574,108 @@ void JournalVirtualDev::Descriptor::truncate(off_t truncate_offset) {
...
+        }
+    }
 
-        // We ideally want to zero out chunks as chunks are reused after free across
-        // logdev's. But zero out chunk is very expensive, We look at crc mismatches
-        // to know the end offset of the log dev during recovery.
-        // Format and add back to pool.
-        m_vdev.m_chunk_pool->enqueue(chunk);
-        LOGINFOMOD(journalvdev, "After truncate released chunk {}", chunk->to_string());
+    if (update_truncate_offset) {
```
yamingk, line 656: is it expected data start offset won't be updated in the case all chunks can be released and `update_truncate_offset` left as false?

src/lib/device/journal_vdev.cpp
```diff
@@ -559,54 +574,108 @@ void JournalVirtualDev::Descriptor::truncate(off_t truncate_offset) {
...
-        // Clear the private chunk data before adding to pool.
-        auto* data = r_cast< JournalChunkPrivate* >(const_cast< uint8_t* >(chunk->user_private()));
-        *data = JournalChunkPrivate{};
-        m_vdev.update_chunk_private(chunk, data);
+            // Release the chunk back to pool.
+            LOGINFOMOD(journalvdev,
+                       "Released chunk_id={} log_dev={} cover={} truncate_offset={} tail={} end_of_chunk={} desc {}",
+                       chunk->chunk_id(), m_logdev_id, to_hex(cover_offset), to_hex(truncate_offset), to_hex(tail_off),
+                       m_vdev.get_end_of_chunk(chunk), to_string());
+            m_vdev.release_chunk_to_pool(chunk);
```
yamingk, line 650: have we thought through if before and after this call (with persistence to disk), everything will be booted up okay?
And we need to have a flip here to simulate this crash and verify boot recovery is fine.

src/lib/device/journal_vdev.cpp
```diff
@@ -559,54 +574,108 @@ void JournalVirtualDev::Descriptor::truncate(off_t truncate_offset) {
...
+            new_head_chunk = m_journal_chunks[index];
+            update_truncate_offset = true;
+            size_to_truncate += (new_head_chunk->size() - end_new_head_chunk);
+        }
+    }
+
     auto* private_data = r_cast< JournalChunkPrivate* >(const_cast< uint8_t* >(new_head_chunk->user_private()));
     private_data->is_head = true;
     private_data->logdev_id = m_logdev_id;
     m_vdev.update_chunk_private(new_head_chunk, private_data);
```
yamingk, line 600: Same comment here:
1. verify the flow if we crash before/after this persistence point, what will be the consequence and does existing recovery flow taken care of this.
2. write a test case to simulate crash point here.

yamingk:
We also need to add flip point when we add new chunks to the journal_vdev, in which we do persistence of the private data and think it through what will happen on reboot and verify with a test case.

yamingk:
Ideally we should have a long running either at log store level or journal vdev level, that we do truncate periodically for a few hours and make sure it is running fine. 
Another test case is run journal or logstore test for 30 mins (with trunation happening more frequently), do a clean shutdown and do another 15 mins run and repeat clean shutdown for multiple times. 

We can review these and see if there is anything comment from others and create issues (not necessaries to be included in this PR).

src/lib/device/journal_vdev.cpp
sanebay: update_data_start_offset is used in this to move to the next chunk size align. We do this when truncate happens to fall in area between end_of_chunk - chunk size or all chunks are truncated. In both cases we roundup to the chunk size. Also update_data_start_offset is in memory, we return the latest data_start_offset back to the caller, which stores in the log dev entry.

src/lib/device/journal_vdev.cpp
sanebay: Ack

src/lib/device/journal_vdev.cpp
sanebay: Yeah periodic log better. But is there a way to do periodic log to log every time when we run UT ? Was hard to debug issues because it wont log every during UT runs.

src/lib/device/journal_vdev.cpp
sanebay: Yeas as discussed above, its in memory. The actual data start offset is returned the caller. Which stores in metablk. LogDev::truncate

src/lib/device/journal_vdev.cpp
sanebay: Added this in test case in the logstore crash recovery doc.

src/lib/device/journal_vdev.cpp
sanebay: Yeah same as above. Added that in doc.

sanebay:
> Ideally we should have a long running either at log store level or journal vdev level, that we do truncate periodically for a few hours and make sure it is running fine. Another test case is run journal or logstore test for 30 mins (with trunation happening more frequently), do a clean shutdown and do another 15 mins run and repeat clean shutdown for multiple times.
> 
> We can review these and see if there is anything comment from others and create issues (not necessaries to be included in this PR).

test_log_store_long_run.cpp is doing that what you mentioned. Will have to run it on 85 namespace.

sanebay:
> We also need to add flip point when we add new chunks to the journal_vdev, in which we do persistence of the private data and think it through what will happen on reboot and verify with a test case.

Added this in the doc.

src/lib/device/journal_vdev.cpp
yamingk: the question here is in else branch, shouldn't we also update the data start offset since all the chunks are truncated? Without setting update_truncate_offset to true, it will just return what ever is there as a staled value in data_start_offset and return to caller? This part I am not able to get it.

src/lib/device/journal_vdev.cpp
yamingk: Ack where? :-)

src/lib/device/journal_vdev.cpp
yamingk: we can just move this log to periodic log and log every time.

src/lib/device/journal_vdev.cpp
yamingk: which doc are you referring to? I think better way is to create an issue, otherwise we will forget which of the tests are there and which are TODOs.

src/lib/device/journal_vdev.cpp
yamingk: see my concerns above. Not clear on the data start offset returned on else branch above.

yamingk:
> > Ideally we should have a long running either at log store level or journal vdev level, that we do truncate periodically for a few hours and make sure it is running fine. Another test case is run journal or logstore test for 30 mins (with trunation happening more frequently), do a clean shutdown and do another 15 mins run and repeat clean shutdown for multiple times.
> > We can review these and see if there is anything comment from others and create issues (not necessaries to be included in this PR).
> 
> test_log_store_long_run.cpp is doing that what you mentioned. Will have to run it on 85 namespace.

Right, is the truncation point ramdomized?

yamingk:
> > We also need to add flip point when we add new chunks to the journal_vdev, in which we do persistence of the private data and think it through what will happen on reboot and verify with a test case.
> 
> Added this in the doc.
Better to create issue otherwise it will lose track which are already there which are todos.

src/lib/device/journal_vdev.cpp
sanebay: We have a issue for this https://github.com/eBay/HomeStore/issues/379

src/lib/device/journal_vdev.cpp
sanebay: I have added that as a comment in line 589

sanebay:
> > > Ideally we should have a long running either at log store level or journal vdev level, that we do truncate periodically for a few hours and make sure it is running fine. Another test case is run journal or logstore test for 30 mins (with trunation happening more frequently), do a clean shutdown and do another 15 mins run and repeat clean shutdown for multiple times.
> > > We can review these and see if there is anything comment from others and create issues (not necessaries to be included in this PR).
> > 
> > 
> > test_log_store_long_run.cpp is doing that what you mentioned. Will have to run it on 85 namespace.
> 
> Right, is the truncation point ramdomized?

Yes  test_log_store_long_run.cpp:459.

src/lib/device/journal_vdev.cpp
sanebay: https://docs.google.com/document/d/1ooA9w_kkg6P9G9l72xDuZpLJ2sRp1mbhCMFwkmRkfRg/edit?pli=1

src/lib/device/journal_vdev.cpp
sanebay: Discussed with Yaming.

src/lib/device/journal_vdev.cpp
sanebay: done.

sanebay:
Created a ticket to use version instead of created time (https://github.com/eBay/HomeStore/issues/441)

yamingk (approved):

yamingk (approved):

PR #426, opened 2024-05-15
add more detail info for metablk debug

JacksonYao287:
the metablk crc mismatch issue is not easy to be reproduced, and one most possible reason is write failure, which will lead to crc mismatch.
as we discussed in homestore meeting,   this PR add more detail debug info which will be helpful when the issue occurs. also in this PR, we will terminate the process with an assert failure so that it will not lead to later CRC mismatch when recovery

the strategy of handling write failures of different service of homestore will be written up to a doc and discussed later on

### DIFF
--- a/src/lib/meta/meta_blk_service.cpp
+++ b/src/lib/meta/meta_blk_service.cpp
@@ -201,7 +201,7 @@ void MetaBlkService::write_ssb() {
         // the offset and buffer length is printed in the error messages of iomgr.
         // buf address here is to show whether the buffer is aligned or not.
         // TODO: hanle this error properly
-        HS_REL_ASSERT(false, "error happens happen during write ssb: {}, buf address: {}", error.what(),
+        HS_REL_ASSERT(false, "error happens happen during write ssb: {}, buf address: {}", error.value(),
                       (const char*)m_ssb);
     }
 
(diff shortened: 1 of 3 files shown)


src/lib/meta/meta_blk_service.cpp
```diff
@@ -196,9 +196,14 @@ void MetaBlkService::format_ssb() {
 // m_meta_lock should be while calling this function;
 void MetaBlkService::write_ssb() {
     // write current ovf blk to disk;
-    try {
-        m_sb_vdev->sync_write((const char*)m_ssb, block_size(), m_ssb->bid);
-    } catch (std::exception& e) { HS_REL_ASSERT(false, "exception happen during write {}", e.what()); }
+    auto error = m_sb_vdev->sync_write((const char*)m_ssb, block_size(), m_ssb->bid);
```
yamingk, line 199: How about create a new function `MetaBlkService::sync_write` and move all error handling to that function, to avoid duplicated error handling?
```
void MetaBlkService::sync_write(const char* data, uint32_t size, BlkId const& bid) { 
        // copy line: 199~206 here 
}
```

JacksonYao287: if we do that , we also have to handle the the error returned by `MetaBlkService::sync_write` whereever it is called. that is almost the same as we use `m_sb_vdev->sync_write` directly

xiaoxichen (approved):
lgtm

src/lib/meta/meta_blk_service.cpp
yamingk: MetaBlkService::sync_write::sync_write will return void and only error handling happens in MetaBlkService::sync_write::sync_write, not in the caller.

JacksonYao287: ok , will do this in another PR

PR #432, opened 2024-05-23
Issue #279 Garbage Collection of Replication request on originator crash/reboot

yamingk:
Why need this change:
=================
When for any reason the replication req's proposer got crashed/rebooted, any requests that rely on this originator to fetch data from or push data from will stay in follower's in-memory cache forever. The new leader once re-elected will resend the requests to the follower with a new replication req(hashed based on new leader). We need a routine to clean up any expired replication requests caused by such error condidtions. 

Specific Changes. 
==============
1. Add Garbage Collection routine for replication requests, it will be triggered in reaper thread on a armed timer. 
2. Add dynamic config to define the timeout for replication config.
3. Add a new test case to simulate originator (leader) reboot that cause garbage requests. 
4. After leader restarts, do more I/Os on the new leader, then tear down the replica group. 

Test:
=====
1. ./bin/test_repl_dev regression passed (all existing test cases passed, except for the known issue that RemoveReplDev can fail as last test case because of no enough space, it happens without this change).
2. New test case passed with logs showing expired reqs got cleaned up.
```
[05/23/24 16:40:55-07:00] [D] [test_raft_repl_dev_replica_1] [83373] [raft_repl_dev.cpp:991] [rd=rdev1] rreq=[dsn=28 term=1 lsn=29 op=HS_DATA_INLINED local_blkid=[] state=[LOG_RECEIVED | LOG_FLUSHED]] is expired, cleaning up
[05/23/24 16:40:55-07:00] [D] [test_raft_repl_dev_replica_1] [83373] [raft_repl_dev.cpp:991] [rd=rdev1] rreq=[dsn=45 term=1 lsn=46 op=HS_DATA_INLINED local_blkid=[] state=[LOG_RECEIVED | LOG_FLUSHED]] is expired, cleaning up

```


src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -198,10 +198,21 @@ raft_buf_ptr_t RaftStateMachine::commit_ext(nuraft::state_machine::ext_op_params
     return m_success_ptr;
 }
 
+void RaftStateMachine::iterate_repl_reqs(std::function< void(int64_t, repl_req_ptr_t rreq) > const& cb) {
+    for (auto const& [key, rreq] : m_lsn_req_map) {
```
hkadayam, line 202: I think its ok for iterator to be accessed by reference, given that m_lsn_req_map is a concurrenthashmap. Have you already checked out this https://github.com/facebook/folly/blob/main/folly/concurrency/ConcurrentHashMap.h#L29 

I just wanted to make sure its safe. If not we can do copy of it because they are cheap to do so.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -977,4 +977,41 @@ void RaftReplDev::cp_flush(CP*) {
 }
 
 void RaftReplDev::cp_cleanup(CP*) {}
+
+void RaftReplDev::gc_repl_reqs() {
+    std::vector< int64_t > expired_keys;
+    m_state_machine->iterate_repl_reqs([this, &expired_keys](auto key, auto rreq) {
```
hkadayam, line 983: Subjective: Given that we are in same class and we don't take any locks, isn't it easier to avoid another method iterate_repl_reqs and simply access the m_lsn_req_map here and iterate?

yamingk: m_lsn_req_map is not same class as RaftReplDev, right? It is under m_state_machine. 
And RaftReplDev shouldn't access it's member's member directly as general principal, otherwise, if someday m_state_machines changes m_lsn_req_map either to a different name or other details, calls doesn't need to be changed.

hkadayam (approved):

hkadayam (approved):

PR #433, opened 2024-05-27
support restart with missing drive

JacksonYao287:
1 support restart with a missing data drive
2 add ut to covery read and free blk residing in a missing drive after restart


xiaoxichen (commented):
Not comment on every line but in general
 1.  better to have `degraded` flag to guard the logic of degrade mode. 
 2. When starting with degrade mode,  we are expecting
    Read. --> should fail on failed drive but success on other drive
    Write ---> should only from replay log?   We need ensuring the log replay can "success".
    Delete(free blk) ---- IMO it should success.

>

xiaoxichen:
also we cannot degrade without meta drive, can we check that as well?

src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -211,13 +217,18 @@ folly::Future< std::error_code > BlkDataService::async_free_blk(MultiBlkId const
...
 
-    m_blk_read_tracker->wait_on(bids, [this, bids, p = std::move(promise)]() mutable {
-        {
-            auto cpg = hs()->cp_mgr().cp_guard();
-            m_vdev->free_blk(bids, s_cast< VDevCPContext* >(cpg.context(cp_consumer_t::BLK_DATA_SVC)));
-        }
-        p.setValue(std::error_code{});
-    });
+    if (!m_vdev->is_blk_exist(bids)) {
+        promise.setValue(std::make_error_code(std::errc::resource_unavailable_try_again));
```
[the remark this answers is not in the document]
JacksonYao287: IMO, dataservice should just return this error to upper layer. and  whether to consider this block as freed, it is up to the upper layer.

src/lib/device/device_manager.cpp
```diff
@@ -144,10 +144,11 @@ void DeviceManager::load_devices() {
     RELEASE_ASSERT_EQ(m_first_blk_hdr.version, first_block_header::CURRENT_SUPERBLOCK_VERSION,
                       "We don't support superblock version upgrade yet");
 
-    RELEASE_ASSERT_EQ(m_first_blk_hdr.num_pdevs, m_dev_infos.size(),
-                      "WARNING: The homestore is formatted with {} devices, but restarted with {} devices. Homestore "
-                      "does not support dynamic addition/removal of devices",
-                      m_first_blk_hdr.num_pdevs, m_dev_infos.size());
+    if (m_first_blk_hdr.num_pdevs != m_dev_infos.size()) {
```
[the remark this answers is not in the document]
JacksonYao287: yea, it also need to handle the case that we start with all the drives living, but mistakenly passing `-degrade-mode`. this can be done in a separate PR

src/lib/device/journal_vdev.cpp
```diff
@@ -81,6 +81,7 @@ void JournalVirtualDev::init() {
 
     // Traverse the chunks and find the heads of the logdev_id's.
     for (auto& chunk : m_all_chunks) {
+        if (!chunk) continue;
```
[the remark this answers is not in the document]
JacksonYao287: as I mentioned above, this PR do not support restarting with any missing fast drive, where meta, log and index services resides.

you are right,  i agree it is better to change this to assert(false) to terminate the process.

src/lib/device/virtual_dev.cpp
```diff
@@ -118,16 +118,18 @@ void VirtualDev::add_chunk(cshared< Chunk >& chunk, bool is_fresh_chunk) {
                                    chunk->physical_dev()->align_size(), chunk->size(), m_auto_recovery,
                                    chunk->chunk_id(), is_fresh_chunk);
     chunk->set_block_allocator(std::move(ba));
-    chunk->set_vdev_ordinal(m_all_chunks.size());
+    // TODO: when vdev_ordinal is  used, revisit here to make sure it is set correctly;
```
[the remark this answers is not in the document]
JacksonYao287: I checked the code , but did not find the place where it is used.
according to the logic of the current code ,  it passes the total chunk num of this vdev to `set_vdev_ordinal` . so I keep this logic here.

src/lib/device/virtual_dev.cpp
```diff
@@ -136,6 +138,7 @@ folly::Future< std::error_code > VirtualDev::async_format() {
     s_futs.clear();
 
     for (auto& chunk : m_all_chunks) {
+        if (!chunk) continue;
```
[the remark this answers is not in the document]
JacksonYao287: it should not happen when first boot. as above , it is better to add an assert here instead of continue as an sanity check.

src/lib/device/virtual_dev.cpp
```diff
@@ -155,6 +158,11 @@ bool VirtualDev::is_blk_alloced(BlkId const& blkid) const {
 
 BlkAllocStatus VirtualDev::commit_blk(BlkId const& blkid) {
     Chunk* chunk = m_dmgr.get_chunk_mutable(blkid.chunk_num());
+    // if we start with missing drive, we will have no chunk for this blkid;
+    if (!chunk) {
```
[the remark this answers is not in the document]
JacksonYao287: commit_blk will happen in 3 cases:
1 repl_dev `on_commit`
2 index commit buf
3 meta service recovery.

so we can not return successfully , for example `meta service recovery`.

src/lib/blkdata_svc/blkdata_service.cpp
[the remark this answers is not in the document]
xiaoxichen: then the return error should be identical enough.

src/lib/device/virtual_dev.cpp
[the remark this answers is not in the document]
xiaoxichen: @hkadayam as the `set_vdev_ordinal` was added in your PR (Device Manager V2),  could you please share some insight? should we drop this or we fix the implementation?

src/tests/test_common/homestore_test_common.hpp
```diff
@@ -221,10 +226,17 @@ class HSTestHelper {
                 }
             }
 
-            for (const auto& name : s_dev_names) {
-                // iomgr::DriveInterface::emulate_drive_type(name, iomgr::drive_type::block_hdd);
-                device_info.emplace_back(name, homestore::HSDevType::Data);
+            if (cust_dev_names.empty()) {
```
yamingk, line 229: I believe this change is not needed.

src/tests/test_common/homestore_test_common.hpp
```diff
@@ -193,8 +193,13 @@ class HSTestHelper {
 
         std::vector< homestore::dev_info > device_info;
         if (!cust_dev_names.empty() || SISL_OPTIONS.count("device_list")) {
-            cust_dev_names.empty() ? s_dev_names = SISL_OPTIONS["device_list"].as< std::vector< std::string > >()
-                                   : s_dev_names = std::move(cust_dev_names);
+            if (cust_dev_names.empty())
```
yamingk, line 196: this is not needed as well.

src/tests/test_common/homestore_test_common.hpp
```diff
@@ -234,7 +246,11 @@ class HSTestHelper {
 
             if (!fake_restart && init_device) { init_files(s_dev_names, dev_size); }
             for (const auto& fname : s_dev_names) {
-                device_info.emplace_back(std::filesystem::canonical(fname).string(), homestore::HSDevType::Data);
+                if (device_info.empty())
```
yamingk, line 249: This shouldn't be needed as well, the data type will be deduced by iomgr and set to FAST automatcially.

src/tests/test_common/hs_repl_test_common.hpp
```diff
@@ -348,7 +349,7 @@ class HSReplTestHelper {
     std::vector< std::string > args_;
     char** argv_;
 
-    std::vector< std::string > dev_list_;
+    std::vector< std::pair< std::string, homestore::HSDevType > > dev_list_;
```
yamingk, line 352: Can you explain why we need to pass in DevType? This is not needed as iomgr detects devType and that is the fundamental assuption between HomeStore and IOMgr.

src/tests/test_common/hs_repl_test_common.hpp
```diff
@@ -169,8 +169,9 @@ class HSReplTestHelper {
                     rdev_list[i].push_back(dev_list_all[i * num_devs_per_replica + j]);
                 }
             }
-
-            dev_list_ = std::move(rdev_list[replica_num_]);
+            for (auto const& dev : rdev_list[replica_num_]) {
```
yamingk, line 172: not needed

src/include/homestore/blkdata_service.hpp
```diff
@@ -142,7 +142,7 @@ class BlkDataService {
      *
      * @param bid The MultiBlkId of the block to commit.
      */
-    void commit_blk(MultiBlkId const& bid);
```
hkadayam, line 145: Isn't it better that the API doesn't care about ignore failure or not and let caller handle that and commit_blk simply returns error or not?

src/lib/device/virtual_dev.cpp
```diff
@@ -118,16 +118,18 @@ void VirtualDev::add_chunk(cshared< Chunk >& chunk, bool is_fresh_chunk) {
...
-    m_all_chunks.push_back(chunk);
+    m_all_chunks[chunk->chunk_id()] = chunk;
     m_chunk_selector->add_chunk(chunk);
 }
 
 void VirtualDev::remove_chunk(cshared< Chunk >& chunk) {
     std::unique_lock lg{m_mgmt_mutex};
     m_all_chunks[chunk->chunk_id()].reset();
     m_all_chunks[chunk->chunk_id()] = nullptr;
+    m_total_chunk_num++;
```
hkadayam, line 132: Why are we incrementing `m_total_chunk_num` on both add and remove chunk?

src/lib/device/virtual_dev.cpp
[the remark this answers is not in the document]
xiaoxichen: I got the intention for `vdev_ordinal` but as far as it is not peristed,  so in restart case it is fully based on the order of chunk load should be same as the time vdev was created.    I am not sure how this get ensured?  also if it is resilient to disk order change (i.e. command line order from Disk A, Disk B, Disk C  to Disk A, Disk C, Disk B)

src/lib/device/virtual_dev.cpp
JacksonYao287: sorry , this is an mistake, thanks

src/include/homestore/blkdata_service.hpp
JacksonYao287: yes, actually ， I wanted to do this also. 
my concern here is that will change the API statement of blkservice , so according to semver ,  we need to bump a MAJOR version of homestore. 
so , i want to do this in a separate PR.

src/tests/test_common/hs_repl_test_common.hpp
JacksonYao287: in test , we always create a file  and consider it as a physical device , right?
when IOManager see this file, it will mark it as a HSDevType:Data.  so I want to explictly mark it as a HSDevType:Fast Drive. so that we can simulate the real-world use case we will face

src/tests/test_common/homestore_test_common.hpp
JacksonYao287: can you point out the code where iomgr will detect and change pdev type?
especially where this changed `std::map< HSDevType, std::vector< PhysicalDev* > > m_pdevs_by_type`

src/tests/test_common/hs_repl_test_common.hpp
yamingk: if you are referring to the file case only (for real drives, iomgr will detect drive type for HomeStore, HomeStore is not supposed to specify drive type as it has no idea (also should not know about) which drive type passed in).

For file case,  there is api called `iomgr::DriveInterface::emulate_drive_type()` provided by iomgr, that is supposed to work with file type and simulate nvme/hdd drive types. It has been used and tested intensively in 1.3 HomeStore/IoMgr.

Bottom line is HomeStore and consumer of homestore, also in real life, shouldn't have knowledge about what drive type is being passed to HomeStore, and we should stick to this rule as it is also what happens in production env.

src/tests/test_common/homestore_test_common.hpp
yamingk:
```
PhysicalDev::read_and_fill_superblock
--> iomgr::DriveInterface::get_attributes
   --> iomgr::DriveInterface::get_drive_type
           // <<<< if emulate_drive_type is called before, get drive type from cache, 
           // <<<< if not called before, call detect_drive_type below
           --> iomgr::DriveInterface::detect_drive_type 
```

JacksonYao287: @yamingk 
1 `PhysicalDev::read_and_fill_superblock` I can not find this in homestore ,can you point out the line and file of this code ? 

2 `iomgr::DriveInterface::get_drive_type` yes， I know we have this, Iomanager does have the capability to detect the device type. however,  according to the current code of master branch , we do not use this capability to determine the device type, right? the device type is always specified by the input parameters?

pls see:
https://github.com/eBay/HomeStore/blob/e9721a0910eacbe36002737dfac015d8875136aa/src/lib/device/device_manager.cpp#L64

and 
https://github.com/eBay/HomeStore/blob/e9721a0910eacbe36002737dfac015d8875136aa/src/lib/device/device_manager.cpp#L129

`dinfo.dev_type` is got from the passed in parameter and never changed by IOmanager, right?

if I miss some code, pls let me know , thanks!

src/tests/test_common/hs_repl_test_common.hpp
JacksonYao287: yea, I absolutely agree with you. we need to face the real-world use case. actually , I also considered about this problem when writing this code. but I did not find the code about the device detection and type change logic. 
pls see my comment here
https://github.com/eBay/HomeStore/pull/433#discussion_r1623146648

 so , I just did all my work based on the current code of master branch and focused on the support of miss drive.
 I also have a plan to add that device detection and type change logic in a seperate PR.

JacksonYao287:
@yamingk let me try to add the device type detection logic in this PR

src/tests/test_common/hs_repl_test_common.hpp
JacksonYao287: @yamingk there is some case that iomanager will mistakenly consider a hdd file as a nvme file. 
for example , in we do some test in  virtual machine, we will create a file in /tmp directory as a physical device. actually it is a hdd file.


however, iomanager will consider it as a fast drive since the mount drive name of /tmp/ is `atomicos-root` and it can not find `atomicos-root` in `/sys/dev/block/`.  `atomicos-root` does not have major and minor dev name. so it will be consider as not rotational, and then it is considered as a file_on_nvme.

src/tests/test_common/homestore_test_common.hpp
yamingk:
```
--> DeviceManager::format_devices
  --> DeviceManager::populate_pdev_info
     --> static bool is_hdd(const std::string& devname)
        --> iomgr::DriveInterface::get_drive_type
```

src/tests/test_common/hs_repl_test_common.hpp
yamingk: As I commented above, the dtype is detected by iomgr from `get_drive_type` (inside `is_hdd(...)`) from HomeStore.

yamingk:
Can you also take care of this comment (PR got merged before we resolve this comment) https://github.com/eBay/HomeStore/pull/426/files#r1621144438, or would you rather make a new PR to do it?

src/tests/test_common/homestore_test_common.hpp
JacksonYao287: I also saw this code, but it just detects the type, but change the type recorded in homestore, right?. 
```
uint32_t DeviceManager::populate_pdev_info(const dev_info& dinfo, const iomgr::drive_attributes& attr,
                                           const uuid_t& uuid, pdev_info_header& pinfo) {
    bool hdd = is_hdd(dinfo.dev_name);

    pinfo.pdev_id = m_cur_pdev_id++;
    pinfo.mirror_super_block = hdd ? 0x01 : 0x00;
    pinfo.max_pdev_chunks = hs_super_blk::max_chunks_in_pdev(dinfo);

    auto sb_size = hs_super_blk::total_size(dinfo);
    pinfo.data_offset = hs_super_blk::first_block_offset() + sb_size;
    pinfo.size = dinfo.dev_size - pinfo.data_offset - (hdd ? sb_size : 0);
    pinfo.dev_attr = attr;
    pinfo.system_uuid = uuid;

    return pinfo.pdev_id;
}
```

but my point is all the type of a pdev is determined by this , right?
```
        auto it = m_pdevs_by_type.find(dinfo.dev_type);
        if (it == m_pdevs_by_type.end()) {
            bool happened;
            std::tie(it, happened) = m_pdevs_by_type.insert(std::pair{dinfo.dev_type, std::vector< PhysicalDev* >{}});
        }
```

if we do not input the pdev type explicitly or modify the pdev type when start , then all the pdev  will be consider as HSDevType::Data.

src/tests/test_data_service.cpp
```diff
@@ -239,6 +243,162 @@ class BlkDataServiceTest : public testing::Test {
...
+        std::vector< std::pair< std::string, homestore::HSDevType > > start_with_devices;
+        auto fast_pdevs = dev_mgr->get_pdevs_by_dev_type(homestore::HSDevType::Fast);
+        for (auto& pdev : fast_pdevs)
+            // can not lose fast drive
+            start_with_devices.emplace_back(pdev->get_devname(), homestore::HSDevType::Fast);
+        // lose one data drive
+        for (auto& pdev : drives) {
+            if (pdev->pdev_id() != missing_pdev->pdev_id())
+                start_with_devices.emplace_back(living_pdev->get_devname(), homestore::HSDevType::Data);
+        }
```
xiaoxichen, line 325: i think you need to move the fast/data check in L253 to ensure the missing_pdev you pick is actually a data drive.

src/tests/test_data_service.cpp
```diff
@@ -239,6 +243,162 @@ class BlkDataServiceTest : public testing::Test {
...
+        });
+
+        wait_for_outstanding_io_done();
+
+        LOGINFO("Step 6: write the blk to living data drive");
+        ++m_outstanding_io_cnt;
+        inst()
+            .async_write(*(sg_write_ptr1.get()), living_drive_blk, false)
+            .thenValue([this, sg_write_ptr1](auto&& err) {
+                RELEASE_ASSERT(!err, "should not be able to write blk on living drive");
```
xiaoxichen, line 364: nit:  should be able

xiaoxichen (approved):
lgtm

PR #434, opened 2024-05-28
Fix raft repl test case for destroying logdev and remove group.

sanebay:
1.Add logstore service metablk for storing last logdev id. logdev id's are always increasing. Because chunks are reused and we get garbage during recovery. We use logdev id to validate if its garbage log entries or not.
2. Test needs to wait till the raft repl is permanently destroyed, as it takes few seconds to do consensus on replica's. Added a new state.

### DIFF
--- a/src/lib/logstore/log_store_service.cpp
+++ b/src/lib/logstore/log_store_service.cpp
@@ -37,8 +37,7 @@ SISL_LOGGING_DECL(logstore)
 LogStoreService& logstore_service() { return hs()->logstore_service(); }
 
 /////////////////////////////////////// LogStoreService Section ///////////////////////////////////////
-LogStoreService::LogStoreService() {
-    m_id_reserver = std::make_unique< sisl::IDReserver >();
+LogStoreService::LogStoreService() : m_sb{"LogStoreServiceSB"} {
     meta_service().register_handler(
         logdev_sb_meta_name,
         [this](meta_blk* mblk, sisl::byte_view buf, size_t size) {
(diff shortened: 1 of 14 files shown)


src/lib/logstore/log_store_service.cpp
```diff
@@ -97,8 +97,6 @@ std::shared_ptr< VirtualDev > LogStoreService::open_vdev(const vdev_info& vinfo,
 void LogStoreService::start(bool format) {
     // hs()->status_mgr()->register_status_cb("LogStore", bind_this(LogStoreService::get_status, 1));
 
-    delete_unopened_logdevs();
```
sanebay, line 100: We let raft repl service to call this after it has loaded all logdev's. Otherwise there is a concurrency issue during restart.

src/tests/test_raft_repl_dev.cpp
```diff
@@ -251,6 +251,13 @@ class RaftReplDevTest : public testing::Test {
                 ASSERT_EQ(err, ReplServiceError::OK) << "Error in destroying the group";
             });
         }
+
+        for (auto const& db : dbs_) {
+            do {
+                std::this_thread::sleep_for(std::chrono::seconds(1));
+            } while (!std::dynamic_pointer_cast< RaftReplDev >(db->repl_dev())->is_destroy_pending());
+            hs()->repl_service().gc_repl_devs();
```
sanebay, line 259: It takes few seconds to actually to get the logdev's permanently deleted.

src/lib/replication/service/raft_repl_service.h
```diff
@@ -64,6 +64,8 @@ class RaftReplService : public GenericReplService,
     void start() override;
     void stop() override;
 
+    void gc_repl_devs() override;
```
hkadayam, line 67: Why can't we make it stay private and test code do
`#define private public`.

This way we protect the code from leaking concerns right?

src/lib/logstore/log_store_service.cpp
```diff
@@ -116,12 +128,17 @@ void LogStoreService::stop() {
         folly::SharedMutexWritePriority::WriteHolder holder(m_logdev_map_mtx);
         m_id_logdev_map.clear();
     }
-    m_id_reserver.reset();
+}
+
+logdev_id_t LogStoreService::get_next_logdev_id() {
+    auto id = ++(m_sb->m_last_logdev_id);
+    m_sb.write();
```
xiaoxichen, line 135: the `write()` is returning void,  curious how we ensure the write will always success?

xiaoxichen: seems like we assert when there is any error

xiaoxichen (approved):
lgtm

src/lib/logstore/log_store_service.cpp
sanebay: Yeah, right alloc_blk is raising exception and if sync_write fails, we assert. I think if any error, then its a big issue. better assert.

Issue #422, opened 2024-05-14
Flush commit_lsn on timer more frequently (outside of CP)


hkadayam:
- Add a timer in repl service to persist durable commit LSN more frequently
- Refer to the method `start_reaper_thread` in `raft_repl_service` to add the timer
- Make sure to persist both checkpoint and durable commit LSN in a checkpoint to maintain the invariant ```Checkpoint LSN <= Durable LSN```
- On restart, reapply the logs between checkpoint commit LSN and durable commit LSN
- Use the method `register_log_found_cb` in log_store to reapply these logs
- Refer to the callback in `solo_repl_dev` for an example

raakella1:
- Add a timer in repl service to persist durable commit LSN more frequently
- Make sure to persist both checkpoint and durable commit LSN in a checkpoint to maintain the invariant Checkpoint LSN <= Durable LSN
- On restart, reapply the logs between checkpoint commit LSN and durable commit LSN
- Use the method `register_log_found_cb` in log_store to reapply these logs

PR #436, opened 2024-05-29
Durable commit idx

raakella1:
More information: https://github.com/eBay/HomeStore/issues/422

### DIFF
--- a/src/lib/common/homestore_config.fbs
+++ b/src/lib/common/homestore_config.fbs
@@ -252,7 +252,7 @@ table Consensus {
     data_receive_timeout_ms: uint64 = 10000;
 
     // Frequency to flush durable commit LSN in millis
-    flush_durable_commit_interval_ms: uint64 = 100;
+    flush_durable_commit_interval_ms: uint64 = 500;
 }
 
 table HomeStoreSettings {
--- a/src/lib/common/homestore_config.fbs
+++ b/src/lib/common/homestore_config.fbs
@@ -250,6 +250,9 @@ table Consensus {
 
     // Timeout for data to be received after raft entry after which raft entry is rejected.
     data_receive_timeout_ms: uint64 = 10000;
+
+    // Frequency to flush durable commit LSN in millis
+    flush_durable_commit_interval_ms: uint64 = 100;
 }
 
 table HomeStoreSettings {
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -38,7 +38,7 @@ RaftReplDev::RaftReplDev(RaftReplService& svc, superblk< raft_repl_dev_superblk 
         m_data_journal =
             std::make_shared< ReplLogStore >(*this, *m_state_machine, m_rd_sb->logdev_id, m_rd_sb->logstore_id);
         m_next_dsn = m_rd_sb->last_applied_dsn + 1;
-        m_commit_upto_lsn = m_rd_sb->commit_lsn;
+        m_commit_upto_lsn = m_rd_sb->durable_commit_lsn;
         m_last_flushed_commit_lsn = m_commit_upto_lsn;
         m_compact_lsn = m_rd_sb->compact_lsn;
 
@@ -470,7 +470,7 @@ void RaftReplDev::check_and_fetch_remote_data(std::vector< repl_req_ptr_t > rreq
 void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
     if (rreqs.size() == 0) { return; }
 
-    std::vector< ::flatbuffers::Offset< RequestEntry > > entries;
+    std::vector<::flatbuffers::Offset< RequestEntry > > entries;
     entries.reserve(rreqs.size());
 
     shared< flatbuffers::FlatBufferBuilder > builder = std::make_shared< flatbuffers::FlatBufferBuilder >();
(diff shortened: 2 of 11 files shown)


src/lib/common/homestore_config.fbs
```diff
@@ -250,6 +250,9 @@ table Consensus {
 
     // Timeout for data to be received after raft entry after which raft entry is rejected.
     data_receive_timeout_ms: uint64 = 10000;
+
+    // Frequency to flush durable commit LSN in millis
+    flush_durable_commit_interval_ms: uint64 = 100;
```
hkadayam, line 255: I think 100 millis is too frequent, would suggest min 500 millis or 1 second.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -37,8 +37,10 @@ RaftReplDev::RaftReplDev(RaftReplService& svc, superblk< raft_repl_dev_superblk
     if (load_existing) {
         m_data_journal =
             std::make_shared< ReplLogStore >(*this, *m_state_machine, m_rd_sb->logdev_id, m_rd_sb->logstore_id);
+        m_data_journal->register_log_found_cb(
+            [this](logstore_seq_num_t lsn, log_buffer buf, void* key) { on_log_found(lsn, buf, key); });
```
hkadayam, line 41: Can we avoid to register and put it in the constructor itself. I wanted to remove register calls going forward anyways.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -470,7 +471,7 @@ void RaftReplDev::check_and_fetch_remote_data(std::vector< repl_req_ptr_t > rreq
 void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
     if (rreqs.size() == 0) { return; }
 
-    std::vector< ::flatbuffers::Offset< RequestEntry > > entries;
+    std::vector<::flatbuffers::Offset< RequestEntry > > entries;
```
hkadayam, line 474: Do you have format_on_save turned on?, since it changed the formatting.

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -343,13 +343,21 @@ void RaftReplService::start_reaper_thread() {
...
+            // Flush durable commit lsns to superblock
+            // FIXUP: what is the best value for flush_durable_commit_interval_ms?
+            m_flush_durable_commit_timer_hdl = iomanager.schedule_thread_timer(
+                HS_DYNAMIC_CONFIG(consensus.flush_durable_commit_interval_ms) * 1000 * 1000, true /* recurring */,
+                nullptr, [this](void*) { flush_durable_commit_lsn(); });
+
             p.setValue();
         } else {
             // Cancel all recurring timers started
+            // FIXUP: When do we cancle the recurring timers?
```
hkadayam, line 355: Aren't we cancelling below? Is the FIXME still relevant?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -968,13 +976,65 @@ void RaftReplDev::cp_flush(CP*) {
...
     m_rd_sb->last_applied_dsn = m_next_dsn.load();
     m_rd_sb.write();
     m_last_flushed_commit_lsn = lsn;
 }
 
 void RaftReplDev::cp_cleanup(CP*) {}
+
+void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx) {
+    // apply the log entry if the lsn is between checkpoint lsn and durable commit lsn
+    if (m_rd_sb->checkpoint_lsn >= lsn || m_rd_sb->durable_commit_lsn <= lsn) { return; }
```
hkadayam, line 993: Nit, interesting that you are checking this way than other way round. This comment is subjective, but to me its easier to read my lsn is out-of-range to whats stored, than other way round :-)

```
if (lsn < m_rd_sb->checkpoint_lsn || lsn < m_rd_sb->durable_commit_lsn) { return; }
```

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -968,13 +976,65 @@ void RaftReplDev::cp_flush(CP*) {
...
+
+    auto const [it, happened] = m_repl_key_req_map.try_emplace(rkey, repl_req_ptr_t(new repl_req_ctx()));
+    RD_DBG_ASSERT((it != m_repl_key_req_map.end()), "Unexpected error in map_repl_key_to_req");
+    auto rreq = it->second;
+    RD_DBG_ASSERT(happened, "rreq already exists for rkey={}", rkey.to_string());
+    MultiBlkId entry_blkid;
+    entry_blkid.deserialize(entry_to_val(jentry), true /* copy */);
+    rreq->set_local_blkid(entry_blkid);
+    rreq->init(rkey, jentry->code, false /* is_proposer */, entry_to_hdr(jentry), entry_to_key(jentry),
+               (entry_blkid.blk_count() * get_blk_size()));
```
hkadayam, line 1029: Nit set_local_blkid could be done after init?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -968,13 +976,65 @@ void RaftReplDev::cp_flush(CP*) {
...
+
+    RD_DBG_ASSERT(rreq, "rreq is nullptr for rkey={}", rkey.to_string());
```
hkadayam, line 1031: Why is this assert needed? Isn't rreq is already used above?

src/lib/common/homestore_config.fbs
raakella1: changed it to 500ms

src/lib/replication/repl_dev/raft_repl_dev.cpp
raakella1: My formatter is removing the spaces, I have to check the clang installation. Changed it for now

src/lib/replication/repl_dev/raft_repl_dev.cpp
raakella1: yeah, makes sense. I was following the block diagram in the doc and the conditional came out to be this :)

src/lib/replication/repl_dev/raft_repl_dev.cpp
raakella1: sounds good

src/lib/replication/repl_dev/raft_repl_dev.cpp
raakella1: It is not needed. Probably carried forward from a copy

hkadayam (approved):

PR #440, opened 2024-06-11
Fix varsize allocator

xiaoxichen:
previously if alloc with is_contiguous = true, we will not try other portion if current portion do not have space. This is wrong as contiguous was ensured by `get_next_contiguous_n_reset_bits`.

This commit fixed the behavior as well as improving logging correctness, before the fix we un-conditionally change a new portion and log a misleading log "alloc direct unable to find in prev portion" even when alloction is success.

MetaBlkMgrTest::recovery_test is a victim of this bug as it tries to use up all the space of meta vdev and will easily hit the case of one portion full.


JacksonYao287:
I am not quite familiar with varsize allocator. i will take some time to understand the implementation, and then leave my comment

JacksonYao287:
@xiaoxichen can you also paste the error stack here?

src/lib/blkalloc/varsize_blk_allocator.cpp
```diff
@@ -582,6 +582,8 @@ blk_count_t VarsizeBlkAllocator::alloc_blks_direct(blk_count_t nblks, blk_alloc_
     if (m_start_portion_num == INVALID_PORTION_NUM) { m_start_portion_num = m_rand_portion_num_generator(re); }
 
     auto portion_num = m_start_portion_num;
+    // save m_start_portion_num to local variable as m_start_portion_num can be changed by other threads.
+    auto start_portion_num = m_start_portion_num;
```
JacksonYao287, line 586: I am not sure why we should have `m_start_portion_num`.
why can not we use a local var `start_portion_num = m_rand_portion_num_generator(re)` for each call of `alloc_blks_direct`. if i understand correctly, here we just only need a random start_portion for loop. coming to what is the start_portion, it does not matter, right?

src/lib/blkalloc/varsize_blk_allocator.cpp
```diff
@@ -614,10 +616,14 @@ blk_count_t VarsizeBlkAllocator::alloc_blks_direct(blk_count_t nblks, blk_alloc_
                 cur_blk_id = b.start_bit + b.nbits;
             }
         }
-        if (++portion_num == get_num_portions()) { portion_num = 0; }
-        BLKALLOC_LOG(TRACE, "alloc direct unable to find in prev portion, searching in portion={}, start_portion={}",
-                     portion_num, m_start_portion_num);
-    } while (nblks_remain && (portion_num != m_start_portion_num) && !hints.is_contiguous && out_blkid.has_room());
```
yamingk, line 620: I think the reason is_contigous is here is we forget to remove it after min_blks is being added that already takes care of is_contigous, e.g. either alloca all of nblks or nothing from current portion.

src/lib/blkalloc/varsize_blk_allocator.cpp
```diff
@@ -614,10 +616,14 @@ blk_count_t VarsizeBlkAllocator::alloc_blks_direct(blk_count_t nblks, blk_alloc_
...
+        if (nblks_remain) {
+            auto curr_portion = portion_num;
+            if (++portion_num == get_num_portions()) { portion_num = 0; }
+            BLKALLOC_LOG(
+                TRACE, "alloc direct unable to find in curr portion {}, will searching in portion={}, start_portion={},continue={}, out_blkid num_pieces={} , max_pieces={}",
+                curr_portion, portion_num, start_portion_num, hints.is_contiguous, out_blkid.num_pieces(), max_pieces);
+        }
+    } while (nblks_remain && (portion_num != start_portion_num) && (out_blkid.num_pieces() < max_pieces));
```
yamingk, line 626: is there a reason why you replace `out_blkid.has_room()` ==> `out_blkid.num_pieces()`? 
My understanding is caller should always access has_room(), and not try to understand pieces inside a blk.

xiaoxichen: it is for  `is_contiguous`, if `is_contiguous` is set we only want one pieces. `has_room()` doesnt serve this purpose 
```
auto const max_pieces = hints.is_contiguous ? 1u : MultiBlkId::max_pieces;
```

src/lib/blkalloc/varsize_blk_allocator.cpp
xiaoxichen: Yes technically it works.  These are two different initiative:

-   In current implementation (with m_start_portion_num shared) it tries to cache the last position we have successful allocation , as a result we tend to use up one portion first then comes to next.
- If we random the start portion on each allocation we tend to have a (statistically)  balance allocation across all regions.  

I dont have strong opinions so I didnt intent to change the behavior.

src/tests/test_meta_blk_mgr.cpp
```diff
@@ -778,7 +778,7 @@ TEST_F(VMetaBlkMgrTest, recovery_test) {
     this->register_client();
 
     // since we are using overflow metablk with 64K metadata, which will cause consume anther 2 metablks
-    auto max_write_times = m_mbm->available_blks() * m_mbm->block_size() / (64 * Ki + 8 * Ki);
+    auto max_write_times = (m_mbm->available_blks() * m_mbm->block_size() - 4 * Mi) / (64 * Ki + 8 * Ki);
```
JacksonYao287, line 781: can you explain why minus 4M?  i am not clear about this

xiaoxichen: there are some boundary issues when we are reaching the size limit,  still WIP to debug, but that is edge case.

It is easy to reproduce by `write_to_full_test`.  Not trying to solve in this pr

JacksonYao287 (approved):
`auto max_write_times = (m_mbm->available_blks() * m_mbm->block_size()) / (64 * Ki + 8 * Ki);`
this is a little confusing. 

i suggest to change it to

`auto max_write_times = m_mbm->available_blks()  / (64 * Ki / m_mbm->block_size()+ 2);`

`64 * Ki / m_mbm->block_size()` means how many blocks is needed to write context_data

`2` means we need two more blk for init_meta_blk and over_flow_meta_blk

src/lib/blkalloc/varsize_blk_allocator.cpp
yamingk: no, num_pieces() means the existing allocated number of pieces on this blk id, technically if there is any blk already in this out_blkid (since it is a MultiBlkId), it will always return false (since max_pieces is 1 when is_contigous is true), right?

If is_contigous is true, we only need 1 pieces, I didn't get why has_room() doesn't serve the purpose, isn't it means out_blkid at least has 1 piece left?

xiaoxichen: in general,  if `is_contigous == true` and we get a partial allocation , where the `num_pieces() == 1` and `has_room == true`,  we should NOT try other portion,  because even we can fulfill in other portion it is still not contiguous.  Through this will not happen as we allocate from bitmap we require all-or-none, but still , `num_pieces() < max_pieces` is more correct compared to `has_room  == true`

in contract, if ``is_contigous == false`, we should try other portion to fulfill the `nblks_remain`.

yamingk: Let's say in replication case, originator wants to send a MultiBlkId to follower, which contains a bunch of BlkIds. 
1st allocation is non-continous and saved to this mbid, and 2nd one is a contigous allocation. 

When originator allocates this alloc down to here, it already contains a few pieces, and this check will be false always (out_blkid.num_pieces() < max_pieces), we will not be able to allocate anything from anywhere, right? 

Have I missed anything?

xiaoxichen: OK if we have a case that a non-empty `out_blkid` passed in,  you are right. Do we do it this way?  I was based on the assumption that `out_blkid` is an out opt.

Now this allocator is mostly on metasvc not in replication flow. But if you are talking nublox 2.0 or something that might be true.

yamingk: yes, I think we should still allow caller to such thing and in nublox 2.x replication, there should be such case.
If we don't allow caller to do such thing (out_blkid has pieces and is_contingous=true), my suggestion is to specifically assert that out_blkid.num_pieces() == 0 if is_contigous = true.

xiaoxichen: I will add an assert and put the discussion context in the code.   So that later in Nublox 2.0 we can easily find out and based on the situation at that time we change the implementation.

PR #445, opened 2024-06-13
register log store callbacks after m_log_store is available

raakella1:
Wait until log store is created to register for the callbacks


sanebay (approved):

xiaoxichen (approved):
yeah this is right!

JacksonYao287 (approved):
LGTM

PR #449, opened 2024-06-19
apply drives num assert only on TestRestartWithMissingDrive

JacksonYao287:
drive num assertion check will only kick in when running BlkDataServiceTest#TestRestartWithMissingDrive

### DIFF
--- a/src/tests/test_common/homestore_test_common.hpp
+++ b/src/tests/test_common/homestore_test_common.hpp
@@ -267,13 +267,6 @@ public:
             }
         }
 
-        if (!fake_restart && token.name_ == "test_data_service")
-            HS_REL_ASSERT_EQ(
-                token.devs_.size() > 2, true,
-                "if not fake restart, we need at least 3 device to run the data_service ut of simulating restart with "
-                "missing drive. current device num is {}",
-                token.devs_.size());
-
         if (is_spdk) {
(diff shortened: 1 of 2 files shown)


src/tests/test_common/homestore_test_common.hpp
```diff
@@ -267,7 +273,7 @@ class HSTestHelper {
             }
         }
 
-        if (!fake_restart && token.name_ == "test_data_service")
+        if (!fake_restart && token.name_ == "test_data_service" && token.ut_name_ == "TestRestartWithMissingDrive")
```
shosseinimotlagh, line 276: The UT has the list of drives. right? you can implement ur algorithm in the UT and not here. Please remove L276-L281 and ut_name__ and corresponding changes.

JacksonYao287: ok , got your idea

JacksonYao287: I already have a assert check for disk amount , so just remove the check here
https://github.com/eBay/HomeStore/blob/4fb24254255c59f5fa61130fe3d049fe0274c075/src/tests/test_data_service.cpp#L252
@shosseinimotlagh

shosseinimotlagh (approved):

PR #458, opened 2024-06-26
Index crash recovery test cases and fixes

hkadayam:
* Fixed the case where if cp is not flushed and homestore crashed, restart would not init certain key parameters. This is fixed by taking cp everytime we do formatting and then we persist this information to homestore first_blk.

* Fixed the crash simulator issue where restart doesn't initialize existing HSTestHelper instances

* Several fixes in IndexTable to make sure we can restart and reapply the logs


src/lib/blkalloc/append_blk_allocator.h
```diff
@@ -109,7 +109,7 @@ class AppendBlkAllocator : public BlkAllocator {
     std::string to_string() const override;
 
     void cp_flush(CP* cp) override;
-
+    void recovery_completed() override {}
```
hkadayam, line 112: Needed to be explicit notification that recovery is completed, so that it can rebuild the caches (for certain blk allocators)

src/lib/blkalloc/fixed_blk_allocator.cpp
```diff
@@ -89,11 +68,29 @@ BlkAllocStatus FixedBlkAllocator::alloc_contiguous(BlkId& out_blkid) { return al
 
 BlkAllocStatus FixedBlkAllocator::reserve_on_cache(BlkId const& b) {
     std::lock_guard lg(m_mark_blk_mtx);
-    HS_DBG_ASSERT(m_state == state_t::RECOVERING, "reserve_on_cache called on non-recovery path");
-    m_marked_blks.insert(b.blk_num());
+    if (m_state == state_t::RECOVERING) { m_marked_blks.insert(b.blk_num()); }
     return BlkAllocStatus::SUCCESS;
 }
 
+void FixedBlkAllocator::recovery_completed() {
```
hkadayam, line 75: Moved the recovery processing from first alloc to the explicit call recovery_completed()

src/lib/common/crash_simulator.hpp
```diff
@@ -9,22 +9,25 @@ namespace homestore {
...
+    CrashSimulator(std::function< void(void) > cb = nullptr) : m_restart_cb{std::move(cb)} {}
     ~CrashSimulator() = default;
 
     void crash() {
         if (m_restart_cb) {
             m_crashed.update([](auto* s) { *s = true; });
 
             // We can restart on a new thread to allow other operations to continue
-            std::thread t([this]() { m_restart_cb(); });
+            std::thread t([cb = std::move(m_restart_cb)]() {
```
hkadayam, line 20: Do not access this pointer (instead move the cb out of this variable)

src/lib/device/device.h
```diff
@@ -149,6 +149,7 @@ class DeviceManager {
 
     bool is_first_time_boot() const { return m_first_time_boot; }
     void format_devices();
+    void commit_formatting();
```
hkadayam, line 152: @yamingk The files device.h, device_manager.cpp, hs_super_blk.h and homestore.cpp are relevant changes regarding taking cp on first_time_boot related changes. Please review these.

src/lib/device/device_manager.cpp
```diff
@@ -176,6 +177,27 @@ void DeviceManager::load_devices() {
...
+        if (!pdev) { continue; }
+
+        auto err = pdev->read_super_block(buf, hs_super_blk::first_block_size(), hs_super_blk::first_block_offset());
+        if (err) {
+            LOGERROR("Failed to read first block from device={}, error={}", pdev->get_devname(), err.message());
+            continue;
+        }
+
+        first_block* fblk = r_cast< first_block* >(buf);
+        fblk->formatting_done = 0x1;
```
yamingk, line 192: In the case of there are three pdevs, and 1 got persisted with formatting_done set to 0x1, before 2 other pdevs persist their superblock, we crash. 

On reboot, what we want is we should return need_format true in homestore::start(). 
However, on reboot in  above case, in `DeviceManager::DeviceManager`, when we detect 1st pdev is_valid() is true (its formatting_done already set to 0x1), we will set `m_first_time_boot` as false, which will return `need_format` as false, which is not expected, right? 

We want to change `DeviceManager::DeviceManager` to only return m_first_time_boot as true when all pdevs is_valid() return true, right?

hkadayam: This needs to be fixed globally and decide strategically, not just for formatting, but in general its a valid device or not. Currently we need superblk from only one device to be valid and if they are, we proceed. We can choose to do it only if all (but we need to worry about new device insertion). I would just use is_valid() and let the DeviceManager::DeviceManager make a decision.

shosseinimotlagh (approved):

src/lib/index/index_cp.cpp
```diff
@@ -265,8 +265,11 @@ void IndexCPContext::process_txn_record(txn_record const* rec, std::map< BlkId,
             if (found) { return buf; }
             real_up_buf->m_down_buffers.emplace_back(buf);
 #endif
-            real_up_buf->m_wait_for_down_buffers.increment(1);
-            buf->m_up_buffer = real_up_buf;
+
+            if (buf->m_up_buffer != real_up_buf) {
+                real_up_buf->m_wait_for_down_buffers.increment(1);
+                buf->m_up_buffer = real_up_buf;
```
shosseinimotlagh, line 271: exactly

src/include/homestore/btree/detail/simple_node.hpp
```diff
@@ -226,66 +232,155 @@ class SimpleNode : public VariantNode< K, V > {
...
-                                (this->is_leaf() ? "LEAF" : "INTERIOR"), snext, this->node_gen());
-            if (!this->is_leaf() && (this->has_valid_edge())) {
-                fmt::format_to(std::back_inserter(str), "edge_id={}.{}", this->edge_info().m_bnodeid,
-                            this->edge_info().m_link_version);
-            }
-            if (this->total_entries() == 0) {
-                fmt::format_to(std::back_inserter(str), " [EMPTY] ");
-                return str;
+        //        return "";
+        // #if 0
```
shosseinimotlagh, line 236: Please return it back` if 0 ` to not break homeobject . I will fix it soon to work for it too, soon

shosseinimotlagh (approved):

PR #459, opened 2024-06-27
disable slab by default in metaservice

JacksonYao287:
this change is made for metaservice. if we have other allocator in the future, we can revisit here

### DIFF
--- a/src/lib/device/virtual_dev.cpp
+++ b/src/lib/device/virtual_dev.cpp
@@ -85,7 +85,7 @@ static std::shared_ptr< BlkAllocator > create_blk_allocator(blk_allocator_type_t
 }
 
 VirtualDev::VirtualDev(DeviceManager& dmgr, vdev_info const& vinfo, vdev_event_cb_t event_cb, bool is_auto_recovery,
-                       shared< ChunkSelector > custom_chunk_selector, bool use_slab_in_blk_allocator) :
+                       shared< ChunkSelector > custom_chunk_selector) :
         m_vdev_info{vinfo},
         m_dmgr{dmgr},
         m_name{vinfo.name},
@@ -94,7 +94,7 @@ VirtualDev::VirtualDev(DeviceManager& dmgr, vdev_info const& vinfo, vdev_event_c
         m_allocator_type{vinfo.alloc_type},
         m_chunk_selector_type{vinfo.chunk_sel_type},
         m_auto_recovery{is_auto_recovery},
-        m_use_slab_in_blk_allocator{use_slab_in_blk_allocator} {
+        m_use_slab_in_blk_allocator{vdev_info.use_slab_in_blk_allocator} {
     switch (m_chunk_selector_type) {
     case chunk_selector_type_t::ROUND_ROBIN: {
         m_chunk_selector = std::make_shared< RoundRobinChunkSelector >(false /* dynamically add chunk */);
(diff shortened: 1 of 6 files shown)


src/lib/device/virtual_dev.cpp
```diff
@@ -68,7 +69,7 @@ static std::shared_ptr< BlkAllocator > create_blk_allocator(blk_allocator_type_t
                                   size,
                                   is_auto_recovery,
                                   std::string("varsize_chunk_") + std::to_string(unique_id),
-                                  is_data_drive_hdd() ? false : true /* use_slabs */};
```
xiaoxichen, line 71: why this has anything to do with `is_data_drive_hdd`?

yamingk: for non-metablk use case, which is the user data blk allocation, when it is HDD drives (for nublox use case), we adapt to HDD streams for best performance, which is each *user data* blk allocation inside a volume is directed to a stream (unquely owned by this volume), so that random write on this volume becomes sequencial write on this stream (internally mapped to a set of physical chunks). If use_slabs is true, stream concept will be overriden and it will choose blks from slabs which is not preferred in HDD use case.

src/lib/device/virtual_dev.hpp
```diff
@@ -99,10 +99,11 @@ class VirtualDev {
     blk_allocator_type_t m_allocator_type;
     chunk_selector_type_t m_chunk_selector_type;
     bool m_auto_recovery;
+    bool m_use_slab_in_blk_allocator;
```
yamingk, line 102: I think we should add a parameter `use_slab_allocator` inside `struct vdev_parameters` and let each service to set this field while calling `DeviceManager::create_vdev`, and in `populate_vdev_info` it needs to be persistent in vdev_info as well to survive recovery restore of the meta service vdev.

src/lib/device/virtual_dev.cpp
```diff
@@ -68,7 +69,7 @@ static std::shared_ptr< BlkAllocator > create_blk_allocator(blk_allocator_type_t
...
+                                  !is_data_drive_hdd() && use_slab_in_blk_allocator /* use_slabs */};
```
yamingk, line 72: This change is fine. 

Now that we have input parameter to set use_slab, ideally, we should move this check `is_data_drive_hdd()` to upper layer (service) to only set use_slab in their parameter? E.g. for meta service, it doesn't care whether there is hdd or not, just set use_slab as false, for other layer like data service, for nublox 2.0 case, data service layer will need to check whether hdd is there or not to set use_slab true or false. 

You can make the decision. If we want to keep this as it is, I think we can have an assert here saying if use_slab is set to true by caller, we should assert is_data_drive_hdd() is false, otherwise, caller would be surprised why I set use_slab to true, but my blk allocator silently set it to false. Make sense?

src/lib/meta/meta_blk_service.cpp
```diff
@@ -465,7 +465,9 @@ void MetaBlkService::write_ovf_blk_to_disk(meta_blk_ovf_hdr* ovf_hdr, const uint
     size_t write_size = ovf_hdr->h.context_sz;
     uint8_t* context_data_aligned{nullptr};
     if (!hs_utils::is_ptr_aligned(write_context_data, align_sz) || !hs_utils::mod_aligned_sz(write_size, align_sz)) {
-        HS_LOG_EVERY_N(WARN, metablk, 50, "[type={}] Unaligned address found for input context_data, ptr {}, size {}, align {} ", type, (void *)write_context_data, write_size, align_sz);
+        HS_LOG_EVERY_N(WARN, metablk, 50,
```
yamingk, line 468: nit: looks like your clang-format line limit is not same as the one who made the change previously.

yamingk (approved):

src/lib/device/virtual_dev.cpp
JacksonYao287: yea, agree, I can add the alert in later pr

PR #460, opened 2024-06-30
fix log flush bug

JacksonYao287:


src/lib/logstore/log_store.cpp
```diff
@@ -422,7 +420,7 @@ void HomeLogStore::flush_sync(logstore_seq_num_t upto_seq_num) {
         m_logdev->flush_if_needed(1);
 
         // Step 4: Wait for completion
-        m_sync_flush_cv.wait(lk, [this, upto_seq_num] { return !m_records.status(upto_seq_num).is_active; });
```
xiaoxichen, line 425: I am not sure if we will have multiple concurrent `flush_sync` , if we have, this implementation is not right. Although the previous one is also not right:)

xiaoxichen: Not necessary concurrent...

First  call with `flush_sync (100)`,  it execute to 416.  At the moment,  the `on_write_completion` run to finish and set the `m_sync_flush_promise`.  The `flush_sync (100)` fulfilled in L417 and not consume the future.
Next time , a `flush_sync (1078576)` will return immediately by consuming the previous fulfilled future.

JacksonYao287: do you mean there is a potential bug？ this part is protected by a lock, and everytime we enter this block, m_sync_flush_promise will be updated to a new one.

JacksonYao287: you can try it locally in your env. it runs well for hours in mine

JacksonYao287:
will use sync flush , so this pr is unnecessary again

PR #461, opened 2024-07-10
sync_flush should wait is_completed not is_active.

xiaoxichen:
Log entry added into StreamTracker in write_async , through m_records.create().  Once create the status is active.

Later on in HomeLogStore::on_write_completion it update the tracker and in the `m_records.update`, is_completed set to true.

We should wait for is_completed == true to indicate a log entry has been persisted.

### DIFF
--- a/src/lib/logstore/log_store.cpp
+++ b/src/lib/logstore/log_store.cpp
@@ -176,10 +176,11 @@ void HomeLogStore::read_async(logstore_seq_num_t seq_num, void* cookie, const lo
 #endif
 
 void HomeLogStore::on_write_completion(logstore_req* req, const logdev_key& ld_key) {
+    std::unique_lock lk(m_sync_flush_mtx);
     // Upon completion, create the mapping between seq_num and log dev key
     m_records.update(req->seq_num, [&](logstore_record& rec) -> bool {
         rec.m_dev_key = ld_key;
-        // THIS_LOGSTORE_LOG(DEBUG, "Completed write of lsn {} logdev_key={}", req->seq_num, ld_key);
+        THIS_LOGSTORE_LOG(DEBUG, "Completed write of store lsn {} logdev_key={}", req->seq_num, ld_key);
         return true;
     });
     // assert(flush_ld_key.idx >= m_last_flush_ldkey.idx);
@@ -405,6 +406,7 @@ void HomeLogStore::flush_sync(logstore_seq_num_t upto_seq_num) {
     HS_DBG_ASSERT_EQ(LogDev::can_flush_in_this_thread(), false,
                      "Logstore flush sync cannot be called on same thread which could do logdev flush");
 
+    std::unique_lock lk(m_single_sync_flush_mtx);
     if (upto_seq_num == invalid_lsn()) { upto_seq_num = m_records.active_upto(); }
 
     // if we have flushed already, we are done
--- a/src/lib/logstore/log_store.cpp
+++ b/src/lib/logstore/log_store.cpp
@@ -125,14 +125,17 @@ logstore_seq_num_t HomeLogStore::append_async(const sisl::io_blob& b, void* cook
 
 log_buffer HomeLogStore::read_sync(logstore_seq_num_t seq_num) {
     // If seq_num has not been flushed yet, but issued, then we flush them before reading
-    auto const s = m_records.status(seq_num);
+    auto s = m_records.status(seq_num);
+    THIS_LOGSTORE_LOG(DEBUG, "read_sync seq_num {}, is_active {}, is_complete {}, is_hole {}, is_out_of_range {}", seq_num, s.is_active, s.is_completed, s.is_hole, s.is_out_of_range);
     if (s.is_out_of_range || s.is_hole) {
         // THIS_LOGSTORE_LOG(ERROR, "ld_key not valid {}", seq_num);
         throw std::out_of_range("key not valid");
     } else if (!s.is_completed) {
-        THIS_LOGSTORE_LOG(TRACE, "Reading lsn={}:{} before flushed, doing flush first", m_store_id, seq_num);
+        THIS_LOGSTORE_LOG(DEBUG, "Reading lsn={}:{} before flushed, doing flush first", m_store_id, seq_num);
         flush_sync(seq_num);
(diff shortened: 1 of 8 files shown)


xiaoxichen:
I dont know how much removing async will remove this part of code  @JacksonYao287

src/lib/logstore/log_store.cpp
```diff
@@ -405,7 +408,7 @@ void HomeLogStore::flush_sync(logstore_seq_num_t upto_seq_num) {
     if (upto_seq_num == invalid_lsn()) { upto_seq_num = m_records.active_upto(); }
 
     // if we have flushed already, we are done
-    if (!m_records.status(upto_seq_num).is_active) { return; }
+    if (m_records.status(upto_seq_num).is_completed) { return; }
```
hkadayam, line 411: This status is a tri-state check.
1) Inactive: We have not received this lsn at all
2) Active: Received LSN, yet to flush
3) Completed: Completed flush.

So we only need to flush if it is not active. This change will flush when it is inactive as well. So I don't think we should make this change.

src/lib/logstore/log_store.cpp
```diff
@@ -416,13 +419,13 @@ void HomeLogStore::flush_sync(logstore_seq_num_t upto_seq_num) {
 
         // Step 2: After marking this lsn, we again do a check, to avoid a race where completion checked for no lsn
         // and the lsn is stored in step 1 above.
-        if (!m_records.status(upto_seq_num).is_active) { return; }
+        if (m_records.status(upto_seq_num).is_completed) { return; }
```
hkadayam, line 422: Same as above.

src/lib/logstore/log_store.cpp
```diff
@@ -416,13 +419,13 @@ void HomeLogStore::flush_sync(logstore_seq_num_t upto_seq_num) {
...
 
         // Step 3: Force a flush (with least threshold)
         m_logdev->flush_if_needed(1);
 
         // Step 4: Wait for completion
-        m_sync_flush_cv.wait(lk, [this, upto_seq_num] { return !m_records.status(upto_seq_num).is_active; });
+        m_sync_flush_cv.wait_for(lk, std::chrono::milliseconds(10), [this, upto_seq_num] { return m_records.status(upto_seq_num).is_completed; });
```
hkadayam, line 428: I am assuming it is temporary, because caller expect data at the end of it. Either we return an error or we assert? In any case, after new flush changes, we probably don't need this cv so at that point it is moot.

src/lib/replication/log_store/repl_log_store.cpp
```diff
@@ -73,6 +81,17 @@ void ReplLogStore::end_of_append_batch(ulong start_lsn, ulong count) {
...
+            if ((rreq == nullptr) || (!rreq->has_linked_data())) { continue; }
+            LOGINFO("Raft Channel: Data after future wait: rreq=[{}]", rreq->to_compact_string());
+        }
+    } else if (!proposer_reqs->empty()) {
+        LOGINFO("Raft Channel: end_of_append_batch, I am proposer, only flush log s from {} , count {}", start_lsn, count);
+        // Mark all the reqs also completely written
+        for (auto const& rreq : *reqs) {
+            if (rreq) { rreq->add_state(repl_req_state_t::LOG_FLUSHED); }
+        }
     }
```
hkadayam, line 95: Need to call ```sisl::VectorPool< repl_req_ptr_t >::free(proposer_reqs);```

src/lib/replication/log_store/repl_log_store.cpp
```diff
@@ -73,6 +81,17 @@ void ReplLogStore::end_of_append_batch(ulong start_lsn, ulong count) {
...
         }
+
+        for (auto const& rreq : *reqs) {
+            if ((rreq == nullptr) || (!rreq->has_linked_data())) { continue; }
+            LOGINFO("Raft Channel: Data after future wait: rreq=[{}]", rreq->to_compact_string());
+        }
+    } else if (!proposer_reqs->empty()) {
+        LOGINFO("Raft Channel: end_of_append_batch, I am proposer, only flush log s from {} , count {}", start_lsn, count);
+        // Mark all the reqs also completely written
+        for (auto const& rreq : *reqs) {
```
hkadayam, line 92: Don't this need to be `*proposer_reqs`?

src/lib/replication/log_store/repl_log_store.cpp
```diff
@@ -41,19 +41,27 @@ void ReplLogStore::end_of_append_batch(ulong start_lsn, ulong count) {
...
+	else {reqs->emplace_back(std::move(rreq));}
     }
 
     RD_LOGT("Raft Channel: end_of_append_batch start_lsn={} count={} num_data_to_be_written={}", start_lsn, count,
             reqs->size());
 
+    for (auto const& rreq : *reqs) {
+        if ((rreq == nullptr) || (!rreq->has_linked_data())) { continue; }
+        LOGINFO("Raft Channel: Data before future wait: rreq=[{}]", rreq->to_compact_string());
+    }
```
hkadayam, line 63: I am assuming this will be removed after we identify the root cause, right?

src/lib/replication/log_store/repl_log_store.cpp
hkadayam: Also we need to call `HomeRaftLogStore::end_of_append_batch(start_lsn, count);` right?

src/lib/logstore/log_store.cpp
xiaoxichen: yes this is temporary for solving the racing we had seen. It is not right , but generally ensure we dont stuck on this CV infinitely for concurrent issue.

src/lib/logstore/log_store.cpp
xiaoxichen: I get the point, we need to flush only active, it make sense.
However  I am not getting how `is_active == false` implies  log getting flushed?  are we reset bit in `m_active_slot_bits` when doing update?  I didnt find this part of code in SISL.

src/lib/replication/log_store/repl_log_store.cpp
xiaoxichen: yes it is temporary

PR #464, opened 2024-07-16
Improve test_raft_repl_dev UT

xiaoxichen:
majar changes:

1. fix device list
2. when leader switches , new leader will continue to write in `write_on_leader`.
3. logging improvement, print out ms/us for timestamp.

Other than UT

implement rollback for statemachine.

### DIFF
--- a/src/tests/test_common/homestore_test_common.hpp
+++ b/src/tests/test_common/homestore_test_common.hpp
@@ -359,6 +359,14 @@ private:
             // Fake restart, device list is unchanged.
             shutdown_homestore(false);
             std::this_thread::sleep_for(std::chrono::seconds{shutdown_delay_sec});
+        } else if (!m_token.devs_.empty()) {
+            // For those UTs already process device list, like test_raft_repl_dev.
+            LOGINFO("Using passed in dev_list: {}",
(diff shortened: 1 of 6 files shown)


src/tests/test_common/homestore_test_common.hpp
```diff
@@ -359,6 +359,14 @@ class HSTestHelper {
             // Fake restart, device list is unchanged.
             shutdown_homestore(false);
             std::this_thread::sleep_for(std::chrono::seconds{shutdown_delay_sec});
+        } else if (!m_token.devs_.empty()) {
```
yamingk, line 362: now we have three use cases, priority-wise:
1. UT specificly passed dev_list, if not there, 
2. --device_list from input, if not there, 
3. m_token.name_ which equeals to test_binary, name. 

Can we put it as in comment at line: 178, so that it can be friendly to us after a few month when we look back and don't need to scratch our head wondering why?

zhiteng:
This PR has been idle for one month.  Do we still want to merge this?

JacksonYao287:
>This PR has been idle for one month. Do we still want to merge this?

this PR is very helpful to make repl_dev unit test stable in github CI. after this , we can  enable `Leader_Restart` which stucks intermittently.

@xiaoxichen pls go ahead and complete this PR when you get cycle.

PR #469, opened 2024-07-20
log store sync flush

JacksonYao287:
this PR implement logstore sync flush based on Hari`s code piece
https://github.com/hkadayam/HomeStore/commit/87d50755195e4b661832b59a2c4a604785600655

this disabled `LogStoreTest.BurstSeqInsertAndTruncateInParallel`, which will lead to stuck previously , is enabled now


src/lib/logstore/log_store.cpp
```diff
@@ -397,96 +260,67 @@ logstore_seq_num_t HomeLogStore::get_contiguous_completed_seq_num(logstore_seq_n
...
-        // Step 4: Wait for completion
-        m_sync_flush_cv.wait(lk, [this, upto_seq_num] { return !m_records.status(upto_seq_num).is_active; });
+    if (upto_lsn == invalid_lsn()) { upto_lsn = m_records.active_upto(); }
 
-        // NOTE: We are not resetting the lsn because same seq number should never have 2 completions and thus not
-        // doing it saves an atomic instruction
-        THIS_LOGSTORE_LOG(TRACE, "flush_sync over upto_seq_num {}", upto_seq_num);
-    }
+    // if we have flushed already, we are done, else issue a flush
+    if (!m_records.status(upto_lsn).is_completed) m_logdev->flush_under_guard();
```
xiaoxichen, line 273: I had been corrected by @hkadayam once :).  this should be `m_records.status(upto_lsn).is_active`,  flushing an inactive LSN (usually indicating the LSN is not yet received)  is a no-op instead of blocking --- This behavior seems a bit strange to me as well though.

src/lib/logstore/log_dev.hpp
```diff
@@ -838,73 +758,84 @@ class LogDev : public std::enable_shared_from_this< LogDev > {
...
     Clock::time_point m_last_flush_time;
 
     logid_t m_last_flush_idx{-1}; // Track last flushed, last device offset and truncated log idx
-    off_t m_last_flush_dev_offset{0};
-    logid_t m_last_truncate_idx{-1};
+    logid_t m_last_truncate_idx{std::numeric_limits< logid_t >::min()}; // logdev truncate up to this idx
+
+    // after completing flush, we can not truncate the log records directly. since we might shedule new flush/append in
+    // the callback of the append_async, which can be found in the logstore test case.
+    // this means we might schedule call on_flush_completion in another on_flush_completion, which is a recursive call.
```
xiaoxichen, line 793: now we still under the `flush guard`  in `on_flush_completion`,  how can  we have recursive call to `on_flush_completion`?

JacksonYao287: logstore supports inline flush mode.  in this flush mode , when we use append_async and in the callback we schedule another flush or append_async , we need to do flush before the previous flush is completed, since the callback will be called in `on_write_completion` , which is also under the protection of flush guard.
pls refer to https://github.com/eBay/HomeStore/blob/39781c3fb7e7b4da3f77f0a9ebb3e0ad73f2ddfc/src/tests/test_log_store.cpp#L153

src/lib/logstore/log_store.cpp
JacksonYao287: 1 when creating a item in `m_records.create`, it will be marked as `active` 
2 when coming to `on_write_completion` which means it has been flushed, we will call `m_records.update`, which will mark this item as `completed`

so , no matter the log entry is flushed , as long as it is created and not truncated, it is active.
only after the entry is flushed, it is completed. 

here, we want to flush those entries which has not been flushed, so `is_completed` is an indicator to indicate whether the log has been flushed. right?

xiaoxichen: c.f https://github.com/eBay/HomeStore/pull/461#discussion_r1674383132

src/lib/logstore/log_dev.hpp
xiaoxichen: sorry if this is a stupid question.
Dont we get deadlock if we are under `flush guard` and calling ` append_async` which is trying to obtain the `flush guard` again?

JacksonYao287: 1 if this two happens in different thread, `append_async` will try to acquire the flush lock, , it will not block if it can not acquire the lock.

2 if this two happens in the same thread, since we use `std:recursive_mutex` now, which can be acquired for multiple times , it will not block and not dead lock happen

src/lib/logstore/log_store.cpp
JacksonYao287: i do not fully get Hari`s point in this comment.

 > Active: Received LSN, yet to flush.
 
 I do not think this is true. when we complete flush, we only mark the log entry as completed,   we do not make any change to the status of active. `completed` and `active` are two different status, which are maintained in two different bitmap. That means , if a log record is active, it can be completed or not.
 
 `So we only need to flush if it is not active`
I do not under this sentence. if it is not active , it mean the log entry has not been received or has been rollbacked, so how to flush those log that does not exist.

@hkadayam PTAL

JacksonYao287: yes, that is right, thank you

src/lib/logstore/log_dev.cpp
```diff
@@ -399,233 +399,152 @@ bool LogDev::flush_if_needed(int64_t threshold_size) {
...
-        if (!m_is_flushing.compare_exchange_strong(expected_flushing, true, std::memory_order_acq_rel)) {
-            // Flushing is blocked already, add it to the callback q
-            if (m_block_flush_q == nullptr) { m_block_flush_q = sisl::VectorPool< flush_blocked_callback >::alloc(); }
-            m_block_flush_q->emplace_back(cb);
-            return false;
-        }
+    // Truncate the log records, as it is no longer needed to be kept in memory. Everytime there is a read, we actually
+    // read from the drives.
+    // TODO:: we can keep this in memory to accelerate read until it is truncated
+    if (from_indx == m_log_records_last_truncate_to_idx + 1) {
```
xiaoxichen, line 480: can we have a policy to keep last N records in memory?

xiaoxichen (commented):
not yet finished all review，put partial

src/lib/logstore/log_dev.cpp
```diff
@@ -399,233 +399,152 @@ bool LogDev::flush_if_needed(int64_t threshold_size) {
...
-        }
-
-        bool expected_flushing{false};
-        if (!m_is_flushing.compare_exchange_strong(expected_flushing, true, std::memory_order_acq_rel)) {
-            // Flushing is blocked already, add it to the callback q
-            if (m_block_flush_q == nullptr) { m_block_flush_q = sisl::VectorPool< flush_blocked_callback >::alloc(); }
-            m_block_flush_q->emplace_back(cb);
-            return false;
-        }
+    // Truncate the log records, as it is no longer needed to be kept in memory. Everytime there is a read, we actually
```
xiaoxichen, line 477:
```
// Truncate the in-memory log records. Further read will be read from drives.
```

src/lib/logstore/log_dev.cpp
```diff
@@ -399,233 +399,152 @@ bool LogDev::flush_if_needed(int64_t threshold_size) {
...
     auto dev_offset = lg->m_log_dev_offset;
     for (auto idx = from_indx; idx <= upto_indx; ++idx) {
         auto& record = m_log_records->at(idx);
-        on_io_completion(record.store_id, logdev_key{idx, dev_offset}, flush_ld_key, upto_indx - idx, record.context);
+        logstore_req* req = s_cast< logstore_req* >(record.context);
+        HomeLogStore* log_store = req->log_store;
+        HS_LOG_ASSERT_EQ(log_store->get_store_id(), record.store_id,
+                         "Expecting store id in log store and flush completion to match");
+
+        log_store->on_write_completion(req, logdev_key{idx, dev_offset});
```
xiaoxichen, line 467: can we do the callback w/o holding flush_guard?

src/lib/logstore/log_dev.cpp
```diff
@@ -259,6 +259,18 @@ int64_t LogDev::append_async(logstore_id_t store_id, logstore_seq_num_t seq_num,
     return idx;
 }
 
+log_buffer LogDev::read_from_cache(const logid_t& idx) {
```
xiaoxichen, line 262: io_blob is NOT a safe buffer.

This implementation doesnt seems right,  the client of LogStore can probably release the byte buffer provided, before truncation.  unless we change from top-down to make it `io_blob_safe`.  

But in general I dont think this is right approach as it will consume a huge amount of memory, if client is lazily truncate.

src/lib/logstore/log_dev.cpp
```diff
@@ -255,31 +250,33 @@ void LogDev::assert_next_pages(log_stream_reader& lstream) {
...
+}
 
-    if (HS_DYNAMIC_CONFIG(logstore.flush_threshold_size) == 0) {
-        // This is set in tests to disable implicit flush. This will be removed in future.
-        return idx;
+log_buffer LogDev::read_from_cache(const logid_t& idx) {
+    sisl::io_blob buf;
+    try {
+        buf = m_log_records->at(idx).data;
+    } catch (const std::out_of_range& e) {
```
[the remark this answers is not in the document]
JacksonYao287: ok, let`s do it follow our decision, so that we do not make any optimization about home_log_store, it will work like before.

we only make the optimization at home_raft_log_store level. 

@xiaoxichen 
for home_log_store, it will always flush and read from vdev like before, so there will not be any potential issue about the safe buffer

src/include/homestore/logstore/log_store_internal.hpp
```diff
@@ -109,9 +106,12 @@ struct log_dump_req {
 
 struct logstore_record {
     logdev_key m_dev_key;
+    logdev_key m_trunc_key;
+    // only when out-of-order write happens and the higher lsn is flushed in a lower LogGroup while the lower lsn is
```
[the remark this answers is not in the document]
JacksonYao287: pls refer to the comments in homeLogStore#on_write_completion. I will make the comments as simple as possible

src/lib/logstore/log_dev.cpp
```diff
@@ -105,26 +108,17 @@ void LogDev::start(bool format) {
...
+                  "LogDev stop attempted while writes to logdev are pending completion");
     {
-        // Wait for the stopped to be completed
-        std::unique_lock< std::mutex > lk{m_block_flush_q_mutex};
-        m_block_flush_q_cv.wait(lk, [&] { return m_stopped; });
+        std::unique_lock lg = flush_guard();
+        m_stopped = true;
     }
+    // after we call stop, we need to do any pending device truncations
+    truncate();
```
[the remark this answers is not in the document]
JacksonYao287: sometimes , log store truncation will only do truncation in memory, they will not issue log dev truncation.
the idea here is  when we stopping log dev, doing a truncation here can remove those stale log as much as possible, which is only be truncated in memory. generally , on log dev can sustain multiple log store, this will make the data which needs to be loaded as less as possible.

logdev truncation is a  best-effort  truncation. that means if raft and  resource manager does  not issue any truncation(log store truncation), log dev truncation will do nothing. log dev truncation  will only truncate those which has already been truncated by log store.

src/lib/logstore/log_dev.cpp
JacksonYao287: ignore this, we will only buffer log entry for home raft log store. so no changes will not be made.

src/lib/logstore/log_dev.cpp
```diff
@@ -399,233 +399,152 @@ bool LogDev::flush_if_needed(int64_t threshold_size) {
...
     bool const flush_by_time =
         !flush_by_size && pending_sz && (elapsed_time > HS_DYNAMIC_CONFIG(logstore.max_time_between_flush_us));
-
     if (flush_by_size || flush_by_time) {
-        // First off, check if we can flush in this thread itself, if not, schedule it into different thread
-        if (!can_flush_in_this_thread()) {
-            iomanager.run_on_forget(logstore_service().flush_thread(),
-                                    [this, threshold_size]() { flush_if_needed(threshold_size); });
-            return false;
+        std::unique_lock lck(m_flush_mtx, std::try_to_lock);
```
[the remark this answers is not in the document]
JacksonYao287: this is in flush_if_necessary ,  it is not a force-to flush , so no need to retry. if it can not be flushed this time, it can be flushed next time.

src/lib/logstore/log_dev.cpp
```diff
@@ -399,233 +399,152 @@ bool LogDev::flush_if_needed(int64_t threshold_size) {
...
-    do_flush_write(lg);
-}
+    auto sz = m_pending_flush_size.fetch_sub(lg->actual_data_size(), std::memory_order_relaxed);
+    HS_REL_ASSERT_GE((sz - lg->actual_data_size()), 0, "size {} lg size{}", sz, lg->actual_data_size());
+    off_t offset = m_vdev_jd->alloc_next_append_blk(lg->header()->total_size());
+    lg->m_log_dev_offset = offset;
+
+    HS_REL_ASSERT_NE(lg->m_log_dev_offset, INVALID_OFFSET, "log dev is full");
+    THIS_LOGDEV_LOG(TRACE, "Flushing log group data size={} at offset=0x{} log_group={}", lg->actual_data_size(),
+                    to_hex(offset), *lg);
```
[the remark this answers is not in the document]
JacksonYao287: ack

src/lib/logstore/log_dev.cpp
```diff
@@ -399,233 +399,152 @@ bool LogDev::flush_if_needed(int64_t threshold_size) {
...
-
-    // if (has_data_service() && data_service().is_fsync_needed()) {
-    //     data_service().fsync([this, lg]() { do_flush_write(lg); })
-    // } else {
-    //     do_flush_write(lg);
-    // }
-    do_flush_write(lg);
-}
+    auto sz = m_pending_flush_size.fetch_sub(lg->actual_data_size(), std::memory_order_relaxed);
+    HS_REL_ASSERT_GE((sz - lg->actual_data_size()), 0, "size {} lg size{}", sz, lg->actual_data_size());
```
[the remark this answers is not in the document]
JacksonYao287: >But the fetch_add is not yet executed.

sorry, can you explain in more detail? 
I think fetch_add will be executed before flush is issued.

src/lib/logstore/log_dev.cpp
```diff
@@ -399,233 +399,152 @@ bool LogDev::flush_if_needed(int64_t threshold_size) {
...
 
-void LogDev::do_flush_write(LogGroup* lg) {
     HISTOGRAM_OBSERVE(logstore_service().m_metrics, logdev_flush_records_distribution, lg->nrecords());
     HISTOGRAM_OBSERVE(logstore_service().m_metrics, logdev_flush_size_distribution, lg->actual_data_size());
-    THIS_LOGDEV_LOG(TRACE, "vdev offset={} log group total size={}", lg->m_log_dev_offset, lg->header()->total_size());
 
     // write log
-    m_vdev_jd->async_pwritev(lg->iovecs().data(), int_cast(lg->iovecs().size()), lg->m_log_dev_offset)
-        .thenValue([this, lg](auto) { on_flush_completion(lg); });
+    m_vdev_jd->sync_pwritev(lg->iovecs().data(), int_cast(lg->iovecs().size()), lg->m_log_dev_offset);
```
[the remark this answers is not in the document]
JacksonYao287: ack

src/lib/logstore/log_dev.cpp
JacksonYao287: done

src/lib/logstore/log_dev.cpp
JacksonYao287: removed

src/lib/logstore/log_dev.cpp
JacksonYao287: removed

src/lib/logstore/log_dev.cpp
```diff
@@ -399,233 +399,152 @@ bool LogDev::flush_if_needed(int64_t threshold_size) {
...
-        bool expected_flushing{false};
-        if (!m_is_flushing.compare_exchange_strong(expected_flushing, true, std::memory_order_acq_rel)) {
-            // Flushing is blocked already, add it to the callback q
-            if (m_block_flush_q == nullptr) { m_block_flush_q = sisl::VectorPool< flush_blocked_callback >::alloc(); }
-            m_block_flush_q->emplace_back(cb);
-            return false;
-        }
+    // Truncate the log records, as it is no longer needed to be kept in memory. Everytime there is a read, we actually
+    // read from the drives.
+    // TODO:: we can keep this in memory to accelerate read until it is truncated
```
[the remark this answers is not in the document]
JacksonYao287: removed

src/lib/logstore/log_dev.cpp
```diff
@@ -399,233 +399,152 @@ bool LogDev::flush_if_needed(int64_t threshold_size) {
...
+    std::unique_lock mg{m_meta_mutex};
+
+    logdev_key min_safe_ld_key = logdev_key::out_of_bound_ld_key();
+    // Walk through all the stores and find the least logdev_key that we can truncate
+    for (auto& [store_id, store] : m_id_logstore_map) {
+        auto lstore = store.log_store;
+        if (lstore == nullptr) { continue; }
+        auto const [trunc_lsn, trunc_ld_key, tail_lsn] = lstore->truncate_info();
+        if (trunc_lsn == tail_lsn) {
+            THIS_LOGDEV_LOG(DEBUG, "Store_id={} didn't have any writes since last truncation, skipping ", store_id);
```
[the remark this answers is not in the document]
JacksonYao287: tail_lsn is the latest lsn that has been flushed.
when we truncate to tail_lsn(truncate all the flushed lsn), then start_lsn will be tail_lsn + 1

src/lib/logstore/log_dev.cpp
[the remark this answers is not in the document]
JacksonYao287: nice catch! that is possible. we can put `fetch_add` before `m_log_records->create` to solve this

src/lib/logstore/log_dev.cpp
```diff
@@ -399,233 +399,152 @@ bool LogDev::flush_if_needed(int64_t threshold_size) {
...
+    // There are no writes or no truncation called for any of the store, so we can't truncate anything
+    if (min_safe_ld_key == logdev_key::out_of_bound_ld_key()) { return 0; }
 
-                return;
-            }
-        }
-        sisl::VectorPool< flush_blocked_callback >::free(flush_q);
-    }
-    m_is_flushing.store(false, std::memory_order_release);
+    uint64_t const num_records_to_truncate = uint64_cast(min_safe_ld_key.idx - m_last_truncate_idx);
```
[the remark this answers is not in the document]
JacksonYao287: I think you are right. we only need to compare `min_safe_ld_key.idx` and `m_last_truncate_idx` . if they are equal to each other, we can return 0.

src/lib/logstore/log_dev.cpp
```diff
@@ -399,233 +398,164 @@ bool LogDev::flush_if_needed(int64_t threshold_size) {
...
+        req_map[idx] = std::make_pair(req, log_store);
+    }
+    HISTOGRAM_OBSERVE(logstore_service().m_metrics, logdev_flush_time_us,
+                      get_elapsed_time_us(m_last_flush_time, done_time));
     HISTOGRAM_OBSERVE(logstore_service().m_metrics, logdev_post_flush_processing_latency,
-                      get_elapsed_time_us(lg->m_post_flush_msg_rcvd_time, lg->m_post_flush_process_done_time));
+                      get_elapsed_time_us(done_time));
     free_log_group(lg);
-    unlock_flush();
+    m_log_records->truncate(upto_indx);
```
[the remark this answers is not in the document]
JacksonYao287: since it is has been flushed and later read will read from vdev , we do not need to keep this in memory, so it can be truncated

src/lib/logstore/log_dev.cpp
```diff
@@ -399,233 +398,164 @@ bool LogDev::flush_if_needed(int64_t threshold_size) {
...
-    }
-    lg->m_post_flush_process_done_time = Clock::now();
-
-    HISTOGRAM_OBSERVE(logstore_service().m_metrics, logdev_flush_done_msg_time_ns,
-                      get_elapsed_time_us(lg->m_flush_finish_time, lg->m_post_flush_msg_rcvd_time));
+        logstore_req* req = s_cast< logstore_req* >(record.context);
+        HomeLogStore* log_store = req->log_store;
+        HS_LOG_ASSERT_EQ(log_store->get_store_id(), record.store_id,
+                         "Expecting store id in log store and flush completion to match");
+        HISTOGRAM_OBSERVE(logstore_service().m_metrics, logstore_append_latency, get_elapsed_time_us(req->start_time));
```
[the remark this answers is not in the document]
JacksonYao287: ack

src/tests/test_log_store.cpp
```diff
@@ -744,42 +710,37 @@ class LogStoreTest : public ::testing::Test {
...
+        if (expect_forward_progress) {
+            // we have multiple log_store for each logdev, so we need to update the highest truncated lsn and then do
+            // the verification
+            for (const auto& lsc : SampleDB::instance().m_log_store_clients) {
+                auto fid = lsc->m_log_store->get_logdev()->get_id();
+                auto trunc_loc = lsc->m_log_store->get_trunc_ld_key();
+                if (trunc_loc == logdev_key::out_of_bound_ld_key())
+                    LOGINFO("No forward progress for device truncation yet.")
+                if (m_truncate_log_idx.count(fid) && trunc_loc.idx > m_truncate_log_idx[fid].load())
+                    m_truncate_log_idx[fid].store(trunc_loc.idx);
```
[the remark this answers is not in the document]
JacksonYao287: >m_truncate_log_idx.count(fid) && trunc_loc.idx > m_truncate_log_idx[fid].load())
we already have this, so no need to add an extra assert here

src/tests/test_log_store_long_run.cpp
```diff
@@ -471,13 +454,8 @@ class LogStoreLongRun : public ::testing::Test {
...
         }
 
-        logstore_service().device_truncate(
-            [this](auto& trunc_lds) {
-                for (auto [fid, trunc_loc] : trunc_lds) {
-                    m_truncate_log_idx[fid].store(trunc_loc.idx);
-                }
-            },
-            true /* wait_till_done */);
+        logstore_service().device_truncate();
```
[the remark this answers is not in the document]
JacksonYao287: seems the current code  in master branch does not validate, so I just keep the same

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -163,26 +173,37 @@ ulong HomeRaftLogStore::append(nuraft::ptr< nuraft::log_entry >& entry) {
...
 }
 
 void HomeRaftLogStore::end_of_append_batch(ulong start, ulong cnt) {
-    store_lsn_t end_lsn = to_store_lsn(start + cnt - 1);
-    m_log_store->flush_sync(end_lsn);
+    auto end_lsn = to_store_lsn(start + cnt - 1);
+    m_log_store->flush(end_lsn);
     m_last_durable_lsn = end_lsn;
+    std::unique_lock lk(m_mutex);
+    m_cached_log_entries.clear();
```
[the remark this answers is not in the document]
JacksonYao287: OK，will change this to a ring buffer

src/tests/test_log_store.cpp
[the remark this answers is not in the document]
JacksonYao287: the test of master branch has only one log store of each log dev. in my PR , I make multiple log_stores on one log_dev.

so trunc_loc.idx of log_store_1 might be lower than that of log_store_2, so this assert is not always correct

src/lib/logstore/log_dev.cpp
```diff
@@ -399,242 +398,176 @@ bool LogDev::flush_if_needed(int64_t threshold_size) {
...
+
     HISTOGRAM_OBSERVE(logstore_service().m_metrics, logdev_flush_records_distribution, lg->nrecords());
     HISTOGRAM_OBSERVE(logstore_service().m_metrics, logdev_flush_size_distribution, lg->actual_data_size());
-    THIS_LOGDEV_LOG(TRACE, "vdev offset={} log group total size={}", lg->m_log_dev_offset, lg->header()->total_size());
 
-    // write log
-    m_vdev_jd->async_pwritev(lg->iovecs().data(), int_cast(lg->iovecs().size()), lg->m_log_dev_offset)
-        .thenValue([this, lg](auto) { on_flush_completion(lg); });
+    // FIXME:: add logic to handle this error in upper layer
+    if (m_vdev_jd->sync_pwritev(lg->iovecs().data(), int_cast(lg->iovecs().size()), lg->m_log_dev_offset)) return false;
```
xiaoxichen, line 451: can we have an ERROR log here?

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -146,13 +163,20 @@ ulong HomeRaftLogStore::append(nuraft::ptr< nuraft::log_entry >& entry) {
...
+        m_log_entry_cache[position_in_cache] = std::make_pair(lsn, entry);
+    }
+    return lsn;
 }
 
 void HomeRaftLogStore::write_at(ulong index, nuraft::ptr< nuraft::log_entry >& entry) {
     auto buf = entry->serialize();
 
-    m_log_store->rollback_async(to_store_lsn(index) - 1, nullptr);
+    m_log_store->rollback(to_store_lsn(index) - 1);
```
xiaoxichen, line 179: here and also in the rollback, do we need to invalidate cache?

src/lib/replication/log_store/home_raft_log_store.h
```diff
@@ -199,6 +199,10 @@ class HomeRaftLogStore : public nuraft::log_store {
     nuraft::ptr< nuraft::log_entry > m_dummy_log_entry;
     store_lsn_t m_last_durable_lsn{-1};
     folly::Future< folly::Unit > m_log_store_future;
+
+    // raft log entry cache related members
+    std::shared_mutex m_mutex;
```
xiaoxichen, line 204: can we use standard mutex , or we need folly mutex for fiber?

xiaoxichen (commented):
lgtm in general.

The `HomeRaftLogStore` io path has a lot of duplicate code across entry_at/term_at/last_entry etc,  better to consolidate those codes.

src/lib/replication/log_store/home_raft_log_store.h
JacksonYao287: this is a mutex in c++ std library , which can be used as a RW lock. we can not use fiber mutex, which implements locks between different fibers in a thread.

home_raft_log_store related function will never run in  fiber, i will run in a separate thread in nuraft.

src/lib/logstore/log_dev.cpp
JacksonYao287: ack

sanebay (approved):
LGTM

src/include/homestore/logstore/log_store.hpp
```diff
@@ -357,21 +265,12 @@ class HomeLogStore : public std::enable_shared_from_this< HomeLogStore > {
     log_req_comp_cb_t m_comp_cb;
     log_found_cb_t m_found_cb;
     log_replay_done_cb_t m_replay_done_cb;
-    std::atomic< logstore_seq_num_t > m_seq_num;
+    std::atomic< logstore_seq_num_t > m_start_lsn;
```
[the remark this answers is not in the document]
JacksonYao287: ok

src/lib/logstore/log_dev.cpp
```diff
@@ -674,7 +598,7 @@ std::shared_ptr< HomeLogStore > LogDev::create_new_log_store(bool append_mode) {
         HS_REL_ASSERT((it == m_id_logstore_map.end()), "store_id {}-{} already exists", m_logdev_id, store_id);
         m_id_logstore_map.insert(std::pair(store_id, logstore_info{.log_store = lstore, .append_mode = append_mode}));
     }
-    HS_LOG(INFO, logstore, "Created log store log_dev={} log_store={}", m_logdev_id, store_id);
+    LOGINFO("Created log store log_dev={} log_store={}", m_logdev_id, store_id);
```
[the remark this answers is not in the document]
JacksonYao287: ack

src/lib/logstore/log_dev.cpp
```diff
@@ -737,137 +638,30 @@ void LogDev::on_log_store_found(logstore_id_t store_id, const logstore_superblk&
         return;
     }
 
-    LOGINFO("Found a logstore log_dev={} log_store={} with start seq_num={}, Creating a new HomeLogStore instance",
+    LOGINFO("Found a logstore log_dev={} log_store={} with start lsn={}, Creating a new HomeLogStore instance",
```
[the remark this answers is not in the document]
JacksonYao287: ack

sanebay (approved):

PR #481, opened 2024-08-05
Set lsn log replay

raakella1:
Repl Dev now provides a callback through listener for the consumer to do any recovery before logs are replayed.

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -1162,6 +1162,7 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
     rreq->init(rkey, jentry->code, false /* is_proposer */, entry_to_hdr(jentry), entry_to_key(jentry),
                (entry_blkid.blk_count() * get_blk_size()));
     rreq->set_local_blkid(entry_blkid);
+    rreq->set_lsn(lsn);
 
     // 2. Pre-commit the log entry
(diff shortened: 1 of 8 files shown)


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -771,8 +771,10 @@ void RaftReplDev::handle_commit(repl_req_ptr_t rreq) {
         m_listener->on_commit(rreq->lsn(), rreq->header(), rreq->key(), rreq->local_blkid(), rreq);
     }
 
-    auto prev_lsn = m_commit_upto_lsn.exchange(rreq->lsn());
-    RD_DBG_ASSERT_GT(rreq->lsn(), prev_lsn, "Out of order commit of lsns, it is not expected in RaftReplDev");
+    if (!recovery) {
```
xiaoxichen, line 774: I didnt get this part.  
1. during recovery,  the `last_commit_index()` returns `m_commit_upto_lsn` so log reply will start from the persisted m_commit_upto_lsn, why we will trigger this assert?

2. during recovery , why we dont want to forward the `m_commit_upto_lsn` as we done replying each log?

raakella1: m_commit_upto_lsn is recovered directly from super block, it is not changed as we replay logs

xiaoxichen: I got it after your explanation , we replay from cp_lsn which is always less than ocmmit_lsn.


Better to add a few line of comment.

xiaoxichen (approved):

PR #495, opened 2024-08-13
fix create chunk

JacksonYao287:
if we can not create chunk on one pdev , we should try the next one.

### DIFF
--- a/src/lib/device/device_manager.cpp
+++ b/src/lib/device/device_manager.cpp
@@ -440,7 +440,13 @@ shared< Chunk > DeviceManager::create_chunk(HSDevType dev_type, uint32_t vdev_id
     // Create a chunk on any pdev of device type.
     for (const auto& dev : pdevs) {
         // Ordinal added in add_chunk.
-        chunk = dev->create_chunk(chunk_id, vdev_id, chunk_size, 0 /* ordinal */, data);
+        try {
+            chunk = dev->create_chunk(chunk_id, vdev_id, chunk_size, 0 /* ordinal */, data);
+        } catch (std::out_of_range const& e) {
+            HS_LOG(DEBUG, device, "can not create new chunk on dev {}, try next one!", dev->get_devname());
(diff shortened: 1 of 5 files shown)


xiaoxichen (commented):
lgtm

xiaoxichen (approved):

shosseinimotlagh (approved):

src/lib/device/device_manager.cpp
```diff
@@ -440,7 +440,13 @@ shared< Chunk > DeviceManager::create_chunk(HSDevType dev_type, uint32_t vdev_id
     // Create a chunk on any pdev of device type.
     for (const auto& dev : pdevs) {
         // Ordinal added in add_chunk.
-        chunk = dev->create_chunk(chunk_id, vdev_id, chunk_size, 0 /* ordinal */, data);
+        try {
+            chunk = dev->create_chunk(chunk_id, vdev_id, chunk_size, 0 /* ordinal */, data);
```
sanebay, line 444: is this the actual fix ? dont we create_chunk only based on num_chunks = size / chunk_size ?

JacksonYao287: sorry , I do not get your piont.
the problem here is that if we do not make this changes, when the first pdev can not create more chunk, it will throw out an exception and if we do not catch the exception , we will have no chance to create chunk in other pdev.

this PR is to make sure the next pdev can be selected to create chunk if the previous one can not

xiaoxichen: i think the actual fix to the issue is changing the index UT to use fast dev.

the  try-catch stuff we dont expect to run into this at all,  but it is logically right.

JacksonYao287: yes, when I test log store, i have ever met this issue. when the first pdev is used up and can not create more chunk, it will not try the next one. that is how I found this issue

Issue #414, opened 2024-05-10
Fetch Blks rpc should always fetch the blks from consumer


hkadayam:
Fetch Blocks RPC when sent to existing member, responds with the data from HomeStore layer itself. The RPC sent all the neccessary information, <dsn, lsn, remote_blkid, key, header>. At present homestore uses remote_blkid (which is local_blkid for this node) and read the data from data_service and fill the data.

In case of HomeObject, this is an issue, because the blkid could have been GC'd and moved to a different block location. The GC does not go through the replication/journal. As a result, reading the blkid could corrupt the follower's data. So homestore needs to use header, key -> pass it on to the consumer and let consumer provide the latest data (just like read_blob path) and respond back to homestore.

xiaoxichen:
This is needed and become a bug once we have GC. 

Pushing this a bit later as GC is pushing to later milestone.

PR #496, opened 2024-08-13
Remove req from lsn map when committing.

xiaoxichen:


sanebay (approved):

sanebay:
Did your test shows memory usage not increasing via grafana ? Whats the current usage ?

xiaoxichen:
yes the homestore long running UT shows stable memory usage. Will continue to check on SM level.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -758,6 +758,8 @@ void RaftReplDev::handle_commit(repl_req_ptr_t rreq, bool recovery) {
 
     // Remove the request from repl_key map.
     m_repl_key_req_map.erase(rreq->rkey());
+    // Remove the request from lsn map.
+    m_state_machine->unlink_lsn_to_req(rreq->lsn());
```
JacksonYao287, line 762: if we do this , we will lose the entry cache in leader.
what`s the problem if remove this.
https://github.com/eBay/HomeStore/blob/72a355e80a8e8368b555637e491840bc6307eb77/src/lib/replication/repl_dev/raft_repl_dev.cpp#L1092

can the reqs in the leader be removed as those in follower when they are expired?

xiaoxichen: this is not used anywhere except in `end_of_append_batch`  ATM.

xiaoxichen: I think the `entry cache` is separated into two part - log and data.
Log is on logstore, as discussed in your logstore PR if the cache you add can be extend and cache last N records, regardless of flush.

Data is on the path of `RaftReplDev::on_fetch_data_received`, right now it just read from drive based on blkid. As https://github.com/eBay/HomeStore/issues/414 stated this approach is wrong after we introduce GC. If the fetch need to go through Index it goes through HO,  I think we may have a unified cache in HO which can serve customer IO as well.

JacksonYao287 (approved):

PR #509, opened 2024-08-16
set rreq time before moving into the map

raakella1:


xiaoxichen (approved):
lgtm.

I still strongly believe we should update the timestamp in `add_state`which makes it self-contain.

PR #510, opened 2024-08-16
Add batching in UT to support million ios for baseline test.

sanebay:
Add batching of read and write values in UT.
Keep mapping a index to key value pairs for quick lookup during snapshot read.
Add shutdown and start to simulate follower down, writes complete, follower start. Earlier we used to restart with sleep, but for 1 million io's or different values of num_io's we dont know how much to sleep. 

To test
./test_raft_repl_dev --gtest_filter=RaftReplDevTest.BaselineTest --num_io=1000000 --log_mods=replication:debug --config_path ./config --dev_size_mb=2048600 --snapsho
t_distance=0

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.4.51"
+    version = "6.4.52"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/lib/replication/log_store/home_raft_log_store.cpp
+++ b/src/lib/replication/log_store/home_raft_log_store.cpp
@@ -315,7 +315,8 @@ bool HomeRaftLogStore::compact(ulong compact_lsn) {
         // release this assert if for some use case, we should tolorant this case;
         // for now, don't expect this case to happen.
         // RELEASE_ASSERT(false, "compact_lsn={} is beyond the current max_lsn={}", compact_lsn, cur_max_lsn);
-
+        REPL_STORE_LOG(DEBUG, "Adding dummy entries during compact from={} upto={}", cur_max_lsn + 1,
+                       to_store_lsn(compact_lsn));
         // We need to fill the remaining entries with dummy data.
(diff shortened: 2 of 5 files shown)


xiaoxichen:
I think we can have a general test case for both baseline and incremental, the pattern is similar, only difference is if
`  X> #snapshot_to_keep * snapshot_distance ? mode = baseline : mode= incremental`
```
while(#IO < TARGET_IOS_TO_RUN) {
  1. write some data
  2. shutdown one replica
  3. write  X reqs
  4. start up the down replica.
  [5. optionally ]keep writing during start up
  6. wait the replica sync.
}
//Lets have a huge one
erase data on one replica, maybe go through format again?  and let is resync from begining.
```

It needs a night for long running,  hope we can exercise recovery more times vs exercise a recovery with big gap.

sanebay:
> 
> It needs a night for long running, hope we can exercise recovery more times vs exercise a recovery with big gap.

Yeah this is a nice test case to add. We can add this in our SM long running. raft_repl UT is more like functionality test with smaller and larger number of inputs.

xiaoxichen:
in SM it is hard to ensure we go through incremental or baseline.   I think it is a good waste to generate 1M IOs and only test one recovery:)

yamingk:
> I think we can have a general test case for both baseline and incremental, the pattern is similar, only difference is if ` X> #snapshot_to_keep * snapshot_distance ? mode = baseline : mode= incremental`
> 
> ```
> while(#IO < TARGET_IOS_TO_RUN) {
>   1. write some data
>   2. shutdown one replica
>   3. write  X reqs
>   4. start up the down replica.
>   [5. optionally ]keep writing during start up
>   6. wait the replica sync.
> }
> //Lets have a huge one
> erase data on one replica, maybe go through format again?  and let is resync from begining.
> ```
> 
> It needs a night for long running, hope we can exercise recovery more times vs exercise a recovery with big gap.

In hour HomeStore 4.x replication long running, we definitely should have test case running recovery a few hundred times (we also do it with nublox 1.3 homestore long run) during the night time, (with I/Os running before and after recovery) and keep rebooting one replica. Our goal is to "abuse" HomeStore so badly so it won't "abuse" us in production.

yamingk (approved):

shosseinimotlagh (approved):

PR #514, opened 2024-08-20
Track timestamp of each state in repl_dev_ctx.

xiaoxichen:
With this we can easily get time diff of each given state, avoid doing the calculation/analysis from grep of logs.

### DIFF
--- a/src/lib/replication/repl_dev/common.cpp
+++ b/src/lib/replication/repl_dev/common.cpp
@@ -146,7 +146,10 @@ bool repl_req_ctx::save_fetched_data(sisl::GenericClientResponse const& fetched_
     return true;
 }
 
-void repl_req_ctx::add_state(repl_req_state_t s) { m_state.fetch_or(uint32_cast(s)); }
+void repl_req_ctx::add_state(repl_req_state_t s) {
+    m_state.fetch_or(uint32_cast(s));
+    m_timestamps[std::bit_width(uint32_cast(s))] = Clock::now();
+}
 
 bool repl_req_ctx::add_state_if_not_already(repl_req_state_t s) {
     bool changed{false};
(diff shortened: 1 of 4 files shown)


src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -222,8 +222,6 @@ void RaftStateMachine::unlink_lsn_to_req(int64_t lsn) {
 void RaftStateMachine::link_lsn_to_req(repl_req_ptr_t rreq, int64_t lsn) {
     rreq->set_lsn(lsn);
     rreq->add_state(repl_req_state_t::LOG_RECEIVED);
-    // reset the rreq created_at time to now https://github.com/eBay/HomeStore/issues/506
-    rreq->set_created_time();
```
yamingk, line 226: is this still needed?

src/lib/replication/repl_dev/common.cpp
```diff
@@ -199,7 +212,7 @@ std::string repl_req_ctx::to_compact_string() const {
 }
 
 bool repl_req_ctx::is_expired() const {
-    return get_elapsed_time_sec(m_start_time) > HS_DYNAMIC_CONFIG(consensus.repl_req_timeout_sec);
+    return time_from(repl_req_state_t::LOG_RECEIVED) > HS_DYNAMIC_CONFIG(consensus.repl_req_timeout_sec);
```
yamingk, line 215: if log received, before follower received the data (rreq will also be created), originator is gone, we should also clean up the rreqs, right?

xiaoxichen: yes right now (the PR) is trying to be compatible with original implementation for `is_expired`,  mainly to bring in all the timestamp we have as a foundation  of next step 

I think we discussed the GC stuff a bit today and that need a seperate design, we can visit that later.

src/lib/replication/repl_dev/raft_state_machine.cpp
xiaoxichen: no as I changed the `is_expired()` implementation.

zhiteng:
Do we still want this change?  I'm cleaning stale/idle PRs.

PR #515, opened 2024-08-20
enable home raft log store UT

JacksonYao287:
1 enable home_raft_log_store UT and add it to docker file and nightly runing
2 remove android directory in github CI in the stage of building cache(previous in create-and-test-package) ,so the space in github CI VM will be definitely freed for every pipeline, since one of CI pipeline does not have create-and-test-package .
3 increase the disk size to 2GB for UT
4 fix log dev flush bug. add a loop to create multiple logGroup to make sure all the expected logs flushed .
5 fix a bug in home_Raft_log_store#pack to make sure the available size of packing buffer is able to hold entry.size() and the length of this entry
6 use truncation instead of flush when home_raft_log_store#compact. this is only a in-memory change. the real truncation will be scheduled by resource manager. also create issue to handle the start_index case when recovery from a crash #530, wo do it in a separate PR.
7 add commit_config log so we can see when a config change is made.
8 fix a bug in handle_raft_event. we need to indentify whether a log entry should be appended to log store. if the req#lsn is not -1, it means this log has been localized and appended before, we should skip it.
9 add a force_leave api for now to handle the case that a follower and a leader has destroyed the raft goup , but the second follower fail to receive this message and will stuck. this is used for fixing raft_repl_dev UT. we can revisit here later if necessary.
10 fix the bug of raft_repl_dev UT when setting last_committed_idx in write_Snapshot_data in follower side. we should get the commited_index from raft_server itself. in our UT, commit_config is not taken account when increasing the commit_count.
11 fix bug bug of raft_repl_dev UT in read_snap_shot of leader. when we search the next start lsn which is the start of sending data. we can not use std::map::find , since if the requested next_lsn is a config_change , it will not be put into kvDB(lsn_index_) and as a result , std::map::find wil return std::end() and nothing will be sent to follower. instead , we should use low_bound, so that we can get the first data kv to be sent , config will be skipped.


src/lib/logstore/log_dev.cpp
```diff
@@ -435,32 +435,36 @@ bool LogDev::flush() {
...
-    off_t offset = m_vdev_jd->alloc_next_append_blk(lg->header()->total_size());
-    lg->m_log_dev_offset = offset;
-
-    HS_REL_ASSERT_NE(lg->m_log_dev_offset, INVALID_OFFSET, "log dev is full");
-    THIS_LOGDEV_LOG(TRACE, "Flushing log group data size={} at offset={} log_group={}", lg->actual_data_size(), offset,
-                    *lg);
-
-    HISTOGRAM_OBSERVE(logstore_service().m_metrics, logdev_flush_records_distribution, lg->nrecords());
-    HISTOGRAM_OBSERVE(logstore_service().m_metrics, logdev_flush_size_distribution, lg->actual_data_size());
+    for (; m_last_flush_idx < new_idx;) {
```
xiaoxichen, line 438: I believe this change is driven by UT, could you share what is the case we need the `for loop` here and put it into the comment of code?

JacksonYao287: the amount of logs which one logGroup can flush  has a upper limit.
pls see https://github.com/eBay/HomeStore/blob/314ee9ac42e7df2b4b0888554fe251aa42008b63/src/lib/logstore/log_group.cpp#L76


here we want to make sure all the log that need to be flushed will be definitely flushed , so we need this loop to create multiple log group if necessary

JacksonYao287: cc @sanebay

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -321,20 +321,7 @@ bool HomeRaftLogStore::compact(ulong compact_lsn) {
...
-    // this will be friendly for multiple logstore on same logdev;
-#ifdef _PRERELEASE
-    if (iomgr_flip::instance()->test_flip("force_home_raft_log_truncate")) {
-        REPL_STORE_LOG(TRACE, "Flip force_home_raft_log_truncate is enabled, force truncation, compact_lsn={}",
-                       compact_lsn);
-        m_log_store->truncate(to_store_lsn(compact_lsn));
-    }
-#endif
-
+    m_log_store->truncate(to_store_lsn(compact_lsn));
```
JacksonYao287, line 324: in compact here , we need to truncate , which will update the start_index. flush will not update start_index

.github/workflows/build_dependencies.yml
```diff
@@ -233,6 +232,7 @@ jobs:
     - name: Code Coverage Run
       run: |
         pre=$([[ "${{ inputs.build-type }}" != "Debug" ]] && echo "-o sisl:prerelease=${{ inputs.prerelease }}" || echo "")
+        sudo rm -rf $ANDROID_HOME
```
JacksonYao287, line 235: in the failure CI , `Create and Test Package` is not scheduled ,so `sudo rm -rf /usr/local/lib/android/` will not be executed.

here , we add this to `Code Coverage Run` will be definitely scheduled in every CI pipeline

@sanebay

.github/workflows/build_dependencies.yml
```diff
@@ -165,6 +165,7 @@ jobs:
     - name: Build Cache
       run: |
         pre=$([[ "${{ inputs.build-type }}" != "Debug" ]] && echo "-o sisl:prerelease=${{ inputs.prerelease }}" || echo "")
+        sudo rm -rf $ANDROID_HOME
```
JacksonYao287, line 168: in the failure CI , Create and Test Package is not scheduled ,so `sudo rm -rf /usr/local/lib/android/` will not be executed in that CI, which might case a no-space left issue.

here , I add this to `Build Cache` phrase, which will be definitely scheduled in every CI pipeline

@sanebay

xiaoxichen (approved):
lgtm
3 issues are solved:

1.  clean up the android home so we can enlarge disk size to 2GB
2. add `for loop` for flush, this make sense based on @JacksonYao287 's explanation and it is perfect we catch this in the re-enabled ut,
3. for compact() call we do compact and move the flush into compact to avoid flushing logs before the start_index

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -264,8 +264,8 @@ raft_buf_ptr_t HomeRaftLogStore::pack(ulong index, int32_t cnt) {
         [this, &out_buf, &remain_cnt]([[maybe_unused]] store_lsn_t cur, const log_buffer& entry) mutable -> bool {
             if (remain_cnt-- > 0) {
                 size_t avail_size = out_buf->size() - out_buf->pos();
-                if (avail_size < entry.size()) {
-                    avail_size += std::max(out_buf->size() * 2, (size_t)entry.size());
+                if (avail_size < entry.size() + sizeof(uint32_t)) {
```
JacksonYao287, line 267: avail_size should be able to hold entry.size() and the length of this entry, which is a uint32

src/lib/logstore/log_dev.cpp
sanebay: But even if we limit, it will be flushed in the next flush. Here if we flush in for loop, how is the ordering and flush completions works ? It becomes too complicated. Shouldnt you test make it work for this.

src/lib/logstore/log_store.cpp
```diff
@@ -177,6 +177,7 @@ void HomeLogStore::on_log_found(logstore_seq_num_t seq_num, const logdev_key& ld
 
 void HomeLogStore::truncate(logstore_seq_num_t upto_lsn, bool in_memory_truncate_only) {
     if (upto_lsn < m_start_lsn) { return; }
+    flush();
```
sanebay, line 180: why do we need a flush here ?

src/lib/replication/log_store/home_raft_log_store.cpp
sanebay: This is done purposefully. We let resource manager do truncation @yamingk  can confirm this.

src/tests/test_common/homestore_test_common.hpp
```diff
@@ -46,7 +46,7 @@ SISL_OPTION_GROUP(
      "number"),
     (num_devs, "", "num_devs", "number of devices to create", ::cxxopts::value< uint32_t >()->default_value("3"),
      "number"),
-    (dev_size_mb, "", "dev_size_mb", "size of each device in MB", ::cxxopts::value< uint64_t >()->default_value("1024"),
+    (dev_size_mb, "", "dev_size_mb", "size of each device in MB", ::cxxopts::value< uint64_t >()->default_value("2048"),
```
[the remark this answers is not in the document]
sanebay: With latest PR, it should work with 1GB also.

src/tests/test_raft_repl_dev.cpp
```diff
@@ -878,14 +878,6 @@ TEST_F(RaftReplDevTest, BaselineTest) {
     LOGINFO("Homestore replica={} setup completed", g_helper->replica_num());
     g_helper->sync_for_test_start();
```
sanebay, line 880: We need this as we are not truncating when raft calls compact

src/lib/logstore/log_dev.cpp
JacksonYao287: only `logstore#on_write_completion` , which will be called after flushing a logGroup, will update the `m_tail_lsn`. 
when we truncate, we can not truncate exceeding `m_tail_lsn`.  

for example , we have a log store and we append upto lsn 100. now, if we flush only to 60, then  we can not truancate exceeding 60 , for example,  80. now , if we want to truncate to 80, we need to flush exceeding 80.

pls see
https://github.com/eBay/HomeStore/blob/3030851cebe24b99a070934bc67c5414a819a359/src/lib/logstore/log_store.cpp#L191

> how is the ordering and flush completions works

I do not get this question. when any logGroup is flushed, the flush completion will work for all the entry in the group.

src/lib/replication/log_store/home_raft_log_store.cpp
JacksonYao287: then , then start_index will not be updated on time, which will probably be wrong to upper layer.
for the view of upper layer , if it compact to lsn n , then the start_index it can see after campact should be lsn + 1.
so , I believe we should do truncate here, and not depend on resource manager

src/tests/test_common/homestore_test_common.hpp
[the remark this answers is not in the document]
JacksonYao287: so ,do you want this change? both are ok for me

src/tests/test_raft_repl_dev.cpp
JacksonYao287: let`s first make a decision on this topic
https://github.com/eBay/HomeStore/pull/515/files/c3e0ab3f3210c42a63b6b5f6a97386115f1126a9#r1724163097

then we can revisit this. if we need do a real truncate, then this can be removed

src/lib/logstore/log_store.cpp
JacksonYao287: if the caller does not do a explicitly flush before truncating, then `m_tail_lsn` will not be updated and then truncating might not be able to truncate to the expected on.
add a `flush` here to make sure we do a flush before truncating. if flush is already scheduled by the called before truncating. then the flush here will do nothing and just return.

src/tests/test_common/homestore_test_common.hpp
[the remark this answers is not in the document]
xiaoxichen: if env can adopt 2GB maybe we keep 2GB, no harm to do this

src/lib/logstore/log_dev.cpp
xiaoxichen:
```
    /**
     * @brief Flush this log store (write/sync to disk) up to the sequence number
     *
     * @param seq_num Sequence number upto which logs are to be flushed. If not provided, will wait to flush all seq
     * numbers issued prior.
     */
    void flush(logstore_seq_num_t upto_seq_num = invalid_lsn());

```

I think the `HomeLogStore::flush`  definition state that we need to flush up to the given LSN, not best effort.  So we need to fix this somewhere.  Also the `LogDev::flush()` doesnt take any argument,  usually that means flush to the point when it was called, like the `std::flush()`

one step back, if we dont change `LogDev::flush()`,  we need to do like below in `HomeLogStore::flush` which is much less optimal 
```
 while (m_records.status(upto_lsn).is_active) {
     m_logdev->flush_under_guard();
     sleep(xxx);
}
```

src/lib/replication/log_store/home_raft_log_store.cpp
sanebay: This needs to be discussed with @yamingk

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1050,7 +1050,13 @@ std::pair< bool, nuraft::cb_func::ReturnCode > RaftReplDev::handle_raft_event(nu
                 if (entry->get_val_type() != nuraft::log_val_type::app_log) { continue; }
                 if (entry->get_buf_ptr()->size() == 0) { continue; }
                 auto req = m_state_machine->localize_journal_entry_prepare(*entry);
-                if (req == nullptr) {
+                // TODO :: we need to indentify whether this log entry should be appended to log store.
```
xiaoxichen, line 1053: is #1 safe here ? as we dont have term in the rreq, how can we ensure this is not a re-written

src/lib/replication/log_store/home_raft_log_store.cpp
yamingk, line 267: can you convert your comment to code comment also?

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -264,8 +264,8 @@ raft_buf_ptr_t HomeRaftLogStore::pack(ulong index, int32_t cnt) {
...
+                    avail_size += std::max(out_buf->size() * 2, (size_t)entry.size() + sizeof(uint32_t));
```
yamingk, line 268: nit: maybe having a `size_t const entry_size =  entry.size() + sizeof(uint32_t);` before line 265, and replace the places with this one.

src/tests/test_raft_repl_dev.cpp
```diff
@@ -421,11 +420,23 @@ class RaftReplDevTest : public testing::Test {
...
                 std::this_thread::sleep_for(std::chrono::seconds(1));
                 auto& raft_repl_svc = dynamic_cast< RaftReplService& >(hs()->repl_service());
                 raft_repl_svc.gc_repl_devs();
                 LOGINFO("Waiting for repl dev to get destroyed");
+
+                // TODO: if leader is destroyed, but the follower does not receive the notification, it will not be
+                // destroyed for ever. we need handle this in raft_repl_dev. revisit here after making changes at
+                // raft_repl_dev side to hanle this case. this is a workaround to avoid the infinite loop for now.
+                if (i++ > 10 && !force_leave) {
+                    LOGWARN("Waiting for repl dev to get destroyed and it is leader, so do a force leave");
```
xiaoxichen, line 435: the log is not accurate as this will be run on every replica, not only leader.

sanebay (approved):

src/lib/replication/log_store/home_raft_log_store.cpp
JacksonYao287: ack ,will do it in the later pr

src/tests/test_raft_repl_dev.cpp
JacksonYao287: yes , it a mistake, it can be run on any replica. will change this in the later PR

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: we have term in rreq. every rreq is identified by {originator, term , dsn}

PR #521, opened 2024-08-22
Inlined data

raakella1:
- Do not allocate blks if data is not inlined
- Implement a basic version of is_resync_mode where we declare the follower to be in resync mode if the diff between leader_commit_lsn and his own last_log_idx from logstore is greater than a threshold

### DIFF
--- a/src/lib/common/homestore_config.fbs
+++ b/src/lib/common/homestore_config.fbs
@@ -261,6 +261,9 @@ table Consensus {
 
     // Frequency to flush durable commit LSN in millis
     flush_durable_commit_interval_ms: uint64 = 500;
+
+    // Log difference to determine if the follower is in resync mode
+    resync_log_idx_threshold: int64 = 100;
 }
 
 table HomeStoreSettings {
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -526,7 +526,7 @@ void RaftReplDev::check_and_fetch_remote_data(std::vector< repl_req_ptr_t > rreq
 void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
     if (rreqs.size() == 0) { return; }
 
-    std::vector< ::flatbuffers::Offset< RequestEntry > > entries;
+    std::vector<::flatbuffers::Offset< RequestEntry > > entries;
     entries.reserve(rreqs.size());
 
     shared< flatbuffers::FlatBufferBuilder > builder = std::make_shared< flatbuffers::FlatBufferBuilder >();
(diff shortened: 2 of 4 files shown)


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1166,12 +1171,17 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
...
     auto rreq = it->second;
     RD_DBG_ASSERT(happened, "rreq already exists for rkey={}", rkey.to_string());
-    MultiBlkId entry_blkid;
-    entry_blkid.deserialize(entry_to_val(jentry), true /* copy */);
-    rreq->init(rkey, jentry->code, false /* is_proposer */, entry_to_hdr(jentry), entry_to_key(jentry),
-               (entry_blkid.blk_count() * get_blk_size()));
-    rreq->set_local_blkid(entry_blkid);
+    uint32_t data_size{0u};
+
+    if ((jentry->code == journal_type_t::HS_DATA_LINKED) && (jentry->value_size > 0)) {
```
sanebay, line 1176: we should assert if data linked, value_size should > 0

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1188,4 +1198,11 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
 
 void RaftReplDev::on_restart() { m_listener->on_restart(); }
 
+bool RaftReplDev::is_resync_mode() {
+    int64_t const leader_commited_lsn = raft_server()->get_leader_committed_log_idx();
```
yamingk, line 1202: is `get_leader_committed_log_idx` very light-weight operation (just a read) or it take some mutex while doing it? Asking since this is_resync_mode will be accessed quite frequently in I/O path, so would be good to double confirm.

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: it is already asserted in repl_contxt_t::init(), no?

raakella1: Yes, we assert in the init

src/lib/replication/repl_dev/raft_repl_dev.cpp
raakella1: Its a couple of atomics
https://github.com/eBay/NuRaft/blob/0a754e0b95f9763d9f3585973eaf51d0a413f1cd/include/libnuraft/raft_server.hxx#L475

yamingk: it should be fine.

src/lib/common/homestore_config.fbs
```diff
@@ -261,6 +261,9 @@ table Consensus {
 
     // Frequency to flush durable commit LSN in millis
     flush_durable_commit_interval_ms: uint64 = 500;
+
+    // Log difference to determine if the follower is in resync mode
+    resync_log_idx_threshold: int64 = 100;
```
yamingk, line 266: we might want to tune this number later with E2E env,  I am fine with this number for now.

xiaoxichen (approved):

PR #523, opened 2024-08-23
Issue: 498 meta service long run fix

yamingk:
Problem Desc:
===========
We have a limitation on max chunk size (uint32_t) which is 4GB, which will be carved into portions in blk allocator with each portion maps to tens of thousands of nblks available for each portion. This specific test case specifiy max io size to to limitation that is larger than the portion total nblks and during the test, the io size is randomized and when it comes to larger than the limit, it can't find available allocation in any chunk so return failed.

Fix:
===
1. adjust the test case to not exceed limitation 
2. refactor the code a little to get max chunk size from Chunk class.

Test:
====
Specific failed test (previously failing consistently) can pass with this.

### DIFF
--- a/src/tests/test_scripts/log_meta_test.py
+++ b/src/tests/test_scripts/log_meta_test.py
@@ -81,7 +81,7 @@ def meta_nightly(options, addln_opts):
     subprocess.check_call(options.dirpath + "test_meta_blk_mgr " + cmd_opts + addln_opts, stderr=subprocess.STDOUT,
                           shell=True)
 
-    cmd_opts = "--min_write_size=10485760 --max_write_size=104857600 --bitmap=1"
+    cmd_opts = "--min_write_size=10485760 --max_write_size=80857600 --bitmap=1"
     subprocess.check_call(options.dirpath + "test_meta_blk_mgr " + cmd_opts + addln_opts, stderr=subprocess.STDOUT,
                           shell=True)
 
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.4.49"
+    version = "6.4.50"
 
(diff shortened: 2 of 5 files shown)


src/tests/test_scripts/log_meta_test.py
```diff
@@ -81,7 +81,7 @@ def meta_nightly(options, addln_opts):
     subprocess.check_call(options.dirpath + "test_meta_blk_mgr " + cmd_opts + addln_opts, stderr=subprocess.STDOUT,
                           shell=True)
 
-    cmd_opts = "--min_write_size=10485760 --max_write_size=104857600 --bitmap=1"
+    cmd_opts = "--min_write_size=10485760 --max_write_size=80857600 --bitmap=1"
```
JacksonYao287, line 84: why `80857600`, is it upper limit of total nblks in the portion ?

yamingk: yes, it is a little smaller than the upper limit.

xiaoxichen: JFYI:  it translate to 2.4TB if blk size = 4K when storing bit map:) 

But maybe we are looking for 10x of this or even more.

xiaoxichen (approved):
lgtm

JacksonYao287 (approved):
LGTM

src/tests/test_scripts/log_meta_test.py
yamingk: Thanks @xiaoxichen for bring it up. Yes, my thought is to create only one chunk for meta blk store, and change chunkz-size to uint64_t which should be big enough. 

Another thing is metablkstore has compression feature so the bitmap can be significantly compressed before written.

xiaoxichen: ok if it is per allocator then we are good?

yamingk: yes, we are good as in nublox bitmap.

PR #524, opened 2024-08-24
commit blk during log replay in rep dev

raakella1:


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1181,10 +1186,12 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
     }
 
     rreq->set_lsn(repl_lsn);
+    rreq->set_lentry(lentry);
     rreq->init(rkey, jentry->code, false /* is_proposer */, entry_to_hdr(jentry), entry_to_key(jentry), data_size);
     RD_LOGD("Replay log on restart, rreq=[{}]", rreq->to_string());
 
     if (repl_lsn > m_rd_sb->durable_commit_lsn) {
+        commit_blk(rreq);
```
xiaoxichen, line 1194: Nit: can we dirty the allocator with these blks and still commit them in handle_commit()?. Just in the case for rollback.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1181,10 +1186,12 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
     }
 
     rreq->set_lsn(repl_lsn);
+    rreq->set_lentry(lentry);
```
xiaoxichen, line 1189: This is a bit hacky as we dont set it in normal IO path ,  we believe nuraft will not release them during the whole lifecycle till `commit()`.

I would vote even for normal operations we keep lentry shared_ptr for safety and more consistent behavior.

xiaoxichen (commented):
LGTM.

We didnt catch both bugs in all our restorability UT,  because both bugs occurs if we have `dc_lsn < last_log_lsn` ,  how can we have this UT to exercise this path?

src/lib/replication/repl_dev/raft_repl_dev.cpp
raakella1: Is there a way to dirty the blks without allocating them?

xiaoxichen: not yet but we can change dataservice api?

JacksonYao287: this rreq will be handled in `handle_commit` line 1205, where `commit_blk` will be called.
I don`t think we need another `commit_blk` here

JacksonYao287: also implement roll_back to free blk

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1181,10 +1186,14 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
     }
```
yamingk, line 1187: Not directly related to this PR, but could impact some of the assert in underlying layer, is SM long run currently use debug build or release build?

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: > Nit: can we dirty the allocator with these blks and still commit them in handle_commit()?. Just in the case for rollback.

Just be noted commit_blk will be expecting blks in allocated state for non-recovery mode (existing handle_commit), and for recovery mode, it will try to reserve on cache also before committing to disk; Need a bit more handling to combine them.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1181,10 +1186,14 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
...
 
     rreq->set_lsn(repl_lsn);
+    // keep lentry in scope for the lyfe cycle of the rreq
```
yamingk, line 1189: I probably didn't get why we need to keep this lentry in rreq. Isn't everything copied and saved in rreq, or is there a place that is referring to lentry until rreq finishes?

xiaoxichen:
As discussed in DM,  this PR is not right as there is no guarantee the data has been written to the allocated blks before the restart.

JacksonYao287:
```
        HomeRaftLogStore::end_of_append_batch(start_lsn, count);
        HISTOGRAM_OBSERVE(m_rd.metrics(), raft_end_of_append_batch_latency_us, get_elapsed_time_us(cur_time));

        cur_time = std::chrono::steady_clock::now();
        // Wait for the fetch and write to be completed successfully.
        std::move(fut).wait();
```
we can put `std::move(fut).wait()` before `end_of_append_batch`.

if data is written but log is not flushed, that`s fine, since the block is not committed

xiaoxichen:
yes we can, it is trading regular IO latency vs space leak during recovery, which I am not sure it worth to .

Prior to that, we have to solve the rollback stuff.

src/lib/replication/repl_dev/raft_repl_dev.cpp
raakella1: Debug build

src/lib/replication/repl_dev/raft_repl_dev.cpp
raakella1: homeobject accesses the header during commit which is formed from the lentry

JacksonYao287:
> Prior to that, we have to solve the rollback stuff.

rollback happens when a request is pre_committed but not committed.
1 in normal case, we can free the blk in `on_rollback`.
2 in recovery case , since the blk is not commited (`commit_blk` will be called in on_commit), so blk leak will not happen.
in `on_log_found`, if the lsn of a log  entry is bigger than dc_sln, we can just discard it.

any input?

xiaoxichen (approved):

xiaoxichen:
discussion: https://ebay-eng.slack.com/archives/CAVQXEAKS/p1724866496708579

xiaoxichen (approved):

PR #532, opened 2024-08-29
Fix rollback when rollback to tail_lsn.

xiaoxichen:
previously we dont properly handle this case,
which will error out in https://github.com/eBay/HomeStore/blob/2080ad42e31a9ddf0a41d3a112db4d1c020c45ee/src/lib/logstore/log_store.cpp#L315 as the to_lsn + 1 is invalid in this case.

### DIFF
--- a/src/lib/logstore/log_store.cpp
+++ b/src/lib/logstore/log_store.cpp
@@ -288,15 +288,8 @@ bool HomeLogStore::rollback(logstore_seq_num_t to_lsn) {
 	return true;
     }
 
-    if (to_lsn > m_tail_lsn.load()) {
-        HS_LOG_ASSERT(false, "Attempted to rollback to {} which is larger than tail_lsn {}", to_lsn, m_tail_lsn.load());
-        return false;
-    }
-
-    // Validate if the lsn to which it is rolledback to is not truncated.
-    auto ret = m_records.status(to_lsn);
-    if (ret.is_out_of_range) {
-        HS_LOG_ASSERT(false, "Attempted to rollback to {} which is already truncated", to_lsn);
+    if (to_lsn > m_tail_lsn.load() || to_lsn < m_start_lsn.load()) {
+        HS_LOG_ASSERT(false, "Attempted to rollback to {} which is not in the range of [{}, {}]", to_lsn, m_start_lsn.load(), m_tail_lsn.load());
         return false;
     }
 
--- a/src/lib/logstore/log_store.cpp
+++ b/src/lib/logstore/log_store.cpp
@@ -283,13 +283,24 @@ void HomeLogStore::flush(logstore_seq_num_t upto_lsn) {
 }
 
 bool HomeLogStore::rollback(logstore_seq_num_t to_lsn) {
+    //Fast path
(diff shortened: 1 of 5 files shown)


src/tests/test_log_store_long_run.cpp
```diff
@@ -166,11 +166,6 @@ class SampleLogStoreClient {
     }
 
     void rollback_validate(uint32_t num_lsns_to_rollback) {
-
```
JacksonYao287, line 169: why this is removed?

src/lib/logstore/log_store.cpp
```diff
@@ -283,13 +283,24 @@ void HomeLogStore::flush(logstore_seq_num_t upto_lsn) {
...
+    //Fast path
+    if (to_lsn == m_tail_lsn.load()) {
+	return true;
+    }
+
+    if (to_lsn > m_tail_lsn.load()) {
+        HS_LOG_ASSERT(false, "Attempted to rollback to {} which is larger than tail_lsn {}", to_lsn, m_tail_lsn.load());
+        return false;
+    }
+
```
JacksonYao287, line 295: actually , what rollback does is to reset the tail_lsn to a valid lsn in the range [start_lsn , tail_lsn]. so if the rollback_to_lsn is not in the range, we can log an error and return false.
so my suggestion is add the following code is enough, and the check in line 297 can also be removed.
```
if(to_lsn < m_start_lsn || to_lsn > m_tail_lsn) {
        LOGERROR;
        return false;
}
```

JacksonYao287 (commented):
the basic logic is right. I left my comment , PATL.
also , please add some detail comments to the function decription to make the function more clear.
```
to_lsn is the tail_lsn after rollback
```

JacksonYao287 (approved):
LGTM

src/lib/logstore/log_store.cpp
```diff
@@ -283,13 +283,17 @@ void HomeLogStore::flush(logstore_seq_num_t upto_lsn) {
...
-        HS_LOG_ASSERT(false, "Attempted to rollback to {} which is already truncated", to_lsn);
+    //Fast path
+    if (to_lsn == m_tail_lsn.load()) {
+	return true;
+    }
+
+    if (to_lsn > m_tail_lsn.load() || to_lsn < m_start_lsn.load()) {
+        HS_LOG_ASSERT(false, "Attempted to rollback to {} which is not in the range of [{}, {}]", to_lsn, m_start_lsn.load(), m_tail_lsn.load());
         return false;
     }
```
yamingk, line 295: Can we also have an assert here for `HS_DEBUG_ASSERT(m_records.status(to_lsn + 1).is_out_of_range == false)`?

xiaoxichen: there are two angle, 

- from api point of view, why we want to assert  for a no-op rollback?  i. e  if last log is 100, rollback to 100 is a valid request
- from the implementation of streamtracker.   the is_out_of_range will not be set for LSN > upper_bound.  
 https://github.com/eBay/sisl/blob/ed773dbb0c7e53f46b29374e477f4be3dc1db99b/include/sisl/fds/stream_tracker.hpp#L128

PR #533, opened 2024-08-29
fix long duration of write_to_full test case and max_io test case

yamingk:
1. change max_io size to 60MB which will fit in with existing portion nblks with 4GB min chunk size;
2. change write_to_full case to allow use file for testing to complete sooner. The original purpose of this test case is to see what meta svc will behave after writing laste byte of its vdev. If with real drives, it will take around 3 days to complete this single test case. We can choose to remove --use_file if we do really want to test that long with meta svc some day...

Testing:
======
1. both max io test case and write to full test case passed by manual run on real drives. 
2. a new round of meta svc long run is kicked off and waiting to complete (will merge after a full round passed).

### DIFF
--- a/.github/workflows/build_dependencies.yml
+++ b/.github/workflows/build_dependencies.yml
@@ -155,6 +155,7 @@ jobs:
     - name: Export Recipes
       run: |
         sudo apt-get install -y python3-pyelftools libaio-dev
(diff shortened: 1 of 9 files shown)


xiaoxichen (dismissed):
lgtm

JacksonYao287 (approved):
LG

PR #535, opened 2024-08-30
add new dsn for destroy group 

JacksonYao287:
if we have a rreq {originator=1, term=1, dsn=0, lsn=7} in follower and a baseline resync is triggerd before the rreq is committed in the follower, then the on_commit of the rreq will not be called and as a result this rreq will become a garbage rreq in this follower. now if we trigger a destroy_group, a new rreq {originator=1, term=1, dsn=0} will created in the follower since the default dsn of a repl_key is 0.after the log of this rreq is appended to log store and get a new lsn, if we link the new lsn to the old rreq which has alread have a lsn, then a assert will be throw out. pls refer to repl_req_ctx::set_lsn

in this pr, we set the dsn to a new one , which is definitely unique in the follower, so that the new rreq will not have a conflict with the old rreq.

### DIFF
--- a/src/lib/replication/log_store/home_raft_log_store.cpp
+++ b/src/lib/replication/log_store/home_raft_log_store.cpp
@@ -262,10 +262,12 @@ raft_buf_ptr_t HomeRaftLogStore::pack(ulong index, int32_t cnt) {
     m_log_store->foreach (
         to_store_lsn(index),
         [this, &out_buf, &remain_cnt]([[maybe_unused]] store_lsn_t cur, const log_buffer& entry) mutable -> bool {
+            size_t const total_entry_size = entry.size() + sizeof(uint32_t);
             if (remain_cnt-- > 0) {
                 size_t avail_size = out_buf->size() - out_buf->pos();
-                if (avail_size < entry.size() + sizeof(uint32_t)) {
-                    avail_size += std::max(out_buf->size() * 2, (size_t)entry.size() + sizeof(uint32_t));
+                // available size of packing buffer should be able to hold entry.size() and the length of this entry
+                if (avail_size < total_entry_size) {
+                    avail_size += std::max(out_buf->size() * 2, total_entry_size);
                     out_buf = nuraft::buffer::expand(*out_buf, avail_size);
                 }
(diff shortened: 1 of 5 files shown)


src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -262,10 +262,12 @@ raft_buf_ptr_t HomeRaftLogStore::pack(ulong index, int32_t cnt) {
     m_log_store->foreach (
         to_store_lsn(index),
         [this, &out_buf, &remain_cnt]([[maybe_unused]] store_lsn_t cur, const log_buffer& entry) mutable -> bool {
+            size_t const total_entry_size = entry.size() + sizeof(uint32_t);
             if (remain_cnt-- > 0) {
                 size_t avail_size = out_buf->size() - out_buf->pos();
-                if (avail_size < entry.size() + sizeof(uint32_t)) {
-                    avail_size += std::max(out_buf->size() * 2, (size_t)entry.size() + sizeof(uint32_t));
+                // available size of packing buffer should be able to hold entry.size() and the length of this entry
```
JacksonYao287, line 268: @yamingk  I address your comments for the last PR here, PATL

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1041,6 +1053,14 @@ std::pair< bool, nuraft::cb_func::ReturnCode > RaftReplDev::handle_raft_event(nu
                 entries.size(), raft_req->get_last_log_term(), start_lsn, start_lsn + entries.size() - 1,
                 m_commit_upto_lsn.load(), raft_req->get_commit_idx());
 
+        if (start_lsn != m_data_journal->next_slot()) {
+            // if the start_lsn of this batch does not match the next_slot, drop this log batch
+            // this will happen when the leader is sending the logs which are already received and appended
+            RD_LOGD("start_lsn={} does not match the next_slot={}, dropping the log batch", start_lsn,
```
xiaoxichen, line 1059: Please add more comments here why we want to skip append on this case.

JacksonYao287: you are right , this part should be remove , nuraft will handle all the cases

xiaoxichen (approved):
lgtm

PR #541, opened 2024-09-09
return empty string upon failed client lookup

raakella1:


JacksonYao287 (approved):
LGTM

shosseinimotlagh (approved):

xiaoxichen (approved):
lgtm.

Where we hit this issue and how we handle lookup failure atm?

raakella1:
> lgtm.
> 
> Where we hit this issue and how we handle lookup failure atm?

This is an aid to the HomeObj PR https://github.com/eBay/HomeObject/pull/210. There can also be other cases when HO failed to return endpoints (CM does not recognize the SM id in lookup req, CM is unreachable etc.) So, nuraft keeps retrying create client until success

PR #543, opened 2024-09-10
fix raft repl dev ut

JacksonYao287:
before install snapshot completes, `raft_server()->get_committed_log_idx()` will be the same if `write_snapshot_data` is called several times, so that `snp_data->offset` can not be updated to new cursor. 
this PR use a separate last_data_committed_lsn to maintain the last committed lsn for writing data, which is use for updating  
`snp_data->offset` in `write_snapshot_data`

### DIFF
--- a/src/tests/test_raft_repl_dev.cpp
+++ b/src/tests/test_raft_repl_dev.cpp
@@ -141,6 +141,7 @@ public:
             std::unique_lock lk(db_mtx_);
             inmem_db_.insert_or_assign(k, v);
             lsn_index_.emplace(lsn, v);
+            last_data_committed_lsn = lsn;
             ++commit_count_;
         }
 
@@ -213,7 +214,7 @@ public:
         sisl::io_blob_safe blob{static_cast< uint32_t >(kv_snapshot_data_size)};
         std::memcpy(blob.bytes(), kv_snapshot_data.data(), kv_snapshot_data_size);
         snp_data->blob = std::move(blob);
-        snp_data->is_last_obj = true;
+        snp_data->is_last_obj = false;
         LOGINFOMOD(replication, "[Replica={}] Read logical snapshot callback obj_id={} term={} idx={} num_items={}",
                    g_helper->replica_num(), snp_data->offset, s->get_last_log_term(), s->get_last_log_idx(),
                    kv_snapshot_data.size());
@@ -233,10 +234,8 @@ public:
 
     void write_snapshot_data(shared< snapshot_context > context, shared< snapshot_data > snp_data) override {
         auto s = std::dynamic_pointer_cast< nuraft_snapshot_context >(context)->nuraft_snapshot();
-        auto last_committed_idx =
-            std::dynamic_pointer_cast< RaftReplDev >(repl_dev())->raft_server()->get_committed_log_idx();
         if (snp_data->offset == 0) {
-            snp_data->offset = last_committed_idx + 1;
+            snp_data->offset = last_data_committed_lsn + 1;
             LOGINFOMOD(replication, "[Replica={}] Save logical snapshot callback return obj_id={}",
                        g_helper->replica_num(), snp_data->offset);
             return;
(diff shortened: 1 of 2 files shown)


src/tests/test_raft_repl_dev.cpp
```diff
@@ -203,20 +204,20 @@ class TestReplicatedDB : public homestore::ReplDevListener {
...
+                       kv_snapshot_data.size());
+        } else {
+            LOGINFOMOD(replication, "no data need to be sent for snapshot");
         }
 
-        int64_t kv_snapshot_data_size = sizeof(KeyValuePair) * kv_snapshot_data.size();
-        sisl::io_blob_safe blob{static_cast< uint32_t >(kv_snapshot_data_size)};
-        std::memcpy(blob.bytes(), kv_snapshot_data.data(), kv_snapshot_data_size);
-        snp_data->blob = std::move(blob);
+        // we always read all the data in one shot, so we need to set is_last_obj to true
```
xiaoxichen, line 219: I dont get this part, we limit each batch to 1000 in L204(```if (kv_snapshot_data.size() >= 1000) { break; }``` )

src/tests/test_raft_repl_dev.cpp
```diff
@@ -269,7 +269,12 @@ class TestReplicatedDB : public homestore::ReplDevListener {
                    "[Replica={}] Save logical snapshot callback obj_id={} term={} idx={} is_last={} num_items={}",
                    g_helper->replica_num(), snp_data->offset, s->get_last_log_term(), s->get_last_log_idx(),
                    snp_data->is_last_obj, num_items);
-        snp_data->offset = last_committed_idx + 1;
+
```
xiaoxichen, line 272: we need to error out if `snp_data->is_last_obj == true` but `last_data_committed_lsn != s->get_last_log_idx()`, that means we have bug on sender side.

src/tests/test_raft_repl_dev.cpp
```diff
@@ -269,7 +269,12 @@ class TestReplicatedDB : public homestore::ReplDevListener {
...
+        // before we finish install snapshot, raft_server()->get_committed_log_idx() will always be the same. so we need
+        // last_data_committed_lsn to notify leader to transfer new data to follower.
+        // actually, leader will transfer all the data in one shot, so this is will not trigger any error, but we should
+        // make the logic correct.
+        snp_data->offset = last_data_committed_lsn + 1;
```
xiaoxichen, line 277: suggesting `snapshot_write_to_lsn` instead of `last_data_committed_lsn`

src/tests/test_raft_repl_dev.cpp
JacksonYao287: yes , you are right, I miss this

src/tests/test_raft_repl_dev.cpp
JacksonYao287: this is will be updated in on_commit

src/tests/test_raft_repl_dev.cpp
JacksonYao287: actually, last_data_committed_lsn should always be smaller than get_last_log_idx(), since at least we have config log entry, which will be added to log_idx but will not be add to  last_data_committed_lsn

xiaoxichen (dismissed):

src/tests/test_raft_repl_dev.cpp
```diff
@@ -269,7 +269,10 @@ class TestReplicatedDB : public homestore::ReplDevListener {
                    "[Replica={}] Save logical snapshot callback obj_id={} term={} idx={} is_last={} num_items={}",
```
sanebay, line 269: nit: Move log one line  after updating the snp_data->offset.

sanebay (commented):
Please run with -num_io=5000/10000

src/tests/test_raft_repl_dev.cpp
JacksonYao287: this log want to print out what we receive from leader , so I think we should keep it as this

JacksonYao287:
@sanebay I have run it with 5000 and it works

xiaoxichen (approved):

PR #546, opened 2024-09-11
Add support for raft repl dev replace member.

sanebay:
When replacing a member, add the new member, sync raft log for replace and finally remove the old member. Once we add new member, baseline or incremental resync will start. Remove the old member will cause nuraft mesg to exit the group and we periodically gc the destroyed group. Made the repl dev base test common so that both tests files can use. Tests by default create repl group with num_replica's. Dynamic tests create additional spare replica's which can be added to the test dynamically by calling replace member.

Testing
1. Bring replica-1, replica-2, replica-3. Replica-1 is the leader. Replace the replica-3 with a new spare replica-4.


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -107,13 +107,79 @@ bool RaftReplDev::join_group() {
...
+    if (m_my_repl_id == member_out_uuid && m_my_repl_id == get_leader_id()) {
+        // If leader is the member requested to move out, then give up leadership and return error.
+        // Client will retry replace_member request to the new leader.
+        raft_server()->yield_leadership(true /* immediate */, -1 /* successor */);
+        LOGINFO("Replace member leader is the member_out so yield leadership");
+        return make_async_error<>(ReplServiceError::NOT_LEADER);
+    }
+
+    // Step 2. Add the new member.
+    return m_msg_mgr.add_member(m_group_id, member_in_uuid)
```
sanebay, line 130: Add member, append long entry and remove the old member

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -836,6 +905,17 @@ void RaftReplDev::handle_error(repl_req_ptr_t const& rreq, ReplServiceError err)
...
 
+void RaftReplDev::replace_member(repl_req_ptr_t rreq) {
+    auto members = r_cast< const replace_members_ctx* >(rreq->header().cbytes());
+    replica_id_t member_in, member_out;
+    std::copy(members->out.begin(), members->out.end(), member_out.begin());
+    std::copy(members->in.begin(), members->in.end(), member_in.begin());
+    RD_LOGI("Raft repl replace_member member_out={} member_in={}", boost::uuids::to_string(member_out),
+            boost::uuids::to_string(member_in));
+
+    m_listener->replace_member(member_out, member_in);
```
sanebay, line 916: homeobject or listener can use this to update their pg metablks to remove old and add new member

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -93,9 +93,12 @@ void RaftReplService::start() {
                         .with_hb_interval(HS_DYNAMIC_CONFIG(consensus.heartbeat_period_ms))
                         .with_max_append_size(HS_DYNAMIC_CONFIG(consensus.max_append_batch_size))
                         .with_log_sync_batch_size(HS_DYNAMIC_CONFIG(consensus.log_sync_batch_size))
+    // TODO fix the log_gap thresholds when adding new member.
+#if 0
```
sanebay, line 97: New member stops syncing from leader due to these thresholds. This needs to be debugged and fixed in a later commit.

src/tests/test_common/hs_repl_test_common.hpp
```diff
@@ -134,11 +136,12 @@ class HSReplTestHelper : public HSTestHelper {
     HSReplTestHelper(std::string const& name, std::vector< std::string > const& args, char** argv) :
             name_{name}, args_{args}, argv_{argv} {}
 
-    void setup() {
+    void setup(uint32_t num_replicas) {
```
sanebay, line 139: setup takes how many replica's to start.

src/tests/test_common/raft_repl_test_base.hpp
```diff
@@ -0,0 +1,629 @@
+/*********************************************************************************
```
sanebay, line 1: Moved common code from test_raft_repl_dev.cpp to raft_test_base.hpp. No change in code. Just moved the classes.

src/tests/test_raft_repl_dev.cpp
```diff
@@ -12,603 +12,9 @@
...
-
-        int64_t kv_snapshot_data_size = sizeof(KeyValuePair) * kv_snapshot_data.size();
-        sisl::io_blob_safe blob{static_cast< uint32_t >(kv_snapshot_data_size)};
-        std::memcpy(blob.bytes(), kv_snapshot_data.data(), kv_snapshot_data_size);
-        snp_data->blob = std::move(blob);
-        snp_data->is_last_obj = true;
-        LOGINFOMOD(replication, "[Replica={}] Read logical snapshot callback obj_id={} term={} idx={} num_items={}",
-                   g_helper->replica_num(), snp_data->offset, s->get_last_log_term(), s->get_last_log_idx(),
-                   kv_snapshot_data.size());
+#include "test_common/raft_repl_test_base.hpp"
```
sanebay, line 15: Moved these code to raft_repl_test_base.hpp header file.

src/tests/test_raft_repl_dev_dynamic.cpp
```diff
@@ -0,0 +1,132 @@
+/*********************************************************************************
```
sanebay, line 1: This test creates repl dev group with 3 replica's by default and then add spare replica's.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -107,13 +107,79 @@ bool RaftReplDev::join_group() {
...
+            if (e.hasError() && e.error() != nuraft::cmd_result_code::CANCELLED) {
+                RD_LOGE("Replace member error in add member : {}", e.error());
+                return make_async_error<>(RaftReplService::to_repl_error(e.error()));
+            }
+            auto member_out = boost::uuids::to_string(member_out_uuid);
+            auto member_in = boost::uuids::to_string(member_in_uuid);
+
+            LOGINFO("Replace member added member={} to group_id={}", member_in, group_id_str());
+
+            // Step 3. Append log entry to mark the old member is out and new member is added.
```
JacksonYao287, line 146: if leader changes after line140 and before line 147, then who can send a log entry to mark the old member is out and remove member_out of out this group?

should we separate replace_member to two calls: add_member and remove_member?

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: will revisit here if we need a separate add_member and remove_member

JacksonYao287 (commented):
@sanebay  thanks for the patch!

we need take leader change into account, pls see my comments and correct me if I am wrong

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -107,13 +107,79 @@ bool RaftReplDev::join_group() {
...
 }
 
+AsyncReplResult<> RaftReplDev::replace_member(replica_id_t member_out_uuid, replica_id_t member_in_uuid) {
+    LOGINFO("Replace member group_id={} member_out={} member_in={}", group_id_str(),
+            boost::uuids::to_string(member_out_uuid), boost::uuids::to_string(member_in_uuid));
+
+    // Step 1: Check if leader itself is requested to move out.
+    if (m_my_repl_id == member_out_uuid && m_my_repl_id == get_leader_id()) {
+        // If leader is the member requested to move out, then give up leadership and return error.
+        // Client will retry replace_member request to the new leader.
```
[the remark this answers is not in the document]
sanebay: Right, better to add it.

src/lib/replication/repl_dev/raft_repl_dev.cpp
sanebay: Yeah valid scenario. In that case log entry append will fail at line 157, CM will retry again with the latest leader. Thats the flow we have decided with CM. It should be safe to send the same message to a different leader.

sanebay:
we may need add_member and remove_member, if we do expansion/shrink. Our use case is only replace and we want to update in the homeobject PG metablk in single write atomically. The member out will have its PG metablk with destroyed flag enabled. GC will use this delete all blobs for that PG.

JacksonYao287:
> The member out will have its PG metablk with destroyed flag enabled. GC will use this delete all blobs for that PG.

do we need gc for the member out? I think if we remove it from the group , we can just re-fromat it with homestore, then it will be  clean , no?

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: >It should be safe to send the same message to a different leader.

if leader changes after `propose_to_raft` but before `m_msg_mgr.rem_member`, then it will return `ReplServiceError::RETRY_REQUEST` .

now , if cm resends the request to the new leader, I think `add_member` will succeed in new leader since the member_in is already added in the group by the previous leader. now, next step is `propose_to_raft`, which will create another `HS_CTRL_REPLACE` log and commit it.

so `HS_CTRL_REPLACE` will be committed twice at the other follower, one is from  previous leader , one is from current leader.

we should make it safe to commit `HS_CTRL_REPLACE` twice（or make committing `HS_CTRL_REPLACE` idempotent） for the same member_out and member_in. so we need to add some checks in `m_listener#replace_member(member_out, member_in);`

so, I think basically it will be safe,  but we should add some code to handle the corner case.

xiaoxichen: more for cm:
if the new leader is the `member_in`,   we havent respond to the call yet.   We get the leader (in nuraft_mesg) from raft which is always up_to_date,  and respond to CM.

Wondering if CM expect HB from a to-be-added SM , or expect retry aganist a `to-be-added SM`.

sanebay: idempotency of HS_CTRL_REPLACE is same as create pg or shard. Gateway can send request leader, leader does raft and before returning success to gateway crashed and restarted, gateway will send to same leader and we might create pg. In HO in add_replace function , in and out members are already applied, its a nop.

sanebay:
> > The member out will have its PG metablk with destroyed flag enabled. GC will use this delete all blobs for that PG.
> 
> do we need gc for the member out? I think if we remove it from the group , we can just re-fromat it with homestore, then it will be clean , no?

One SM could have multiple PG's on same disk. We could move out some PG for space issues or load balance.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -107,13 +107,79 @@ bool RaftReplDev::join_group() {
...
+        LOGINFO("Replace member leader is the member_out so yield leadership");
+        return make_async_error<>(ReplServiceError::NOT_LEADER);
+    }
+
+    // Step 2. Add the new member.
+    return m_msg_mgr.add_member(m_group_id, member_in_uuid)
+        .via(&folly::InlineExecutor::instance())
+        .thenValue([this, member_in_uuid, member_out_uuid](auto&& e) -> AsyncReplResult<> {
+            // TODO Currently we ignore the cancelled, fix nuraft_mesg to not timeout
+            // when adding member. Member is added to cluster config until member syncs fully
```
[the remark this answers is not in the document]
sanebay: This is a nuraft behaviour. I had same doubt. This what he shared.
"""
* there are three servers: s1, s2, s3. So quorum size is 2. 
* let's say s3 is lagging behind, but still s1 can make a commit, by forming quorum {s1, s2}.
* now new member s4 joins the cluster. 
* if s4 joins the cluster immediately without catch-up, it is also lagging behind.
* since it joins the cluster immediately, now quorum size becomes 3. 
* now s1 cannot make a commit, no matter which member is in the quorum: {s1, s2, s3}, {s1, s2, s4}.
"""

src/tests/test_raft_repl_dev_dynamic.cpp
```diff
@@ -0,0 +1,132 @@
...
+    uint32_t member_out = num_replicas - 1;
+    uint32_t member_in = num_replicas;
+
+    g_helper->sync_for_test_start(num_members);
+    if (g_helper->replica_num() < num_replicas) {
+        // With existing raft repl dev group, write IO's, validate and call replace_member on leader.
+        LOGINFO("Writing on leader num_io={} replica={}", num_io_entries, g_helper->replica_num());
+        this->write_on_leader(num_io_entries, true /* wait_for_commit */);
+
+        replace_member(db, g_helper->replica_id(member_out), g_helper->replica_id(member_in));
```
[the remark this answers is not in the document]
sanebay: We shouldnt allow replace_member on leader. We can add a check in HomeObjectImpl::replace_member. We should fix this condition too in HO. https://github.com/eBay/HomeObject/blob/main/src/lib/pg_manager.cpp#L31

JacksonYao287:
the code generally looks good to me.  pls resolve the conflict

src/lib/replication/repl_dev/raft_repl_dev.cpp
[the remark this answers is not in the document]
JacksonYao287: this scenario makes sense.
another question , do we have a plan to have a separate add_member and rem_member?

for example, if we have a very heavy read workload for one pg,  we want to add some new members to take over some read workload. so we only need add_memeber.

src/tests/test_raft_repl_dev_dynamic.cpp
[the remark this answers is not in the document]
xiaoxichen: no I mean, replace_member call(targeting a follower) was wrongly send to follower instead of leader.
We should error out in this cae.

src/lib/replication/repl_dev/raft_repl_dev.cpp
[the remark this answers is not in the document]
xiaoxichen: @JacksonYao287 we dont want that as of now. We dont have the flexibility to control replication size of each PG in control plane.  Lets revisit if we have solid use case.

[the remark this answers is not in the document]
xiaoxichen: @sanebay that makes sense from the raft point of view, but make it very tricky for cm->sm protocal.  Once leader SM reboot, what would be the state of the in-flight add_srv? will it continue till success or being discard?  who should take the responsibility to respond to CM? 

Is there a possibility that we can add new_member as learner and promoted to participant once it catches up?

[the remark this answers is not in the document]
xiaoxichen: also for a failed/forgot add_srv, who should take the responsibility to reclaim space?   If CM retries with same PG to same SM I believe baseline resync will erase and re-do.  If CM attempt with other PG to this SM I am not sure how we handle this --- linking to the fixed pg size design @JacksonYao287

src/lib/replication/service/raft_repl_service.cpp
sanebay: If we dont truncate, raft only does incremental resync. This is the log of the new member added to the group.

```
[09/19/24 12:05:40-07:00] [I] [test_raft_repl_dev_dynamic_replica_3] [297795] [raft_repl_dev.cpp:79] [rd=rdev1] Started New RaftReplDev group_id=3b25627e-f07b-41e8-b992-4696be833582, replica_id=00030123-4567-89ab-cdef-01234567
89ab, raft_server_id=558356 commited_lsn=0, compact_lsn=0, checkpoint_lsn:0, next_dsn=0 log_dev=1 log_store=0
....
[09/19/24 12:05:40-07:00] [D] [test_raft_repl_dev_dynamic_replica_3] [297760] [raft_repl_dev.cpp:1133] [rd=rdev1] Raft channel: Received 64 append entries on follower from leader, term 0, lsn 1 ~ 64 , my commited lsn 0 , leade
r commmited lsn 1004
```

src/lib/replication/repl_dev/raft_repl_dev.cpp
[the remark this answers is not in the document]
sanebay: > @sanebay that makes sense from the raft point of view, but make it very tricky for cm->sm protocal. Once leader SM reboot, what would be the state of the in-flight add_srv? will it continue till success or being discard? who should take the responsibility to respond to CM?
> 
> Is there a possibility that we can add new_member as learner and promoted to participant once it catches up?

add_srv internally does cluster config change with log entry. So it needs commit . Its treated as same append entry log. If committed and crashed, CM can send again and we return error that server already exists.  If it didnt commit, it will rollback once another leader selects. I havent explored the learner option.


> also for a failed/forgot add_srv, who should take the responsibility to reclaim space? 
No as mentioned, add_srv should fail with error SERVER_ALREADY_EXISTS . As per design, if new member failed after sometime, manual intervnetion is needed and a new PG move is needed(https://github.com/eBay/NuRaft/blob/master/src/handle_join_leave.cxx#L68)

src/tests/test_raft_repl_dev_dynamic.cpp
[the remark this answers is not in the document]
sanebay: just like create_pg, create shard, replace_member should check for if its leader in HO. We can add additional check in HS too.

src/lib/replication/repl_dev/raft_repl_dev.cpp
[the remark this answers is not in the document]
JacksonYao287: >also for a failed/forgot add_srv, who should take the responsibility to reclaim space? If CM retries with same PG to same SM I believe baseline resync will erase and re-do. If CM attempt with other PG to this SM I am not sure how we handle this

yes, a failed add_srv will leave some stale data in the disk of member_in , it is caused by the snapshot before the config change is committed.

I have gone through the raft code. I will add the solution in the doc

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -93,7 +93,10 @@ void RaftReplService::start() {
                         .with_hb_interval(HS_DYNAMIC_CONFIG(consensus.heartbeat_period_ms))
                         .with_max_append_size(HS_DYNAMIC_CONFIG(consensus.max_append_batch_size))
                         .with_log_sync_batch_size(HS_DYNAMIC_CONFIG(consensus.log_sync_batch_size))
+    // TODO fix the log_gap thresholds when adding new member.
```
sanebay, line 96: Shared logs with nuraft team, but we couldnt find the root cause immediately. Will have to debug with nuraft team when they are available. Will create a ticket to keep track of this. https://github.com/eBay/HomeStore/issues/561

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -107,13 +107,89 @@ bool RaftReplDev::join_group() {
...
+    m_msg_mgr.get_srv_config_all(m_group_id, configs_out);
+    if (std::find_if(configs_out.begin(), configs_out.end(), [member_out_uuid](const auto& cfg) {
+            return boost::uuids::to_string(member_out_uuid) == cfg->get_endpoint();
+        }) == configs_out.end()) {
+        RD_LOGE("Member out not found in the group.");
+        return make_async_error<>(ReplServiceError::SERVER_NOT_FOUND);
+    }
+
+    // Step 1: Check if leader itself is requested to move out.
+    if (m_my_repl_id == member_out_uuid && m_my_repl_id == get_leader_id()) {
```
yamingk, line 131: One issue I see is: when SM experienced disk down, if it happens on leader, HomeStore should yield leadership immediately, the reason is, if we rely on CM to do that, there needs to be some manual ops to kicks in or even if it is automated from CM layer, it also takes some time for CM to detect that. If Homestore can yield leadership it to other member, the other two copies can serve traffic immeidately without any SOPs need to be taken in production.

sanebay: Created an issue for this https://github.com/eBay/HomeStore/issues/563

src/lib/replication/repl_dev/raft_repl_dev.h
```diff
@@ -35,6 +35,11 @@ using raft_cluster_config_ptr_t = nuraft::ptr< nuraft::cluster_config >;
 
 ENUM(repl_dev_stage_t, uint8_t, INIT, ACTIVE, DESTROYING, DESTROYED, PERMANENT_DESTROYED);
 
+struct replace_members_ctx {
+    std::array< uint8_t, 16 > out;
```
yamingk, line 39: can we put comment for out/in, that it is replica_id_t, or can we just name them out_replica_id, in_replica_id

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -107,13 +107,89 @@ bool RaftReplDev::join_group() {
...
+            }
+            auto member_out = boost::uuids::to_string(member_out_uuid);
+            auto member_in = boost::uuids::to_string(member_in_uuid);
+
+            RD_LOGI("Replace member added member={} to group_id={}", member_in, group_id_str());
+
+            // Step 3. Append log entry to mark the old member is out and new member is added.
+            auto rreq = repl_req_ptr_t(new repl_req_ctx{});
+            replace_members_ctx members;
+            std::copy(member_in_uuid.begin(), member_in_uuid.end(), members.in.begin());
```
yamingk, line 159: nit: we can't do `member.in = std::move(members_in_uuid)`, can we?

src/lib/replication/service/raft_repl_service.cpp
yamingk: Can we put the comment why it is disabled, also we need to create an issue to track this, put in the description what is now happening after disable this, and what was happening without this.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -107,13 +107,89 @@ bool RaftReplDev::join_group() {
...
     return true;
 }
 
+AsyncReplResult<> RaftReplDev::replace_member(replica_id_t member_out_uuid, replica_id_t member_in_uuid) {
+    LOGINFO("Replace member group_id={} member_out={} member_in={}", group_id_str(),
+            boost::uuids::to_string(member_out_uuid), boost::uuids::to_string(member_in_uuid));
+
+    // Check if member_out exists in the group.
+    std::vector< nuraft::ptr< nuraft::srv_config > > configs_out;
+    m_msg_mgr.get_srv_config_all(m_group_id, configs_out);
```
yamingk, line 122: if the out is not found, shouldn't we just move ahead to add the in_member. The reason is we have retry logic, and we don't need to worry about which operation actually did moved out the member with mix of crash in between and and if one operation did manage to move the old member, shouldn't we just go ahead ignore the out_member not seen in current group and add in_member? 

Also we should check in_member is not already there in the raft group. Since we have the retry logic, and this retry needs to be idomponent.

src/lib/replication/repl_dev/raft_repl_dev.h
sanebay: Ack

src/lib/replication/service/raft_repl_service.cpp
sanebay: Ack. Have an issue created https://github.com/eBay/HomeStore/issues/561

src/lib/replication/repl_dev/raft_repl_dev.cpp
sanebay: We need a char array as we are serializing two members. members_in_uuid is a string, so we couldnt use it.

src/lib/replication/repl_dev/raft_repl_dev.cpp
sanebay: Yeah we can do that. Instead of doing here, we can depend on raft server itself returns nuraft::cmd_result_code::SERVER_ALREADY_EXISTS for add and nuraft::cmd_result_code::SERVER_NOT_FOUND for remove and ignore those errors.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -107,13 +107,89 @@ bool RaftReplDev::join_group() {
...
+        .via(&folly::InlineExecutor::instance())
+        .thenValue([this, member_in_uuid, member_out_uuid](auto&& e) -> AsyncReplResult<> {
+            // TODO Currently we ignore the cancelled, fix nuraft_mesg to not timeout
+            // when adding member. Member is added to cluster config until member syncs fully
+            // with atleast stop gap. This will take a lot of time for block or
+            // object storage.
+            if (e.hasError()) {
+                if (e.error() == nuraft::cmd_result_code::CANCELLED ||
+                    e.error() == nuraft::cmd_result_code::SERVER_ALREADY_EXISTS) {
+                    RD_LOGW("Ignoring error returned from nuraft add_member {}", e.error());
```
yamingk, line 140: Please put a comment here why we can safely ignoring the above error, e.g. because of retry rely on idempotent.
We should have flip to simulate retry path also, it would be difficult to verify this path in end-2-end env.

yamingk (approved):

src/lib/replication/service/raft_repl_service.cpp
sanebay: No. Looks like there is some error in nuraft/nuraft_mesg after sending the first pack. Created a issue to debug this with nuraft. Logs are attached to that issue.

PR #548, opened 2024-09-12
Fix crash simulation

shosseinimotlagh:
The problem was in case of crash, either index service or wb cache can be nullptr during crash simulation while accessing in different code path. This will prevent accessing them while triggering crash simulator


JacksonYao287 (approved):
LGTM

koujl (approved):
lgtm

PR #554, opened 2024-09-18
FIX oom and blk alloc overflow which cause OOM

shosseinimotlagh:


yamingk (dismissed):

src/engine/blkstore/blkstore.hpp
```diff
@@ -336,20 +336,19 @@ class BlkStore {
...
-                const auto ret{m_vdev.alloc_blk(nblks_op, hints, result_blkid)};
-                if (ret != BlkAllocStatus::SUCCESS) { return ret; }
-                out_blkid.insert(std::end(out_blkid), std::make_move_iterator(std::begin(result_blkid)),
-                                 std::make_move_iterator(std::end(result_blkid)));
-                nblks -= nblks_op;
-            }
+        // Using a 16-bit value (blk_count_t) for page counts can lead to overflow issues when managing large storage
+        // devices. With a page size of 4096 bytes, a 16-bit value limits the addressable storage to 256 MB (65,536
+        // pages * 4096 bytes/page).
+        uint32_t nblks{static_cast< uint32_t >(size / m_pagesz)};
```
yamingk, line 342: Sorry, I had a second thought on this, I think now the reason for line 340~342 is a shortcut for performance consideration, because 99.99% of the blk_allocation will fall into if, not else. cc @szmyd , can we change it back? sorry.

shosseinimotlagh: done

src/engine/blkstore/blkstore.hpp
```diff
@@ -336,14 +336,17 @@ class BlkStore {
     BlkAllocStatus alloc_blk(const uint32_t size, blk_alloc_hints& hints, std::vector< BlkId >& out_blkid) {
         // Allocate a block from the device manager
         assert(size % m_pagesz == 0);
-        blk_count_t nblks{static_cast< blk_count_t >(size / m_pagesz)};
+        // Using a 16-bit value (blk_count_t) for page counts can lead to overflow issues when managing large storage
+        // devices. With a page size of 4096 bytes, a 16-bit value limits the addressable storage to 256 MB (65,536
+        // pages * 4096 bytes/page).
+        uint32_t nblks{static_cast< uint32_t >(size / m_pagesz)};
         if (nblks <= BlkId::max_blks_in_op()) {
             return (m_vdev.alloc_blk(nblks, hints, out_blkid));
```
yamingk, line 344: line 344 will convert uint32_t to uint16_t because m_vdev.alloc_blk takes 16bit as first parameter, which is fine here, becase we only are in if branch when nblks <= BlkId::max_blks_in_op(), but it is better to put a comment and a debug assert here by saying BlkId::max_blks_in_op() must be <= `std::numeric_limits<decltype<blk_count_t>>::max()`. 

Also please put a static_cast before nblks, when passing to alloc_blk, for the ease of read.

src/engine/blkstore/blkstore.hpp
szmyd: This looks okay to me; in the future maybe we should move things like this into a subroutine that can more easily be tested with a UT that does not need to go through the alloc_blk call? Just a thought when working in 4.x for situations such as this.

szmyd (approved):

yamingk (approved):

src/engine/blkstore/blkstore.hpp
```diff
@@ -325,26 +325,31 @@ class BlkStore {
     BlkAllocStatus alloc_contiguous_blk(const uint32_t size, blk_alloc_hints& hints, BlkId* const out_blkid) {
         // Allocate a block from the device manager
         assert(size % m_pagesz == 0);
-        const blk_count_t nblks{static_cast< blk_count_t >(size / m_pagesz)};
+        const uint32_t nblks{static_cast< uint32_t >(size / m_pagesz)};
```
szmyd, line 328: This doesn't make sense to me. We change this from a 16bit value to a 32bit value (to avoid overflow?) and then later immediately cast it back to a 16bit parameter in the call to `alloc_contig...`; what's the point?

szmyd (dismissed):
looks ok to me; might want to wait for @yamingk to also approve.

src/engine/blkstore/blkstore.hpp
```diff
@@ -336,15 +339,20 @@ class BlkStore {
...
         assert(size % m_pagesz == 0);
-        blk_count_t nblks{static_cast< blk_count_t >(size / m_pagesz)};
+        // Using a 16-bit value (blk_count_t) for page counts can lead to overflow issues when managing large storage
+        // devices. With a page size of 4096 bytes, a 16-bit value limits the addressable storage to 256 MB (65,536
+        // pages * 4096 bytes/page).
+        uint32_t nblks{static_cast< uint32_t >(size / m_pagesz)};
         if (nblks <= BlkId::max_blks_in_op()) {
-            return (m_vdev.alloc_blk(nblks, hints, out_blkid));
+            auto max_val = std::numeric_limits< blk_count_t >::max();
+            HS_DBG_ASSERT_LE(nblks, max_val, "nblks must be less than {}", max_val);
```
yamingk, line 348: nit:  do we still this assert since you've added static assert at blk.h? I mean the static assert at blk.h can make sure this assert always true, right? Also can we just use one line: `HS_DBG_ASSERT_LE(nblks, std::numeric_limits< blk_count_t >::max(), "nblks must be less than {}", max_val);`, although compiler might do the same for optimization.  I am fine either way.

src/engine/blkstore/blkstore.hpp
```diff
@@ -325,6 +325,9 @@ class BlkStore {
     BlkAllocStatus alloc_contiguous_blk(const uint32_t size, blk_alloc_hints& hints, BlkId* const out_blkid) {
         // Allocate a block from the device manager
         assert(size % m_pagesz == 0);
+        static uint32_t max_suppoted_size = m_pagesz * std::numeric_limits< blk_count_t >::max();
```
yamingk, line 328: is there any reason it needs to be `static`? it can be `uint32_t const max_supported_size =...`, right?

yamingk, line 328: nit:  ideally this `max_supported_size` should come from 1Chunk::BlkAlloc::max_supported_size()`. We can do it from 4.x

src/engine/blkstore/blkstore.hpp
shosseinimotlagh: I remove it

szmyd (approved):

PR #556, opened 2024-09-19
Add index CR UT for basic merge

koujl:
https://docs.google.com/presentation/d/1NUvu1ZlDKqJWfBayg2lSWHSB6TEaZ4Y3VbnGc996drM/edit#slide=id.g2f8c68a8252_1_10

------

Add basic index CR logic & UT.

Major changes including:
* Reconstruct our handling strategy in `wb_cache::recover()` for recovered new/freed buffers
* Add protection for several corner cases in `index_table::repair_link()`

As well as some bug fixes:
* Fix several down_buffer counter leaks
* Fix a bug in wb_cache async_flush that may cause one buffer get processed multiple times
* Fix some bugs in racing conditions during homestore restart

------

There are still some known issues in some corner cases, such as incorrect next_node pointer. However, since solving these may introduce complicated modification to the current logic, let's leave them to future PRs and discuss later.

### DIFF
--- a/src/include/homestore/btree/detail/btree_node.hpp
+++ b/src/include/homestore/btree/detail/btree_node.hpp
@@ -37,6 +37,7 @@ struct transient_hdr_t {
     /* these variables are accessed without taking lock and are not expected to change after init */
     uint8_t leaf_node{0};
     uint64_t max_keys_in_node{0};
+    uint64_t min_keys_in_node{0}; // to specify the threshold for triggering merge
 
     bool is_leaf() const { return (leaf_node != 0); }
 };
@@ -116,6 +117,7 @@ public:
         m_trans_hdr.leaf_node = is_leaf;
 #ifdef _PRERELEASE
         m_trans_hdr.max_keys_in_node = cfg.m_max_keys_in_node;
+        m_trans_hdr.min_keys_in_node = cfg.m_min_keys_in_node;
 #endif
 
     }
@@ -299,6 +301,7 @@ public:
 
     template < typename K >
     K get_first_key() const {
+        if (total_entries() == 0) { return K{}; }
         return get_nth_key< K >(0, true);
     }
 
@@ -333,6 +336,7 @@ public:
 
     // uint32_t total_entries() const { return (has_valid_edge() ? total_entries() + 1 : total_entries()); }
     uint64_t max_keys_in_node() const { return m_trans_hdr.max_keys_in_node; }
+    uint64_t min_keys_in_node() const { return m_trans_hdr.min_keys_in_node; }
 
     void lock(locktype_t l) const {
         if (l == locktype_t::READ) {
@@ -392,6 +396,12 @@ public:
             }
             fmt::format_to(std::back_inserter(str), "]");
         }
+
+        // Should not happen
+        if (this->is_node_deleted()) {
+            fmt::format_to(std::back_inserter(str), " **DELETED** ");
+        }
+
         return str;
     }
 
@@ -527,15 +537,10 @@ public:
 
     virtual uint32_t occupied_size() const { return (node_data_size() - available_size()); }
     bool is_merge_needed(const BtreeConfig& cfg) const {
-#if 0
 #ifdef _PRERELEASE
-       if (iomgr_flip::instance()->test_flip("btree_merge_node") && occupied_size() < node_data_size) {
-           return true;
-       }
-
-       auto ret = iomgr_flip::instance()->get_test_flip< uint64_t >("btree_merge_node_pct");
-       if (ret && occupied_size() < (ret.get() * node_data_size() / 100)) { return true; }
-#endif
+        if (min_keys_in_node()) {
+            return total_entries() < min_keys_in_node();
+        }
 #endif
         return (occupied_size() < cfg.suggested_min_size());
     }
--- a/src/include/homestore/index/index_table.hpp
+++ b/src/include/homestore/index/index_table.hpp
@@ -78,7 +78,8 @@ public:
     }
 
     void destroy() override {
-        Btree< K, V >::destroy_btree(nullptr);
+        auto cpg = cp_mgr().cp_guard();
+        Btree<K, V>::destroy_btree(cpg.context(cp_consumer_t::INDEX_SVC));
         m_sb.destroy();
     }
 
@@ -130,13 +131,16 @@ public:
         idx_buf->m_dirtied_cp_id = cpg->id();
         BtreeNodePtr bn = BtreeNodePtr{n};
 
-        LOGTRACEMOD(wbcache, "repair_node cp={} buf={}", cpg->id(), idx_buf->to_string());
-        repair_links(bn, (void*)cpg.context(cp_consumer_t::INDEX_SVC));
+        // Only for interior nodes we need to repair its links
+        if (!bn->is_leaf()) {
+            LOGTRACEMOD(wbcache, "repair_node cp={} buf={}", cpg->id(), idx_buf->to_string());
+            repair_links(bn, (void *) cpg.context(cp_consumer_t::INDEX_SVC));
+        }
 
         if (idx_buf->m_up_buffer && idx_buf->m_up_buffer->is_meta_buf()) {
             // Our up buffer is a meta buffer, which means that we are the new root node, we need to update the
             // meta_buf with new root as well
-            on_root_changed(bn, (void*)cpg.context(cp_consumer_t::INDEX_SVC));
+            on_root_changed(bn, (void *) cpg.context(cp_consumer_t::INDEX_SVC));
         }
     }
 
@@ -223,7 +227,8 @@ protected:
         wb_cache().free_buf(n->m_idx_buf, r_cast< CPContext* >(context));
     }
 
-    btree_status_t on_root_changed(BtreeNodePtr const& new_root, void* context) override {
+    btree_status_t
+    on_root_changed(BtreeNodePtr const &new_root, void *context) override {
         // todo: if(m_sb->root_node == new_root->node_id() && m_sb->root_link_version == new_root->link_version()){
         // return btree_status_t::success;}
         m_sb->root_node = new_root->node_id();
@@ -235,12 +240,12 @@ protected:
         }
 
         auto& root_buf = static_cast< IndexBtreeNode* >(new_root.get())->m_idx_buf;
-        wb_cache().transact_bufs(ordinal(), m_sb_buffer, root_buf, {}, {}, r_cast< CPContext* >(context));
+        wb_cache().transact_bufs(ordinal(), m_sb_buffer, root_buf, {}, {}, r_cast<CPContext *>(context));
         return btree_status_t::success;
     }
 
     btree_status_t repair_links(BtreeNodePtr const& parent_node, void* cp_ctx) {
-        BT_LOG(DEBUG, "Repairing links for parent node {}", parent_node->to_string());
+        BT_LOG(DEBUG, "Repairing links for parent node [{}]", parent_node->to_string());
         // TODO: is it possible that repairing many nodes causes an increase to level of btree? If so, then this needs
         // to be handled. Get the last key in the node
         auto const last_parent_key = parent_node->get_last_key< K >();
@@ -250,7 +255,15 @@ protected:
                           parent_node->node_id());
             return btree_status_t::not_found;
         }
-        BT_LOG(INFO, "Repairing node={} with last_parent_key={}", parent_node->to_string(),
+
+        // Get all original child ids as a support to check if we are beyond the last child node
+        std::set<bnodeid_t> orig_child_ids;
+        for (uint32_t i = 0; i < parent_node->total_entries(); ++i) {
+            BtreeLinkInfo link_info;
+            parent_node->get_nth_value(i, &link_info, true);
+            orig_child_ids.insert(link_info.bnode_id());
+        }
+        BT_LOG(INFO, "Repairing node=[{}] with last_parent_key={}", parent_node->to_string(),
                last_parent_key.to_string());
 
         // Get the first child node and its link info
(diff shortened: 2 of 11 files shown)


szmyd (commented):
Are we intending to do more work on this PR? It does not meet the 80% coverage requirement and will lower the current overall coverage once merged which we committed to avoiding.

src/lib/index/wb_cache.cpp
```diff
@@ -418,52 +433,79 @@ void IndexWBCache::recover(sisl::byte_view sb) {
...
+    std::vector<IndexBufferPtr> pending_bufs;
+    std::vector<IndexBufferPtr> deleted_bufs;
     for (auto const& [_, buf] : bufs) {
-        if (buf->m_node_freed || (buf->m_created_cp_id == icp_ctx->id())) {
+        if (buf->m_node_freed) {
+            // Freed node
+            if (buf->m_bytes == nullptr) {
+                buf->m_bytes = hs_utils::iobuf_alloc(m_node_size, sisl::buftag::btree_node, m_vdev->align_size());
+                m_vdev->sync_read(r_cast<char *>(buf->m_bytes), m_node_size, buf->blkid());
+            }
```
shosseinimotlagh, line 445: if you read the buf->m_bytes , them the is valid and modified cp will not be set in was_commited and you need to check it later (which is not NIT). Also you need to read sync_read twice which is unnecessary and cause performance degradation

src/lib/index/wb_cache.cpp
```diff
@@ -286,10 +283,9 @@ void IndexWBCache::transact_bufs(uint32_t index_ordinal, IndexBufferPtr const& p
     } else {
         icp_ctx->add_to_txn_journal(index_ordinal,          // Ordinal
                                     child_buf->m_up_buffer, // real up buffer
-                                    new_node_bufs.empty() ? freed_node_bufs[0]->m_up_buffer
-                                                          : new_node_bufs[0]->m_up_buffer, // real in place child
-                                    new_node_bufs,                                         // new node bufs
-                                    freed_node_bufs                                        // free_node_bufs
+                                    child_buf, // real in place child
```
shosseinimotlagh, line 286: please write comments here

src/include/homestore/btree/detail/btree_mutate_impl.ipp
```diff
@@ -236,9 +236,9 @@ btree_status_t Btree< K, V >::check_split_root(ReqT& req) {
 
     ret = split_node(root, child_node, root->total_entries(), &split_key, req.m_op_context);
     if (ret != btree_status_t::success) {
+        on_root_changed(child_node, root, req.m_op_context); // Revert it back
```
shosseinimotlagh, line 239: please explain why do we need to have child_node for revert the root id in superblk?

src/include/homestore/index/index_table.hpp
```diff
@@ -247,7 +258,15 @@ class IndexTable : public IndexTableBase, public Btree< K, V > {
                           parent_node->node_id());
             return btree_status_t::not_found;
         }
-        BT_LOG(INFO, "Repairing node={} with last_parent_key={}", parent_node->to_string(),
+
+        // Get all current child ids
```
shosseinimotlagh, line 262: please explain the algorithm here

src/lib/common/crash_simulator.hpp
```diff
@@ -13,9 +13,8 @@ class CrashSimulator {
     ~CrashSimulator() = default;
 
     void crash() {
+        m_crashed.update([](auto *s) { *s = true; });
```
shosseinimotlagh, line 16: in case of no cb, sigkill needs to be called. this change can prevent it.

src/include/homestore/index/index_table.hpp
koujl: Added more comments.

The original child IDs serve as a supportive condition for deciding when to exit the loop. If a lower-level merge is committed while the parent-level merge is not, we may see children with keys beyond the original last key of the parent. In such case, we need to check if it's in the original children list to determine its parent.

src/include/homestore/btree/detail/btree_mutate_impl.ipp
koujl: The child_node here is the original root node - and we're reverting what we're doing at L230.

koujl:
@szmyd Is there any coverage report I can refer to for this specific commit? I can only find this but seems its comparison base is still not the latest:
https://app.codecov.io/gh/eBay/HomeStore/commit/0c2481714651b1ed5881fa37a9b8c979e0af7178

src/include/homestore/btree/detail/btree_node.hpp
```diff
@@ -527,6 +537,9 @@ class BtreeNode : public sisl::ObjLifeCounter< BtreeNode > {
 
     virtual uint32_t occupied_size() const { return (node_data_size() - available_size()); }
     bool is_merge_needed(const BtreeConfig& cfg) const {
+        if (min_keys_in_node()) {
```
shosseinimotlagh, line 540: anyway, put it in _PRERELEASE

src/lib/common/crash_simulator.hpp
shosseinimotlagh: crash without cb can dominate the one with cb

src/lib/index/wb_cache.cpp
```diff
@@ -289,13 +293,7 @@ void IndexWBCache::transact_bufs(uint32_t index_ordinal, IndexBufferPtr const& p
...
         icp_ctx->add_to_txn_journal(index_ordinal, parent_buf, nullptr, {child_buf}, {});
     } else {
-        icp_ctx->add_to_txn_journal(index_ordinal,          // Ordinal
-                                    child_buf->m_up_buffer, // real up buffer
-                                    new_node_bufs.empty() ? freed_node_bufs[0]->m_up_buffer
-                                                          : new_node_bufs[0]->m_up_buffer, // real in place child
-                                    new_node_bufs,                                         // new node bufs
-                                    freed_node_bufs                                        // free_node_bufs
-        );
+        icp_ctx->add_to_txn_journal(index_ordinal, parent_buf, child_buf, new_node_bufs, freed_node_bufs);
```
koujl, line 296: Just record the original input bufs in journal - the up_buf links will be recovered in `IndexCPContext::process_txn_record`.

src/include/homestore/btree/detail/btree_mutate_impl.ipp
shosseinimotlagh: ok . the meta buffer is link to root(or new root) why do we need to pass the new argument?
It is like meta->root (whichever it is)
Then in merge and split with create the rest of DAG

shosseinimotlagh: please elaborate for me the necessity of this change with a simple counterexample

src/include/homestore/btree/detail/btree_node.hpp
```diff
@@ -527,15 +537,10 @@ class BtreeNode : public sisl::ObjLifeCounter< BtreeNode > {
...
-#if 0
 #ifdef _PRERELEASE
-       if (iomgr_flip::instance()->test_flip("btree_merge_node") && occupied_size() < node_data_size) {
-           return true;
-       }
-
-       auto ret = iomgr_flip::instance()->get_test_flip< uint64_t >("btree_merge_node_pct");
-       if (ret && occupied_size() < (ret.get() * node_data_size() / 100)) { return true; }
-#endif
+        if (min_keys_in_node()) {
```
shosseinimotlagh, line 541: what if we need min_keys_in_node is 0? should we set it as 1?

src/include/homestore/index/index_table.hpp
```diff
@@ -235,12 +240,18 @@ class IndexTable : public IndexTableBase, public Btree< K, V > {
         }
 
         auto& root_buf = static_cast< IndexBtreeNode* >(new_root.get())->m_idx_buf;
-        wb_cache().transact_bufs(ordinal(), m_sb_buffer, root_buf, {}, {}, r_cast< CPContext* >(context));
+        IndexBufferPtrList freed_bufs;
+        if (freed_root) {
+            freed_bufs.push_back(static_cast<IndexBtreeNode *>(freed_root.get())->m_idx_buf);
+        }
+        // Meta is similar to a leftmost child here - it should always be the up buffer for (both) root(s)
+        wb_cache().transact_bufs(ordinal(), nullptr, m_sb_buffer, {root_buf}, freed_bufs,
```
shosseinimotlagh, line 248: what would be the problem of meta to be parent? would you please elaborate here?

src/include/homestore/index/index_table.hpp
```diff
@@ -250,7 +261,15 @@ class IndexTable : public IndexTableBase, public Btree< K, V > {
...
             return btree_status_t::not_found;
         }
-        BT_LOG(INFO, "Repairing node={} with last_parent_key={}", parent_node->to_string(),
+
+        // Get all original child ids as a support to check if we are beyond the last child node
+        std::set<bnodeid_t> orig_child_ids;
+        for (uint32_t i = 0; i < parent_node->total_entries(); ++i) {
+            BtreeLinkInfo link_info;
+            parent_node->get_nth_value(i, &link_info, true);
+            orig_child_ids.insert(link_info.bnode_id());
```
shosseinimotlagh, line 270: what is the reason of having orig_child_ids?

shosseinimotlagh (commented):
Suppose there is two consecutive merges: 

         A 
     B     C   D
1- first merge: free D , DAG: A-> C
2- second merge: Free C, DAG: A->B->C(deleted is true)
Where does your change handle this situation in recover?

src/tests/test_index_crash_recovery.cpp
```diff
@@ -377,13 +402,15 @@ struct IndexCrashTest : public test_common::HSTestHelper, BtreeTestHelper< TestT
 
     void crash_and_recover(uint32_t s_key, uint32_t e_key) {
         this->print_keys("Btree prior to CP and susbsequent simulated crash: ");
-        test_common::HSTestHelper::trigger_cp(false);
+        trigger_cp(false);
         this->wait_for_crash_recovery();
         // this->visualize_keys("tree_after_crash_" + std::to_string(s_key) + "_" + std::to_string(e_key) + ".dot");
 
         this->print_keys("Post crash and recovery, btree structure: ");
```
shosseinimotlagh, line 409: add sanity check here

src/lib/index/wb_cache.cpp
```diff
@@ -289,13 +293,8 @@ void IndexWBCache::transact_bufs(uint32_t index_ordinal, IndexBufferPtr const& p
...
         icp_ctx->add_to_txn_journal(index_ordinal, parent_buf, nullptr, {child_buf}, {});
     } else {
-        icp_ctx->add_to_txn_journal(index_ordinal,          // Ordinal
-                                    child_buf->m_up_buffer, // real up buffer
-                                    new_node_bufs.empty() ? freed_node_bufs[0]->m_up_buffer
-                                                          : new_node_bufs[0]->m_up_buffer, // real in place child
-                                    new_node_bufs,                                         // new node bufs
-                                    freed_node_bufs                                        // free_node_bufs
-        );
+        icp_ctx->add_to_txn_journal(index_ordinal, child_buf->m_up_buffer /* real up buffer */, child_buf,
```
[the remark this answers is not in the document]
shosseinimotlagh: Resolved!

src/include/homestore/index/index_table.hpp
```diff
@@ -312,31 +354,59 @@ class IndexTable : public IndexTableBase, public Btree< K, V > {
             }
 
             // Insert the last key of the child node into parent node
-            cur_parent->insert(cur_parent->total_entries(), child_last_key,
-                               BtreeLinkInfo{child_node->node_id(), child_node->link_version()});
+            if (!child_node->is_node_deleted()) {
+                cur_parent->insert(cur_parent->total_entries(),
+                                   child_node->total_entries() > 0 ? child_last_key : last_parent_key,
```
shosseinimotlagh, line 359: b) leave the empty children but assign unused parent last keys to the empty children

src/include/homestore/btree/detail/btree_node.hpp
koujl: Since here's returning `total_entries() < min_keys_in_node()`, I don't think setting it to 0 can be meaningful. Or do you mean using `min_keys_in_node=0` to disable merge?

src/lib/device/virtual_dev.cpp
```diff
@@ -424,6 +424,8 @@ std::error_code VirtualDev::sync_write(const char* buf, uint32_t size, BlkId con
 
     Chunk* chunk;
     uint64_t const dev_offset = to_dev_offset(bid, &chunk);
+    HS_LOG(TRACE, device, "Writing sync in device: {}, offset = {}", chunk->physical_dev_mutable()->pdev_id(),
```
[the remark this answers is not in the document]
koujl: I think we might need some sync write traces as well - but if it can be flooding in some cases it's ok to remove them.

src/include/homestore/index/index_table.hpp
koujl: Refer to https://docs.google.com/presentation/d/1NUvu1ZlDKqJWfBayg2lSWHSB6TEaZ4Y3VbnGc996drM/edit#slide=id.g3068b64e06d_0_0

Corner case 2 & 3, since the first/last key can both change and exceed the original parent last key, we need this table to see if they should belong to the parent.

src/lib/index/wb_cache.cpp
```diff
@@ -764,7 +806,7 @@ void IndexWBCache::get_next_bufs_internal(IndexCPContext* cp_ctx, uint32_t max_c
         std::optional< IndexBufferPtr > buf = cp_ctx->next_dirty();
         if (!buf) { break; } // End of list
 
-        if ((*buf)->m_wait_for_down_buffers.testz()) {
+        if ((*buf)->state() == index_buf_state_t::DIRTY && (*buf)->m_wait_for_down_buffers.testz()) {
```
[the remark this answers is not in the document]
koujl: I met a case where the fiber in `async_cp_flush` was iterating over dirty buffers, while `process_write_completion` tried to track potential dirty up_buffer - these two processes can reach the same buf multiple times.

src/include/homestore/index/index_table.hpp
koujl: I removed this change since it may not be that necessary and I cannot recall the detailed case it was for.
We can go back to this later when we find more corner cases regarding the root node.

shosseinimotlagh (approved):

koujl:
Rebased onto HEAD of master.

koujl:
@JacksonYao287 Can you help with an approval for merging? Just lost Mehdi's previous approval due to rebase.

JacksonYao287 (approved):

PR #557, opened 2024-09-22
Draft cp repl_svc

xiaoxichen:


raakella1 (commented):
LG! I will wait for the commit to call repl_dev's switc_over_cp callback before other customers.

raakella1 (approved):
LG!

src/include/homestore/checkpoint/cp.hpp
```diff
@@ -70,6 +70,11 @@ class CPContext;
 class CPManager;
 
 VENUM(cp_consumer_t, uint8_t,
+      // Sealer is a special consumer that provides information regarding where the cp is up to.
+      // It will be the first one during cp switch over , as a conservative marker of everything
+      // before or equals to this point, should be in current cp, possibly some consumer are above this point which is fine.
+      // And Sealer is the last one during cp flush after all other services flushed successfully.
+      SEALER = 3,
```
JacksonYao287, line 77: SEAKER has the same emum number as REPLICATION_SVC, both of them are 3 ?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1103,15 +1104,31 @@ void RaftReplDev::cp_flush(CP* cp) {
 
     std::unique_lock lg{m_sb_mtx};
     m_rd_sb->compact_lsn = clsn;
-    m_rd_sb->durable_commit_lsn = lsn;
+    // dc_lsn is also flushed in flush_durable_commit_lsn()
+    // we need to take a max to avoid rolling back.
+    m_rd_sb->durable_commit_lsn = std::max(lsn, m_rd_sb->durable_commit_lsn);
```
JacksonYao287, line 1109: here, lsn is equal to m_commit_upto_lsn.load(), and m_rd_sb->durable_commit_lsn should always be equal or less than lsn, right?
in which case, lsn is bigger than  m_rd_sb->durable_commit_lsn?

src/lib/checkpoint/cp_mgr.cpp
```diff
@@ -218,20 +224,18 @@ void CPManager::cp_start_flush(CP* cp) {
...
         auto& consumer = m_cp_cb_table[svcid];
         if (consumer) { futs.emplace_back(std::move(consumer->cp_flush(cp))); }
     }
 
     folly::collectAllUnsafe(futs).thenValue([this, cp](auto) {
-        // Sync flushing replication svc at last as the cp_lsn updated here
-        // other component should at least flushed to cp_lsn
-        auto& repl_cp = m_cp_cb_table[(size_t)cp_consumer_t::REPLICATION_SVC];
-        if (repl_cp) {repl_cp->cp_flush(cp).wait();}
+        // Sync flushing SEALER svc which is the replication service
```
JacksonYao287, line 234: sorry, I am still confused here. why SEALER here stand for replication service. maybe I miss something?

src/include/homestore/checkpoint/cp.hpp
xiaoxichen: it is not made for dynamic registration which I dont think we have the use case atm.

src/lib/replication/repl_dev/raft_repl_dev.cpp
xiaoxichen: the lsn is from L1096, which was captured from the state_machine during cp_switchover(say LSN=100)
There can be some commits happen in the state machine afterward,  when we reaching here, the latest LSN from state machine can be 140,  but the dc_lsn, persisted every 500ms by reaper_thread, can be 120.

src/lib/checkpoint/cp_mgr.cpp
xiaoxichen: we make SEALER = REPLICATION_SVC = 3.

There is one special service , in our case , is replication svc , act as a sealer ,  it "seals" a cp by labling which LSN the cp contains up to. Also , it needs to be flushed at last.  So whatever we read out from the *sealer* svc ,  it guarantee other svcs also at least flushing to this point

src/lib/replication/service/raft_repl_service.h
```diff
@@ -83,6 +83,23 @@ class RaftReplService : public GenericReplService,
...
+// cp context for repl_dev, repl_dev cp_lsn is critical cursor in the system,
+// anything below the cp_lsn we believed is persisted through cp and will not
+// go through replay.  The cp_lsn need to be kept into ctx when switchover_cp,
+// and the persist of repl_dev_cp need to be done after all other consumers succeed.
+
+struct ReplDevCPContext;
+
+class ReplSvcCPContext : public CPContext {
+    std::shared_mutex m_cp_map_mtx;
+    std::map< cshared< ReplDev >, cshared<ReplDevCPContext> > m_cp_ctx_map;
```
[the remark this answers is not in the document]
xiaoxichen: make sense, thx

src/lib/checkpoint/cp_mgr.cpp
JacksonYao287: oh, I see, actually SEALER is an alias of REPLICATION_SVC, if we have another svc need to be flush at last, then it will be the new sealer, right?

xiaoxichen: We dont support 2 svcs need to be flushed at last ATM. Will extend based on need.

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: sorry , I have give a wrong question in my above comment. 
I was planning to ask in which case lsn is smaller than m_rd_sb->durable_commit_lsn?

xiaoxichen: if CP happens right after dc_lsn persist. DC_LSN persist to 100,  we made 2 more commits then do a CP:)

JacksonYao287: >if CP happens right after dc_lsn persist. DC_LSN persist to 100, we made 2 more commits then do a CP:)

this also means m_rd_sb->durable_commit_lsn should always be equal or less than lsn, right?
I mean we can do `m_rd_sb->durable_commit_lsn = lsn` , this is enough.

xiaoxichen: Let me try to explain this one more time.

We capture LSN from state-machine during cp_switchover(say LSN=100 at that time), saved into cp ctx.
During CP flush  here , the we use lsn:100 to update cp_lsn and maybe dc_lsn.


However, there is a time window, between cp_switchover and cp_flush,  the state machine still moving ahead, it can be 120 when we reaching cp_flush. And we have reaper_thread which persist dc_lsn , it can happen at the moment to persist 120.  So at the moment,  `m_rd_sb->durable_commit_lsn == 120`

Now the situation is:
state machine has committed up to 120
index and other component (cp) persist up to 100.

We should make cp_lsn = 100 and keep dc_lsn = 120.

JacksonYao287: ok , make sense

JacksonYao287 (approved):

raakella1 (approved):
LG!

PR #559, opened 2024-09-23
Replace member commit quorum

sanebay:


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -166,20 +174,32 @@ AsyncReplResult<> RaftReplDev::replace_member(replica_id_t member_out_uuid, repl
...
                     return make_async_success<>();
                 });
         });
 }
 
+void RaftReplDev::reset_quorum_size(uint32_t commit_quorum) {
+    RD_LOGI("Reset raft quorum size={}", commit_quorum);
+    nuraft::raft_params params = raft_server()->get_current_params();
+    params.with_custom_commit_quorum_size(commit_quorum);
+    params.with_custom_election_quorum_size(commit_quorum);
```
xiaoxichen, line 199: I am not sure it is right to set election_quorum_size = 1 as once the member_in added, it is possible they both elected themselves as leader.

But on the other side,  w/o this value,  if the last member alive is not the leader ,  can we proceed?

xiaoxichen:
@sanebay  lgtm , could you pls update the conan version?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -190,11 +197,22 @@ AsyncReplResult<> RaftReplDev::replace_member(replica_id_t member_out_uuid, repl
                     } else {
                         RD_LOGI("Replace member removed member={} from group_id={}", member_out, group_id_str());
                     }
+
+                    // Revert the quorum size back to 0.
+                    reset_quorum_size(0);
```
xiaoxichen, line 202: we better revert the quorum size after step 2?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -190,11 +197,22 @@ AsyncReplResult<> RaftReplDev::replace_member(replica_id_t member_out_uuid, repl
...
                     return make_async_success<>();
                 });
         });
 }
 
+void RaftReplDev::reset_quorum_size(uint32_t commit_quorum) {
```
JacksonYao287, line 208: do we need add some check here to disable setting those quorum sizes without intersection, since it might case  data loss or log diverging.
pls see 
https://github.com/eBay/NuRaft/blob/092997ab67eb2529c5b40556e1d301c8220c315d/docs/custom_quorum_size.md?plain=1#L29

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -165,6 +171,7 @@ AsyncReplResult<> RaftReplDev::replace_member(replica_id_t member_out_uuid, repl
             auto err = m_state_machine->propose_to_raft(std::move(rreq));
             if (err != ReplServiceError::OK) {
                 LOGERROR("Replace member propose to raft failed {}", err);
+                reset_quorum_size(0);
```
JacksonYao287, line 174: we also need reset_quorum_size back to 0 in 152 and 132 before we return an error

src/lib/replication/repl_dev/raft_repl_dev.cpp
xiaoxichen: this is used when we have double failure , e.g ( 2 out of 3 members are down )--- so for sure the number *seems* wrong.  

We need other mechanism to shot on the head of other members , ensuring they are deadly dead.

JacksonYao287: when we make this config change, there might be some on-going write or delete, we need to carefully handle this quorum size change.

src/lib/replication/repl_dev/raft_repl_dev.cpp
sanebay: We cant do any leader related operation. add/remove/propose.

src/lib/replication/repl_dev/raft_repl_dev.cpp
sanebay: We need to make these PG's read only as two nodes are down. Will check with CM, in nublocks once removed  they are never added back to cluster.

src/lib/replication/repl_dev/raft_repl_dev.cpp
sanebay: Ack

src/lib/replication/repl_dev/raft_repl_dev.cpp
xiaoxichen: After step 2 we have 2 members. Why we still need this quorumn size to move forward?

sanebay: New follower not added to quorum till the baseline is done right ?

xiaoxichen: Sorry I got it we are in 2(live)/4(all),

xiaoxichen: But you don't proceed to step3 before new member added into quorumn right?

sanebay: No its async operation. add_srv in nuraft only the leader accepted the request. CM will use hearbeats  with LSN(snapshot sync shard lsn) to know if its added or not.

xiaoxichen (approved):
lgtm

PR #566, opened 2024-10-15
Add cert watcher and restart raft service when cert is updated

yuwmao:


JacksonYao287:
I will first go through how file_watcher works , then I will be back and add my comments

shosseinimotlagh (commented):
@JacksonYao287  you can find the use case in AM https://github.corp.ebay.com/SDS/access-mgr/blob/e808ef0327b1db5be1aced15ae3d2d570f4afb2b/src/storage_engine/storage_engine.cpp#L866
Hopefully it helps you to review faster

JacksonYao287:
> @JacksonYao287 you can find the use case in AM https://github.corp.ebay.com/SDS/access-mgr/blob/e808ef0327b1db5be1aced15ae3d2d570f4afb2b/src/storage_engine/storage_engine.cpp#L866 Hopefully it helps you to review faster

@shosseinimotlagh ok, thank you

yuwmao:
FYI, here is an another use case for file_watcher in SM.
https://github.corp.ebay.com/SDS/storage_mgr/blob/main/src/lib/sm_lib.cpp#L221

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -175,6 +182,46 @@ void RaftReplService::stop() {
...
+        // wait for the deleted file to be added again
+        throw std::runtime_error(fmt::format("file {} not found! Can not start grpc server", filepath));
+    }
+    const std::unique_lock lock(raft_restart_mutex);
+    m_msg_mgr->restart_server();
+    if (deleted) { monitor_cert_changes(); }
+}
+
+bool RaftReplService::wait_for_cert(const std::string& filepath) {
+    for (auto i = cert_change_timeout; i > 0; --i) {
```
xiaoxichen, line 218: nit: cert_change_timeout looks like an absolute time ,  but the implementation make it to `cert_change_timeout * cert_check_sleep`

Suggesting make `cert_change_timeout` an absolute timeout by `i = cert_change_timeout/cert_check_sleep`

xiaoxichen (approved):
lgtm.

The cert feature is not yet tested with IO.

@yuwmao could we create an issue in SM to track the cert update while IO load is ongoing ?

src/lib/replication/service/raft_repl_service.cpp
yuwmao: Sure.

yuwmao:
> lgtm.
> 
> The cert feature is not yet tested with IO.
> 
> @yuwmao could we create an issue in SM to track the cert update while IO load is ongoing ?

Issue created, https://github.corp.ebay.com/SDS/storage_mgr/issues/143

xiaoxichen (approved):

xiaoxichen:
@shosseinimotlagh would you pls take a second look and merge if it thumb up :)

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -175,6 +182,47 @@ void RaftReplService::stop() {
...
+                   "hs_ssl_cert_watcher", ioenvironment.get_ssl_cert());
+    }
+    //monitor ssl key file
+    if (!fw->register_listener(ioenvironment.get_ssl_key(), "hs_ssl_key_watcher", cert_change_cb)) {
+        LOGERROR("Failed to register listner, {} to watch file {}, Not monitoring cert files",
+                   "hs_ssl_key_watcher", ioenvironment.get_ssl_key());
+    }
+}
+
+void RaftReplService::restart_raft_svc(const std::string filepath, const bool deleted){
```
JacksonYao287, line 206: this method will be called when cert or key is deleted. that means when key is deleted we restart once , when cert is deleted we restart once.

will there be a moment that a new key exists with a old cert?  will this bring any issue if we restart with a new key and old cert? I mean can we restart raft svc only when both key and cert is updated?

pls double check. if there is no harm to restart with a new key and old cert , then I thinks it will be good  since raft svc will be restarted soon when the new cert is updated.

also, in monitor_cert_changes, we will register listener for both key and cert. if we restart with the update of key, we will register listener for cert again. pls also double check if this is good

JacksonYao287 (commented):
logic looks good to me.

I only concern about the thread model.
cert_change_cb will create a separate thread , which means the update of the two files(key and cert) will spawn the other two thread, since cert_change_cb will be called for both cert and key.  will the spawned two threads spawn four thread?

pls think about this , and correct me if I am wrong

JacksonYao287 (approved):
OK，since after two times restart , we can get a good state eventually. it seems good , let`s go ahead

PR #567, opened 2024-10-17
FIX wbcache for put and modify long running index

shosseinimotlagh:


src/lib/index/wb_cache.cpp
```diff
@@ -431,32 +485,103 @@ void IndexWBCache::recover(sisl::byte_view sb) {
...
+                    //    │    │    └── D (FLUSHED)
+                    //    │    │         └── E (FLUSHED)
+                    //    │    └── F (FLUSHED)
+                    //    └── G (FLUSHED)
+                    // we need to recover B and A. During recover_buf(A), we need to recover F and commit its blk. So in
+                    // this case, we give a second chance to new buffers to be a part of repair. There is still a case
+                    // that committing new buffers is not needed since they are not part of recovery paths.
+                    auto up_buf = buf->m_up_buffer;
+                    while (up_buf) {
+                        recovering_upbuffers.insert(up_buf);
```
koujl, line 502: We can break here if the insertion fails to skip duplicate loops.

src/lib/index/wb_cache.cpp
```diff
@@ -431,32 +485,103 @@ void IndexWBCache::recover(sisl::byte_view sb) {
...
+            }
+        }
+    }
+    // remove the second tier buffers from the recovering_upbuffers if they are in it and upgrade their up buffers
+    for (auto const& buf : second_tier_buffers) {
+        // if buf->m_up_buffers is in recovering_upbuffers, commit the buf  and add it to l0_bufs o.w, remove the
+        // down_count of upbuffer
+        if (recovering_upbuffers.find(buf->m_up_buffer) != recovering_upbuffers.end()) {
+            m_vdev->commit_blk(buf->m_blkid);
+            l0_bufs.push_back(buf);
```
koujl, line 536: What if `buf` itself is in the `recovering_upbuffers`? This might cause duplicate decrements to the `down_buf` counter

src/lib/index/wb_cache.cpp
shosseinimotlagh: it is a set

src/lib/index/wb_cache.cpp
shosseinimotlagh: buf is new node, so it cannot be included in up buffers. Did I miss something?

src/lib/index/wb_cache.cpp
```diff
@@ -431,32 +485,103 @@ void IndexWBCache::recover(sisl::byte_view sb) {
...
+                 ++it) {
+                auto sp = it->lock();
+                if (sp && sp == buf) {
+                    found = true;
+                    buf->m_up_buffer->m_down_buffers.erase(it);
+                    break;
                 }
             }
+            HS_DBG_ASSERT(found, "Down buffer is linked to Up buf, but up_buf doesn't have down_buf in its list");
+#endif
```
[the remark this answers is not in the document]
shosseinimotlagh: It is a cosmetic change. I intend to change the type of m_down_buffers from vector to set. It can be done later

src/lib/index/wb_cache.cpp
```diff
@@ -292,10 +297,49 @@ void IndexWBCache::transact_bufs(uint32_t index_ordinal, IndexBufferPtr const& p
...
+        std::string child_str = (child_buf && child_buf->blkid().to_integer() != 0)
+            ? std::to_string(child_buf->blkid().to_integer())
+            : "empty";
+
+        fmt::format_to(std::back_inserter(txn), "\n{} - parent={} child={} new=[{}] freed=[{}]", txn_id, parent_str,
+                       child_str, new_nodes, freed_nodes);
+    }
+    LOGTRACEMOD(wbcache, "\ttranasction till now: cp: {} \n{}\n", icp_ctx->id(), txn);
+    txn_id++;
+#endif
```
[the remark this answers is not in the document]
shosseinimotlagh: Passing parameters to journal is slightly different. I prefer to have it here to exactly track the buffers.

koujl (approved):
lgtm

PR #568, opened 2024-10-18
Fix read_io in dataservice test.

xiaoxichen:
Previous code can overflow the io_size, i.e

remaining_io_size -= sub_io_size;

where sub_io_size > remaining_io_size, and
remaining_io_size is unsigned which will be
a huge number, takes ages to finish.


shosseinimotlagh (commented):
please bump up the version

shosseinimotlagh (approved):
LGTM

PR #569, opened 2024-10-21
Count in ovf headers.

xiaoxichen:
We see no space error in write_to_full ut, might
due to when left space == max_wrt_sz and we take max_wrt_sz, however two extra blks are needed.


JacksonYao287 (approved):
LGTM

PR #570, opened 2024-10-21
Change replace member api signature.

sanebay:
Changed the api signature because homeobject gets Member info with name, priority etc. But homestore api's dont have that.
Add replica member info with name, priority and id. Use replica member info for replace member api and listener callbacks.

### DIFF
--- a/src/include/homestore/replication/repl_decls.h
+++ b/src/include/homestore/replication/repl_decls.h
@@ -78,6 +78,13 @@ struct peer_info {
     uint64_t last_succ_resp_us_;
 };
 
+struct replica_member_info {
+    static constexpr uint64_t max_name_len = 128;
+    replica_id_t id;
+    char name[max_name_len];
+    int32_t priority{0};
+};
+
 } // namespace homestore
 
 // hash function definitions
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.4.67"
+    version = "6.5.1"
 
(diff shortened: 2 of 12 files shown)


xiaoxichen (approved):
LGTM except suggesting bump conan version to 6.5.1 as this is a breaking api change

src/include/homestore/replication/repl_decls.h
```diff
@@ -78,6 +78,13 @@ struct peer_info {
     uint64_t last_succ_resp_us_;
 };
 
+struct replica_member_info {
+    static constexpr uint64_t max_name_len = 32;
```
xiaoxichen, line 82: I am concerning the 32 is enough.  UUID is 128bit however string representation of UUID is 36byte due to 4 "-" are added.   And plus the "\n" it should at least 37.

As PG member change is rare,  maybe we can trade space with convenient make it 64bytes or even longer?

xiaoxichen: sorry it is 16bytes....

xiaoxichen: update:  CM use the name of SM when SM register to CM.  SM fill that with POD NAME atm.
https://github.corp.ebay.com/SDS/storage_mgr/blob/main/src/lib/sm_lib.cpp#L345-L347

Which makes the fix_len more risky , we may want to 
a). use json serialization/ deserialization similar to what we do for create_pg. 
b). make this limit larger and enforcing checks

sanebay: Made it 128 bytes which should be good enough

xiaoxichen (approved):
lgtm

JacksonYao287 (approved):
LGTM

PR #572, opened 2024-10-23
add chunksize to vchunk interface

JacksonYao287:
when implementing fixed pg size, we need to know the exact size of a real chunk. add this interface to support this function.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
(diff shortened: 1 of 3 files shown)


xiaoxichen:
can we targeting all fix_pg_size change to a branch and tested ?

JacksonYao287:
> can we targeting all fix_pg_size change to a branch and tested ?

ATM, actually, I am not quite sure how many changes will we exactly need for fix pg size. so I add this small step here, so that it will not block the initial development of fixed pg size feature.

xiaoxichen (approved):

PR #574, opened 2024-10-25
Add additional tests for replace member

sanebay:
Add OneMemberRestart , LeaderReplace, OneMemberDown test case.
Comments added in start of unit test

### DIFF
--- a/src/tests/test_raft_repl_dev_dynamic.cpp
+++ b/src/tests/test_raft_repl_dev_dynamic.cpp
@@ -15,11 +15,17 @@
 #include "test_common/raft_repl_test_base.hpp"
 
 // Dynamic tests spawn spare replica's also which can be used to add and remove from a repl dev.
-class ReplDevDynamicTest : public RaftReplDevTestBase {};
+class ReplDevDynamicTest : public RaftReplDevTestBase {
+private:
+    bool is_replica_num_in(const std::set< uint32_t >& replicas) {
+        // Check if the current replica process is in this set.
+        return replicas.count(g_helper->replica_num()) != 0 ? true : false;
+    }
+};
 
 TEST_F(ReplDevDynamicTest, ReplaceMember) {
+    LOGINFO("ReplaceMember test started replica={}", g_helper->replica_num());
     // Write some IO's, replace a member, validate all members data except which is out.
-    LOGINFO("Homestore replica={} setup completed", g_helper->replica_num());
     auto db = dbs_.back();
     auto num_replicas = SISL_OPTIONS["replicas"].as< uint32_t >();
     auto num_members = SISL_OPTIONS["replicas"].as< uint32_t >() + SISL_OPTIONS["spare_replicas"].as< uint32_t >();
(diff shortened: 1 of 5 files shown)


src/tests/test_raft_repl_dev_dynamic.cpp
```diff
@@ -110,28 +116,176 @@ TEST_F(ReplDevDynamicTest, TwoMemberDown) {
...
+    g_helper->sync_for_test_start(num_members);
+
+    this->shutdown_replica(2);
+    LOGINFO("Shutdown replica 2");
+
+    std::this_thread::sleep_for(std::chrono::seconds(3));
+    if (g_helper->replica_num() == 0) {
+        // With existing raft repl dev group, write IO's, validate and call replace_member on leader.
+        LOGINFO("Writing on leader num_io={} replica={}", num_io_entries, g_helper->replica_num());
+        this->write_on_leader(num_io_entries, true /* wait_for_commit */);
```
xiaoxichen, line 164: I think the `write_on_leader` needs to be execute on r0-r2.
In current implementation  R1/R2 skips the if-clause from L161-L171 and wait on L173.  If R1 is laggy (only commit on R0 and R3),   it is possible the R1 failed during verify

JacksonYao287: agree. in write_on_leader , the real leader will schedule real io, others will wait_for_commit.

src/tests/test_raft_repl_dev_dynamic.cpp
```diff
@@ -110,28 +116,176 @@ TEST_F(ReplDevDynamicTest, TwoMemberDown) {
...
+
+    if (g_helper->replica_num() != member_in) {
+        LOGINFO("Writing on leader num_io={} replica={}", num_io_entries, g_helper->replica_num());
+        // With existing raft repl dev group, write IO's, validate and call replace_member on leader.
+        this->write_on_leader(num_io_entries, true /* wait_for_commit */);
+
+        // Leader will return error NOT_LEADER and yield leadership, sleep and connect again
+        // to the new leader.
+        LOGINFO("Replace old leader");
+        replace_member(db, g_helper->replica_id(member_out), g_helper->replica_id(member_in), 0,
```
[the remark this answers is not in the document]
sanebay: But why client needs to know ? One option was to add additional error code which gateway needs to handle. Or send retry but gateway might retry same SM. Here the leader yield leadership and returns not leader and gateway retries by refreshing PG map.

src/tests/test_raft_repl_dev_dynamic.cpp
sanebay: write_on_leader is nop on other replica's. replica 0 is the leader and we are calling replace member with run_on_leader. I havent observed any replica's lagging in any of this or the original raft repl tests.

src/tests/test_raft_repl_dev_dynamic.cpp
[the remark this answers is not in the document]
xiaoxichen: I am not talking about GW but CM,  when CM send a PG_MOVE request and get back a NOT_LEADER, should it take it as success or fail?

src/tests/test_raft_repl_dev_dynamic.cpp
xiaoxichen: >write_on_leader is nop on other replica's. 
It is not that "nop" but waiting for IOs to finish before proceeding.

I think all of the UTs except this one follows the pattern.  For example this:
https://github.com/eBay/HomeStore/pull/574/files#diff-8d61e3935e3a521ab097bbc27963ca7e583d90d6a597d1301a0378fa2cf33854R213

src/tests/test_raft_repl_dev_dynamic.cpp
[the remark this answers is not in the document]
sanebay: Created an issue for CM to track this.

src/tests/test_raft_repl_dev_dynamic.cpp
sanebay: The reason is we are shutting down replica 2 and it doesnt db instances and write on leader will crash at line auto leader_uuid = dbs_[0]->repl_dev()->get_leader_id();

src/tests/test_raft_repl_dev_dynamic.cpp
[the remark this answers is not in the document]
xiaoxichen: why it is a CM ticket instead that we send another error code?

src/tests/test_raft_repl_dev_dynamic.cpp
xiaoxichen: yeah that makes sense, 
```if (g_helper->replica_num() != 2) ```
if this is better?

JacksonYao287: >leader will crash at line auto leader_uuid = dbs_[0]->repl_dev()->get_leader_id();

can we add a check like this `if (is_replica_num_in({0, 1, member_in}))` in write_on_leader

JacksonYao287 (commented):
generally, LGTM!

src/tests/test_raft_repl_dev_dynamic.cpp
sanebay: write_on_leader is generic and this case is specific for stop .

xiaoxichen (approved):

xiaoxichen:
i still dont think skipping `wait_for_commit()` on other leader is right....but as it is an UT,  we can revisit if we see it flips.

JacksonYao287 (approved):
LGTM！

PR #576, opened 2024-10-30
Implement GC_REPL_REQ Based on DSN to Prevent Resource Leaks

xiaoxichen:
This commit introduces a mechanism to garbage collect (GC) replication requests (rreqs) that may hang indefinitely, thereby consuming memory and disk resources unnecessarily. These rreqs can enter a hanging state under several circumstances, as outlined below:

1. Scenario with Delayed Commit:
   - Follower F1 receives LSN 100 and DSN 104 from Leader L1 and takes longer than the raft timeout to precommit/commit it.
   - L1 resends LSN 100, causing F1 to fetch the data again. Since LSN 100 was committed in a previous attempt, this log entry is skipped, leaving the rreq hanging indefinitely.

2. Scenario with Leader Failure Before Data Completion:
   - Follower F1 receives LSN 100 from L1, but before all data is fetched/pushed, L1 fails and L2 becomes the new leader.
   - L2 resends LSN 100 with L2 as the new originator. F1 proceeds with the new rreq and commits it, but the initial rreq from L1 hangs indefinitely as it cannot fetch data from the new leader L2.

3. Scenario with Leader Failure After Data Write:
   - Follower F1 receives data (DSN 104) from L1 and writes it. Before the log of LSN 100 reaches F1, L1 fails and L2 becomes the new leader.
   - L2 resends LSN 100 to F1, and F1 fetches DSN 104 from L2, leaving the original rreq hanging.

This garbage collection process cleans up based on DSN. Any rreqs in `m_repl_key_req_map`, whose DSN is already committed (`rreq->dsn < repl_dev->m_next_dsn`), will be GC'd. This is safe on the follower side, as the follower updates `m_next_dsn` during commit. Any DSN below `cur_dsn` should already be committed, implying that the rreq should already be removed from `m_repl_key_req_map`.

On the leader side, since `m_next_dsn` is updated when sending out the proposal, it is not safe to clean up based on `m_next_dsn`. Therefore, we explicitly skip the leader in this GC process.

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -1192,7 +1192,7 @@ std::pair< bool, nuraft::cb_func::ReturnCode > RaftReplDev::handle_raft_event(nu
 
             auto reqs = sisl::VectorPool< repl_req_ptr_t >::alloc();
             auto last_commit_lsn = uint64_cast(get_last_commit_lsn());
-            for (unsigned long i=0; i < entries.size(); i++) {
+            for (unsigned long i = 0; i < entries.size(); i++) {
                 auto& entry = entries[i];
                 auto lsn = start_lsn + i;
                 auto term = entry->get_term();
@@ -1201,8 +1201,8 @@ std::pair< bool, nuraft::cb_func::ReturnCode > RaftReplDev::handle_raft_event(nu
                 // skipping localize for already committed log(dup), they anyway will be discard
                 // by nuraft before append_log.
                 if (lsn <= last_commit_lsn) {
-                    RD_LOGT("Raft channel: term {}, lsn {}, skipping dup, last_commit_lsn {}",
-                            term, lsn, last_commit_lsn);
+                    RD_LOGT("Raft channel: term {}, lsn {}, skipping dup, last_commit_lsn {}", term, lsn,
+                            last_commit_lsn);
                     continue;
                 }
                 // Those LSNs already in logstore but not yet committed, will be dedup here,
@@ -1290,18 +1290,20 @@ void RaftReplDev::gc_repl_reqs() {
         // FIXME: Skipping proposer for now, the DSN in proposer increased in proposing stage, not when commit().
         // Need other mechanism.
         if (rreq->is_proposer()) {
-            RD_LOGD("Skipping rreq=[{}] due to is_proposer, elapsed_time_sec{};", rreq->to_string(), get_elapsed_time_sec(rreq->created_time()));
+            RD_LOGD("Skipping rreq=[{}] due to is_proposer, elapsed_time_sec{};", rreq->to_string(),
+                    get_elapsed_time_sec(rreq->created_time()));
             // don't clean up proposer's request
             continue;
         }
 
-        if (rreq->dsn() < cur_dsn ) {
-            RD_LOGD("legacy req with commited DSN, rreq=[{}] , dsn = {}, next_dsn = {}, gap= {}", rreq->to_string(), rreq->dsn(),
-                    cur_dsn, cur_dsn - rreq->dsn());
+        if (rreq->dsn() < cur_dsn) {
+            RD_LOGD("legacy req with commited DSN, rreq=[{}] , dsn = {}, next_dsn = {}, gap= {}", rreq->to_string(),
+                    rreq->dsn(), cur_dsn, cur_dsn - rreq->dsn());
             // FIXME: Wait till the rreq expired is obviously safer, though as commited request will
             //  be removed from map in on_commit(), we probably don't need wait till expired.
-	    if (rreq->is_expired()) {
-                RD_LOGD("Expired rreq =[{}], elapsed_time_sec {} ", rreq->to_string(), get_elapsed_time_sec(rreq->created_time()));
+            if (rreq->is_expired()) {
+                RD_LOGD("Expired rreq =[{}], elapsed_time_sec {} ", rreq->to_string(),
+                        get_elapsed_time_sec(rreq->created_time()));
                 expired_rreqs.push_back(rreq);
             }
         }
@@ -1320,8 +1322,8 @@ void RaftReplDev::gc_repl_reqs() {
             return;
         }
         if (rreq->dsn() < cur_dsn) {
-            RD_LOGD("legacy req, rreq=[{}] , dsn = {}, next_dsn = {}, gap= {}", rreq->to_string(), rreq->dsn(),
-                    cur_dsn, cur_dsn - rreq->dsn());
+            RD_LOGD("legacy req, rreq=[{}] , dsn = {}, next_dsn = {}, gap= {}", rreq->to_string(), rreq->dsn(), cur_dsn,
+                    cur_dsn - rreq->dsn());
         }
         if (rreq->is_expired()) {
             RD_LOGD("rreq=[{}] is expired, cleaning up; elapsed_time_sec{};", rreq->to_string(),
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -1265,39 +1265,79 @@ cshared< ReplDevCPContext > RaftReplDev::get_cp_ctx(CP* cp) {
 void RaftReplDev::cp_cleanup(CP*) {}
 
 void RaftReplDev::gc_repl_reqs() {
-    std::vector< int64_t > expired_keys;
-    m_state_machine->iterate_repl_reqs([this, &expired_keys](auto key, auto rreq) {
+    auto cur_dsn = m_next_dsn.load();
+    if (cur_dsn != 0) cur_dsn = cur_dsn - 1;
+    // On follower, any DSN below cur_dsn , should already be commited.
+    // That implies the rreq should already be removed from m_repl_key_req_map
+    std::vector< repl_req_ptr_t > expired_rreqs;
+
+    auto req_map_size = m_repl_key_req_map.size();
+    RD_LOGW("m_repl_key_req_map size is {};", req_map_size);
+    for (auto [key, rreq] : m_repl_key_req_map) {
+        // FIXME: Skipping proposer for now, the DSN in proposer increased in proposing stage, not when commit().
+        // Need other mechanism.
         if (rreq->is_proposer()) {
+            RD_LOGD("Skipping rreq=[{}] due to is_proposer, elapsed_time_sec{};", rreq->to_string(), get_elapsed_time_sec(rreq->created_time()));
+            // don't clean up proposer's request
+            continue;
+        }
+
+        if (rreq->dsn() < cur_dsn ) {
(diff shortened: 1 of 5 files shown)


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1265,39 +1265,79 @@ cshared< ReplDevCPContext > RaftReplDev::get_cp_ctx(CP* cp) {
...
-            return;
+            continue;
         }
 
+        if (rreq->dsn() < cur_dsn ) {
+            RD_LOGD("legacy req with commited DSN, rreq=[{}] , dsn = {}, next_dsn = {}, gap= {}", rreq->to_string(), rreq->dsn(),
+                    cur_dsn, cur_dsn - rreq->dsn());
+            // FIXME: Wait till the rreq expired is obviously safer, though as commited request will
+            //  be removed from map in on_commit(), we probably don't need wait till expired.
+	    if (rreq->is_expired()) {
```
zhiteng, line 1303: It's a mixed of 'tab' and space.  The indent seems only works fine with tabspace = 8 setup.  I think this is 'cos it was copied from original line 1275.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1265,39 +1265,79 @@ cshared< ReplDevCPContext > RaftReplDev::get_cp_ctx(CP* cp) {
...
+            return;
+        }
+        if (rreq->dsn() < cur_dsn) {
+            RD_LOGD("legacy req, rreq=[{}] , dsn = {}, next_dsn = {}, gap= {}", rreq->to_string(), rreq->dsn(),
+                    cur_dsn, cur_dsn - rreq->dsn());
+        }
         if (rreq->is_expired()) {
-            expired_keys.push_back(key);
             RD_LOGD("rreq=[{}] is expired, cleaning up; elapsed_time_sec{};", rreq->to_string(),
                     get_elapsed_time_sec(rreq->created_time()));
```
zhiteng, line 1328: I think this logging should be moved to line 1338 to be accurate since actual cleaning up action is done there.

zhiteng (changes requested):

src/lib/replication/repl_dev/raft_repl_dev.cpp
xiaoxichen: changed

zhiteng (dismissed):
lgtm

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1265,39 +1278,81 @@ cshared< ReplDevCPContext > RaftReplDev::get_cp_ctx(CP* cp) {
...
+            });
+        }
+        // 2. remove from the m_repl_key_req_map
+        if (m_repl_key_req_map.find(removing_rreq->rkey()) != m_repl_key_req_map.end()) {
+            m_repl_key_req_map.erase(removing_rreq->rkey());
+        }
+        // 3. remove from state-machine
+        if (removing_rreq->has_state(repl_req_state_t::LOG_FLUSHED)) {
+            RD_LOGW("Removing rreq [{}] from state machine, it is risky")
+            m_state_machine->unlink_lsn_to_req(removing_rreq->lsn());
```
JacksonYao287, line 1354: if the log at lsn N is overwritten(rollback happens), a new rreq will be linked to lsn N.
now , if you unlink the lsn N which is hold in the stale rreq, the new rreq will be erased from statemachine, where I believe an error will happen.

xiaoxichen: reaching here means we have <term =1, lsn =N,  DSN = K>  committed.

How can it possible we have a <term = 2, LSN = N>  in state-machine?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1191,9 +1191,22 @@ std::pair< bool, nuraft::cb_func::ReturnCode > RaftReplDev::handle_raft_event(nu
...
+            auto last_commit_lsn = uint64_cast(get_last_commit_lsn());
+            for (unsigned long i = 0; i < entries.size(); i++) {
+                auto& entry = entries[i];
+                auto lsn = start_lsn + i;
+                auto term = entry->get_term();
                 if (entry->get_val_type() != nuraft::log_val_type::app_log) { continue; }
                 if (entry->get_buf_ptr()->size() == 0) { continue; }
+                // skipping localize for already committed log(dup), they anyway will be discard
+                // by nuraft before append_log.
+                if (lsn <= last_commit_lsn) {
```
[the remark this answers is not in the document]
xiaoxichen: Yes you are right , I had thought about this and choose a lazy way

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: Assuming we have a leader L1 , it generates a rreq <term =1, lsn =100, DSN = 110>, now it`s `m_next_dsn` is 111. 
well, in the two followers, `m_next_dsn` is 100, since they do not receive rreq with dsn 101-110.  (for example , L1 is network partitioned). so in the leader`s memory ,  rreqs(term 1) with dsn 101-110  exists.

now , leader switch to a new one L2,  it generates  rreqs from <term =2, lsn=91, DSN = 101> to  <term =2, lsn=100, DSN = 110>  and they are committed.

if L1 reconnects to the network, it will received rreqs from <term =2, lsn=91, DSN = 101> to  <term =2, lsn=100, DSN = 110>,  it will create new rreq with dsn 101-110 and lsn 91-100 in memory , but please note that the `m_next_dsn` in L1 is 111 . now, if the stale(term 1 and expired) rreqs with dsn 101-110(they are less than 111) is gc, according to the code here, they will unlink lsn 91-100 in memory, which is now linked to the new rreq(term 2) with dsn 101-110.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1265,39 +1278,81 @@ cshared< ReplDevCPContext > RaftReplDev::get_cp_ctx(CP* cp) {
...
+                              blkid.to_string());
+                RD_LOGD("GC rreq: Releasing blkid={} freed successfully", blkid.to_string());
+            });
+        }
+        // 2. remove from the m_repl_key_req_map
+        if (m_repl_key_req_map.find(removing_rreq->rkey()) != m_repl_key_req_map.end()) {
+            m_repl_key_req_map.erase(removing_rreq->rkey());
+        }
+        // 3. remove from state-machine
+        if (removing_rreq->has_state(repl_req_state_t::LOG_FLUSHED)) {
```
sanebay, line 1352: While we are iterating, we delete or unlink from the same map. Is it safe ?

src/lib/replication/repl_dev/raft_repl_dev.cpp
xiaoxichen: added sanity check

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1265,39 +1278,81 @@ cshared< ReplDevCPContext > RaftReplDev::get_cp_ctx(CP* cp) {
...
-    std::vector< int64_t > expired_keys;
-    m_state_machine->iterate_repl_reqs([this, &expired_keys](auto key, auto rreq) {
+    auto cur_dsn = m_next_dsn.load();
+    if (cur_dsn != 0) cur_dsn = cur_dsn - 1;
+    // On follower, any DSN below cur_dsn , should already be commited.
+    // That implies the rreq should already be removed from m_repl_key_req_map
+    std::vector< repl_req_ptr_t > expired_rreqs;
+
+    auto req_map_size = m_repl_key_req_map.size();
+    RD_LOGW("m_repl_key_req_map size is {};", req_map_size);
```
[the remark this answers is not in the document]
xiaoxichen: Moved it to LOGI,  this logging is helpful as we hit mem-leak twice around same place that in certain cases we dont remove rreq from m_repl_key_req_map.  The first one is what you found and fixed that we forgot to remove in on_commit,  this patch is the second time...

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -221,11 +221,18 @@ uint64_t RaftStateMachine::last_commit_index() {
...
+void RaftStateMachine::unlink_lsn_to_req(int64_t lsn, repl_req_ptr_t rreq) {
     auto const it = m_lsn_req_map.find(lsn);
+    // it is possible a LSN mapped to different rreq in history
+    // due to log overwritten. Verify the rreq before removing
     if (it != m_lsn_req_map.cend()) {
-        RD_LOG(DEBUG, "Raft channel: erase lsn {},  rreq {}", lsn, it->second->to_string());
-        m_lsn_req_map.erase(lsn);
+        if (it->second == rreq) {
+            RD_LOG(DEBUG, "Raft channel: erase lsn {},  rreq {}", lsn, it->second->to_string());
+            m_lsn_req_map.erase(lsn);
```
JacksonYao287, line 231: there might a very small case that some change happens between line 229 and line 231. 
my suggestion is using `erase_if_equal` instead
https://github.com/facebook/folly/blob/30a4e783a7618f17a5b24048625872e363068887/folly/concurrency/ConcurrentHashMap.h#L497

JacksonYao287 (commented):
only a minor comment, other parts look good

src/lib/replication/repl_dev/raft_repl_dev.cpp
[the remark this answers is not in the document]
sanebay: Better keep a metrics of this repl related map's of count and total memory usage.

src/lib/replication/repl_dev/raft_repl_dev.cpp
sanebay: iterate_repl_reqs is iterating through m_lsn_req_map and unlink is erasing it from m_lsn_req_map. Its not iterator so it looks safe.

JacksonYao287 (dismissed):
LG

JacksonYao287 (approved):
LGTM! this a very nice step for sm long run

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1265,39 +1278,71 @@ cshared< ReplDevCPContext > RaftReplDev::get_cp_ctx(CP* cp) {
...
+    auto req_map_size = m_repl_key_req_map.size();
+    RD_LOGI("m_repl_key_req_map size is {};", req_map_size);
+    for (auto [key, rreq] : m_repl_key_req_map) {
+        // FIXME: Skipping proposer for now, the DSN in proposer increased in proposing stage, not when commit().
+        // Need other mechanism.
+        if (rreq->is_proposer()) {
+            // don't clean up proposer's request
+            continue;
+        }
+        if (rreq->dsn() < cur_dsn && rreq->is_expired()) {
```
JacksonYao287, line 1299: we can revisit here if we have better solution to accurately identify the garbage in the future.
for now, let`s go ahead and not block sm long run.

JacksonYao287 (approved):
LGTM for now, let`s revisit here if necessary in the future.

PR #577, opened 2024-11-01
Add protection for concurrent access to m_down_buffers

koujl:
Issue: <https://github.com/eBay/HomeStore/issues/578>

The `down_buf` has the `real_up_buf` (the meta buf) as its up_buffer, but cannot be found in the `read_up_buf->m_down_buffers`. The inconsistency is caused by unprotected concurrent writes to the vector `m_down_buffers`.

Add a mutex lock to `IndexBuffer` as well as extracting add/remove operations into member functions to make the vector thread-safe.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.5.5"
+    version = "6.5.6"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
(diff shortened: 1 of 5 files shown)


zhiteng (commented):
The PR message should be put into commit message as well with some modification, which the reference to eBay internal JIRA isn't suitable for public github.com hosted open source project.  My suggestion is to create an issue and refer to that issue instead of JIRA.  That way people from outside of eBay may see the entirety of whole issue along with this fix.

koujl:
Linked to https://github.com/eBay/HomeStore/issues/578

shosseinimotlagh (approved):
LG!
Todo: needs to change to set

Issue #580, opened 2024-11-05
OOM after follower lagging


Besroy:
Description: During the log replay process, there is a potential issue where new data is transmitted simultaneously. This concurrent data transmission can lead to an excessive accumulation of data in memory. If the data stored in memory exceed the quota, it results in OOM.

Additional Details: For a more detailed analysis and potential mitigation strategies, please refer to the documentation available at the following link: [Google Docs](https://docs.google.com/document/d/16DMqv9J-JuNs5c25IBuX01QXe5WbxPm4B4hAfz4JNZM/edit?tab=t.0#heading=h.b4j7215dfoup)

xiaoxichen:
@Besroy pls check if we can close this issue

Besroy:
Test passed, close this issue

PR #583, opened 2024-11-07
Releasing data buf from memory after written to disk.

xiaoxichen:
Data buffer persists in memory until rreq is committed or rolled back.

This approach poses issues during recovery. As new data arrives via push_data and is written to disk, it remains in memory for an extended period until the replica catches up and commits the rreq.


xiaoxichen:
fixes : https://github.com/eBay/HomeStore/issues/580

JacksonYao287 (approved):
LG

PR #584, opened 2024-11-07
PushData only pushed to active followers.

xiaoxichen:
If a follower is lagging too far, do not flood it with data from new IOs (new rreq, new LSNs) ,  reserve the capability for catching up,  that follower can request data via FetchData.

### DIFF
--- a/src/lib/common/homestore_config.fbs
+++ b/src/lib/common/homestore_config.fbs
@@ -264,6 +264,9 @@ table Consensus {
 
     // Log difference to determine if the follower is in resync mode
     resync_log_idx_threshold: int64 = 100;
+
+    // Log difference from leader's point of view,  to determine if the follower is laggy.
+    laggy_threshold: int64 = 2000;
 }
 
 table HomeStoreSettings {
--- a/src/lib/common/homestore_config.fbs
+++ b/src/lib/common/homestore_config.fbs
@@ -265,7 +265,8 @@ table Consensus {
     // Log difference to determine if the follower is in resync mode
     resync_log_idx_threshold: int64 = 100;
 
-    // Log difference from leader's point of view,  to determine if the follower is laggy.
+    // Log difference from leader's point of view,  to determine if the
+    // follower is laggy and if so, leader will stop pushing data until it drops under this threshold.
     laggy_threshold: int64 = 2000;
 }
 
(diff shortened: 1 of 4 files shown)


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -363,23 +363,30 @@ void RaftReplDev::push_data_to_all_followers(repl_req_ptr_t rreq, sisl::sg_list
...
-                return;
+    auto peers = get_active_peers();
+    auto calls = std::vector< nuraft_mesg::NullAsyncResult >();
+    for (auto peer : peers) {
+        RD_LOGD("Data Channel: Pushing data to follower {}, rreq=[{}]", peer, rreq->to_string());
+        calls.push_back(group_msg_service()
+                            ->data_service_request_unidirectional(peer, PUSH_DATA, rreq->m_pkts)
+                            .via(&folly::InlineExecutor::instance()));
+    }
+    folly::collectAllUnsafe(calls).thenValue([this, rreq](auto&& v_res) {
```
xiaoxichen, line 374: we wait till all the pushdata complete,  is that necessary? 

IMO we can return a future and continue the path of write_local, propose_to_raft etc

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -363,23 +363,30 @@ void RaftReplDev::push_data_to_all_followers(repl_req_ptr_t rreq, sisl::sg_list
...
+        for (auto const& res : v_res) {
+            if (sisl_likely(res.value())) {
+                auto r = res.value();
+                if (r.hasError()) {
+                    // Just logging PushData error, no action is needed as follower can try by fetchData.
```
JacksonYao287, line 379: this part makes sense

JacksonYao287 (approved):
LGTM

src/lib/common/homestore_config.fbs
```diff
@@ -264,6 +264,9 @@ table Consensus {
 
     // Log difference to determine if the follower is in resync mode
     resync_log_idx_threshold: int64 = 100;
+
+    // Log difference from leader's point of view,  to determine if the follower is laggy.
```
yamingk, line 268: ... to determine if the follower is laggy and if so, leader will stop pushing data until it drops under this threshold.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1020,6 +1027,25 @@ std::vector< peer_info > RaftReplDev::get_replication_status() const {
     return pi;
 }
 
+std::set< replica_id_t > RaftReplDev::get_active_peers() const {
+    auto repl_status = get_replication_status();
+    std::set< replica_id_t > res;
+    auto my_committed_idx = m_commit_upto_lsn.load();
+    uint64_t lagThreshold = my_committed_idx > HS_DYNAMIC_CONFIG(consensus.laggy_threshold)
```
yamingk, line 1034: nit: since HomeStore uses `my_local_var` name format, can we change to `lag_threshold`?

yamingk (approved):

yamingk (approved):

Issue #582, opened 2024-11-06
Implement rollback


Besroy:
When the leader is restarted, it will become follower and append logs from new leader, if the log exists, nuraft will call rollback to rollback uncommited logs, then overwrite log store progress. Since we haven't implement rollback on state machine yet, when the log store progress is overwritten and log is appended, lsn_to_req will return old req.

details in [issue7](https://docs.google.com/document/d/16DMqv9J-JuNs5c25IBuX01QXe5WbxPm4B4hAfz4JNZM/edit?tab=t.0#heading=h.xb2h24mjbmr2)

PR #585, opened 2024-11-08
add rollback on state machine add open Leader_Restart ut

Besroy:
https://github.com/eBay/HomeStore/issues/582

### DIFF
--- a/src/lib/replication/repl_dev/raft_state_machine.cpp
+++ b/src/lib/replication/repl_dev/raft_state_machine.cpp
@@ -208,8 +208,13 @@ void RaftStateMachine::commit_config(const ulong log_idx, raft_cluster_config_pt
     // TODO:add more logic here if necessary
 }
 
-void RaftStateMachine::rollback(const ulong log_idx, nuraft::buffer& data) {
-    int64_t lsn = s_cast< int64_t >(log_idx);
+void RaftStateMachine::rollback_config(const ulong log_idx, raft_cluster_config_ptr_t& conf) {
+    RD_LOGD("Raft channel: Rollback cluster conf , log_idx = {}", log_idx);
+    // TODO:add more logic here if necessary
+}
+
+void RaftStateMachine::rollback_ext(const nuraft::state_machine::ext_op_params& params) {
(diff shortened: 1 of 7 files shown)


src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -208,6 +208,18 @@ void RaftStateMachine::commit_config(const ulong log_idx, raft_cluster_config_pt
     // TODO:add more logic here if necessary
 }
 
+void RaftStateMachine::rollback(const ulong log_idx, nuraft::buffer& data) {
```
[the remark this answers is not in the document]
xiaoxichen:
```
    virtual void rollback_ext(const ext_op_params& params)
    {   rollback(params.log_idx, *params.data);  }
```
default implementation of `rollback_ext` called rollback.
I think we can choose to implement `rollback_ext` but not mandatory.

xiaoxichen (dismissed):
lgtm.

Also good to see we enabled one more UT.

xiaoxichen (dismissed):

xiaoxichen (approved):

src/tests/test_raft_repl_dev.cpp
```diff
@@ -152,6 +152,7 @@ TEST_F(RaftReplDevTest, Resync_From_Non_Originator) {
 }
 
 #if 0
```
yamingk, line 154: Can this test case be enabled or not yet?

xiaoxichen: @Besroy had tried to enable this but failed.  The test of restart leader is passing in Storage hammer but not here..  @JacksonYao287 traiged this, pls share more findings?

Besroy: As far as I know, the UT's `write_on_leader` function cannot detect changes in the leader. If a leader switch occurs while writing data, this UT will get stuck. Although it works locally, it may fail occasionally in Jenkins, which is why it has been commented out again. @JacksonYao287 Please correct me if I'm wrong.

PR #586, opened 2024-11-11
Checking received data size and reject if not match.

xiaoxichen:
We dont need to panic in this case, fetchData can handle this.


JacksonYao287 (approved):
LG

PR #587, opened 2024-11-13
Set min_log_gap_to_join to max_int32.

xiaoxichen:


JacksonYao287:
looks good, can you also set `use_new_joiner_type_` to true in this PR?

src/lib/common/homestore_config.fbs
```diff
@@ -242,7 +242,7 @@ table Consensus {
     stale_log_gap_lo_threshold: int32 = 30;
 
     // Minimum log gap a replica has to be from leader before joining the replica set.
-    min_log_gap_to_join: int32 = 30;
+    min_log_gap_to_join: int32 = 2147483647;
```
JacksonYao287, line 245: NIT : can we change 2147483647 to some kind like 0x7FFFFFFF or INT32_MAX , which will be more readable.

JacksonYao287 (approved):
LGTM

PR #588, opened 2024-11-13
handle nagtive log batch size returned by follower

JacksonYao287:
when follower hits some error before appending log entries, it will set `batch_size_hint_in_bytes` to -1 to ask leader do not send more log entries in the next append_log_req.
https://github.com/eBay/NuRaft/blob/eabdeeda538a27370943f79a2b08b5738b697ac3/src/handle_append_entries.cxx#L760

in nuobject case , if a new member is added to a raft group and it tries to append create_shard log entry , which will try to alllocate block from the chunks of the pg,  before the create_pg log is committed , which will allocated chunks to this pg, and error will happen and the log batch containing create_shard log entry will be wholy rejected and set `batch_size_hint_in_bytes` to -1  in the response to leader.

this pr aims to set the log count in the next batch sent to follower to 1, so that: 

if the create_pg and create_shard are in the same log batch , the pr will first reject this log batch and leader will send only create_pg in the next batch , which will be accepted by follower , since  it will only create this pg.

if the create_pg and create_shard are not in the same log batch, and create_shard is trying to allocate block before the pg it created(chunks of this pg is alllocated), then , with this pr, follower will reject this batch so that it will give more time to creating pg.  create_shard log will be resent in the next batch , and at that moment pg  has probably already been successfully created.

### DIFF
--- a/src/lib/replication/log_store/home_raft_log_store.cpp
+++ b/src/lib/replication/log_store/home_raft_log_store.cpp
@@ -192,9 +192,7 @@ void HomeRaftLogStore::write_at(ulong index, nuraft::ptr< nuraft::log_entry >& e
 
         // remove all cached entries after this index
         for (size_t i{0}; i < m_log_entry_cache.size(); ++i) {
-            if (m_log_entry_cache[i].first > index) {
-                m_log_entry_cache[i] = std::make_pair(0, nullptr);
-            }
+            if (m_log_entry_cache[i].first > index) { m_log_entry_cache[i] = std::make_pair(0, nullptr); }
         }
     }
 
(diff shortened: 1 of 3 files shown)


xiaoxichen (commented):
lgtm

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -223,6 +221,18 @@ nuraft::ptr< std::vector< nuraft::ptr< nuraft::log_entry > > > HomeRaftLogStore:
...
 
+nuraft::ptr< std::vector< nuraft::ptr< nuraft::log_entry > > >
+HomeRaftLogStore::log_entries_ext(ulong start, ulong end, int64_t batch_size_hint_in_bytes) {
+    // in nuraft , batch_size_hint_in_bytes < 0 indicats that follower is busy now and do not want to receive any more
+    // log entries ATM. here we just send one log entry if this happens which is helpful for nuobject case and no harm
+    // to other case.
+    if (batch_size_hint_in_bytes < 0) end = start + 1;
+
+    // for the case where batch_size_hint_in_bytes >= 0, we do not take any size check here for now.
+    // TODO: limit the size of the returned entries by batch_size_hint_in_bytes int the future if necessary
```
xiaoxichen, line 232: @yamingk we discussed pressure_back , do you think we can use the batch_size_hint for this purpose?
i.e. follower can respond with `batch_size_hint` to leader through the response of appendEntries request.

xiaoxichen (approved):

src/lib/replication/log_store/home_raft_log_store.cpp
yamingk: yes, looks like follower can set the next_batch_size_hint in append_entries_resp and leader will set it in its next batch size and consume it when next batch happens:
https://github.com/eBay/NuRaft/blob/eabdeeda538a27370943f79a2b08b5738b697ac3/src/handle_append_entries.cxx#L1069~1072

PR #590, opened 2024-11-13
push data asynchronously at leader

JacksonYao287:
now we always start appending log to log store at leader until all the futures of pushing data to followers are completed , which will involve unnecessary latency

this PR aims to make pushing data asynchronously,  so that logs can be set to follower ASAP. it does not matter whether push_data is successful, since follower will try to fetch data from the leader if push_data fails

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.5.11"
+    version = "6.5.12"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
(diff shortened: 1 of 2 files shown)


xiaoxichen (commented):
it seems right but I would like to hold this till we solve and bake a stable version for PG move.

JacksonYao287:
> it seems right but I would like to hold this till we solve and bake a stable version for PG move.

sure, will also loop other members in

JacksonYao287:
>we dont wait for the future at all

we use `folly::collectAllUnsafe` to wait for the completion of all the future, no?

xiaoxichen:
> collectAllUnsafe

https://github.com/facebook/folly/blob/main/folly/futures/Future.h#L2360-L2385
it returns a future which satisfied when all *to_be_collected* features satisfied.

JacksonYao287:
>  collectAllUnsafe

I see , you are right. I misunderstood the meaning of `collectAllUnsafe`. now I think , as you said , there is no difference compared to `run_on_forget`. let me close this pr

PR #592, opened 2024-11-14
Add debugging log for index cp.

xiaoxichen:


src/lib/index/wb_cache.cpp
```diff
@@ -167,6 +167,8 @@ bool IndexWBCache::get_writable_buf(const BtreeNodePtr& node, CPContext* context
     if (idx_buf->m_dirtied_cp_id == icp_ctx->id()) {
         return true; // For same cp, we don't need a copy, we can rewrite on the same buffer
     } else if (idx_buf->m_dirtied_cp_id > icp_ctx->id()) {
+        LOGINFOMOD(wbcache, "cp_id = {} < buffer dirtied_cp_id, rejecting, node hdr {}", icp_ctx->id(),
+                   idx_buf->m_dirtied_cp_id, ((persistent_hdr_t*)idx_buf->m_bytes)->to_string());
```
koujl, line 171: Need to add braces after dirtied_cp_id

src/lib/index/index_service.cpp
```diff
@@ -89,7 +89,7 @@ void IndexService::start() {
     }
     // Force taking cp after recovery done. This makes sure that the index table is in consistent state and dirty buffer
     // after recovery can be added to dirty list for flushing in the new cp
-    hs()->cp_mgr().trigger_cp_flush(true /* force */);
+    hs()->cp_mgr().trigger_cp_flush(true /* force */).wait();
```
shosseinimotlagh, line 92: trigger_cp_flush(true) waits until cp flush is done. Did you test it without .wait()

xiaoxichen: Right now all the tests are done without the wait()... I dont know if we do need a wait() but it seems make things easier...

shosseinimotlagh (approved):

PR #594, opened 2024-11-19
handle config change event in statemachine

JacksonYao287:
in nuobject, when a member is move out of a raft group, we should clear the pg related resource in the member , which will be done in the config change handler. this PR aims to add this interface


xiaoxichen (changes requested):
I think we need more evidence to expose the "raft implementation details" to HO.
As the `raft_cluster_config_ptr_t` is a ptr to `nuraft::cluster_config` which is very much  implementation related structure.

I think the interpretation should be in HS and call back to listener with specific callback, like `on_joining_cluster`, `on_leaving_cluster`, etc.

JacksonYao287:
> like `on_joining_cluster`, `on_leaving_cluster`
this is a good suggestion, will make this change

xiaoxichen:
Then in this case, should we use 
https://github.com/eBay/NuRaft/blob/092997ab67eb2529c5b40556e1d301c8220c315d/include/libnuraft/callback.hxx#L95
and 
https://github.com/eBay/NuRaft/blob/092997ab67eb2529c5b40556e1d301c8220c315d/include/libnuraft/callback.hxx#L101


more callback type can be consumed from `handle_raft_event`

JacksonYao287:
> more callback type can be consumed from `handle_raft_event`

1 in this PR , I only want to handle the leave cluster case. we can create another PR for join cluster case

2 `RemovedFromCluster ` callback will be called before [the new configuration is applied](https://github.com/eBay/NuRaft/blob/eabdeeda538a27370943f79a2b08b5738b697ac3/src/handle_commit.cxx#L902) , IMO, it is better to call it after the new configuration is applied.
additionally,  `handle_raft_event` is designed to handle raft level or repl_dev level event,  it is better to offload upper layer event  to the statemachine related callbacks.

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -204,8 +204,49 @@ raft_buf_ptr_t RaftStateMachine::commit_ext(nuraft::state_machine::ext_op_params
...
 
 void RaftStateMachine::commit_config(const ulong log_idx, raft_cluster_config_ptr_t& new_conf) {
+    // when reaching here, the config change log has already been committed, and the new config has been applied to the
+    // cluster
+
     RD_LOGD("Raft channel: Commit new cluster conf , log_idx = {}", log_idx);
-    // TODO:add more logic here if necessary
+    auto& servers_in_new_conf = new_conf->get_servers();
+    std::unordered_set< int32_t > server_ids_in_new_conf;
+    for (auto& server : servers_in_new_conf)
```
yamingk, line 213:
```
for (auto& server : servers_in_new_conf) {
#ifdef _PRERELEASE
        server_ids_in_new_conf.insert(server->get_id());
#endif
    if (m_rd.server_id() == server->getid()) {
         m_rd.m_listener->on_leave_cluster(m_rd.group_id());
         m_rd.leave();
#ifndef _PRERELEASE
         break;
#endif
    }
}
```

src/include/homestore/replication/repl_dev.h
```diff
@@ -379,6 +380,21 @@ class ReplDevListener {
...
+    // destroy is slightly different from leave_cluster.
+
+    // 1 destory will propose a normal raft log, when this log it committed, it will move the node to destroying state,
+    // and gc thread will stop the raft server and reclaim all the resources allocated to the raft group. there is no
+    // config change, and all the member of the whole raft group will be detructed.
+
+    // 2 well , leave_cluster will propose a config change log, when this log is committed, this node will be removed
+    // from the cluster configuration. so, this callback will only be called on the removed node, which is no longer in
+    // the cluster configuration
+    virtual void on_leave_cluster(const group_id_t& group_id){
```
yamingk, line 393: is there a reason we can't bump conan version to 6.6 and make this pure virtual function, and when make HomeObject PR, consume 6.6 homestore to consume this change?

JacksonYao287: the only concern is that before homeobject implements this function, all other PR of homeobject will get a build failure in CI build.

src/lib/replication/repl_dev/raft_state_machine.cpp
JacksonYao287: remove , I make the whole part off from release build. we will not handle leave cluster event here

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1276,9 +1276,44 @@ std::pair< bool, nuraft::cb_func::ReturnCode > RaftReplDev::handle_raft_event(nu
...
+        // raft group.
+
+        // applicaiton should cleanup related applicaiton layer resource
+
+        // in nuobject case this is expected to
+        // 1 delete indextable and pg_suerblk
+        // 2 reset the append blk allocator all the chunks of this pg
+        // 3 return all the chunks to chunkselector.
+
+        m_listener->on_leave_cluster(group_id());
```
xiaoxichen, line 1298: we called `on_destroy()` in `leave()`,  why do we want an extra callback?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1276,9 +1276,44 @@ std::pair< bool, nuraft::cb_func::ReturnCode > RaftReplDev::handle_raft_event(nu
             sisl::VectorPool< repl_req_ptr_t >::free(reqs);
         }
         return {true, ret};
-    } else {
-        return {false, ret};
     }
+
+    case nuraft::cb_func::Type::RemovedFromCluster: {
+        // a node will reach here when :
```
xiaoxichen, line 1282: this is a good writting to summarize the intent

xiaoxichen:
I hope the commits and commit message can be cleanup ,  there are two parts in this PR:

1. consume `nuraft::cb_func::Type::RemovedFromCluster` callback
2. add `reset` function to allocator/vchunk as a preparation for implementing `m_listener->on_destroy()`

JacksonYao287:
> The pr title/commit messages are a bit confusing.

sure, will refine this commit message

xiaoxichen (approved):
lgtm

PR #595, opened 2024-11-22
Fix grpc crash

Besroy:
If data channel and raft channel release the same data at the same time, there is a chance that send_response may be called twice, then grpc will crash with error TOO_MANY_OPERATIONS. This pr make sure data channel release data at first.

Signed-off-by: Yawen Zhang <yawzhang@ebay.com>

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.5.16"
+    version = "6.5.17"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/lib/replication/repl_dev/common.cpp
+++ b/src/lib/replication/repl_dev/common.cpp
@@ -164,7 +164,6 @@ bool repl_req_ctx::add_state_if_not_already(repl_req_state_t s) {
 void repl_req_ctx::clear() {
     m_header = sisl::blob{};
     m_key = sisl::blob{};
-    release_data();
     m_pkts.clear();
 }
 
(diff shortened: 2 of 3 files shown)


xiaoxichen (approved):

JacksonYao287 (approved):
now, before we append log, we will make sure that the data corresponding to this log is flushed and the rreq state is changed to DATA_WRITTEN.

so I am think can we remove the release data op in commit phrase?

Besroy:
now, before we append log, we will make sure that the data corresponding to this log is flushed and the rreq state is changed to DATA_WRITTEN.

so I am think can we remove the release data op in commit phrase?

I think remove release data in req->clear() is a strength way to avoid double release ---  actually my fix cannot guarantee the sequence 'data channel release -> append log->commit release', because I found another conflict:
1. data channel set DATA_WRITTEN at https://github.com/eBay/HomeStore/blob/master/src/lib/replication/repl_dev/raft_repl_dev.cpp#L456 at first
2. check data received or not before append logs: if state==DATA_WRITTEM, will return directly at https://github.com/eBay/HomeStore/blob/master/src/lib/replication/repl_dev/raft_repl_dev.cpp#L545
3. data channel release_data && handle_commit release_data at the same time

And other way to guarantee the sequence is to add lock on release_data. cc: @JacksonYao287  @xiaoxichen @raakella1

xiaoxichen:
@Besroy i think you mean here
https://github.com/eBay/HomeStore/blob/f83679af8ce71e71fb5b6f57df6d54047cd7d940/src/lib/replication/repl_dev/raft_repl_dev.cpp#L875？

There are 3 places we do release_data()

1. on_push_data_reveive()
2. handle_fetch_data_response()
3. on_commit()

conflicts between 1/3 and 2/3 can be resolved by either remove 3, or finish `release_data` in data path before setting the DATA_WRITTEN flag.

Conflict between 1/2 can happen when push_data arrives later, log arrived first and triggered fetch_data, there are two parallel data path operations in flight.

I think adding a lock is a right fix similar to the block allocation we did in `applier_create_req`.

JacksonYao287 (approved):
LGTM

PR #596, opened 2024-11-25
Support Baseline Resync

yuwmao:
For Nuraft baseline resync, we separate the process into two layers: HomeStore layer and Application layer.
We use the first bit of the `obj_id` to indicate the message type: 0 is for HS, 1 is for Application.

In the HomeStore layer, leader needs to transmit the DSN to the follower, this is intended to handle the following case:
- Leader sends snapshot at LSN T1 to follower F1.
- F1 fully receives the snapshot and now at T1.
- Leader yield its leadership, F1 elected as leader.
In this sequence the incremental resync will not kicked in to update the m_next_dsn, and as result, duplication may occur.

### DIFF
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -112,9 +112,9 @@ private:
     nuraft::ptr< nuraft::snapshot > snapshot_;
 };
 
-struct snapshot_data {
+struct snapshot_obj {
     void* user_ctx{nullptr};
-    int64_t offset{0};
+    uint64_t offset{0};
     sisl::io_blob_safe blob;
     bool is_first_obj{false};
     bool is_last_obj{false};
@@ -368,16 +368,16 @@ public:
     /// uses offset given by the follower to the know the current state of the follower.
     /// Leader sends the snapshot data to the follower in batch. This callback is called multiple
     /// times on the leader till all the data is transferred to the follower. is_last_obj in
-    /// snapshot_data will be true once all the data has been trasnferred. After this the raft on
+    /// snapshot_obj will be true once all the data has been trasnferred. After this the raft on
     /// the follower side can do the incremental resync.
-    virtual int read_snapshot_data(shared< snapshot_context > context, shared< snapshot_data > snp_data) = 0;
+    virtual int read_snapshot_obj(shared< snapshot_context > context, shared< snapshot_obj > snp_obj) = 0;
 
     /// @brief Called on the follower when the leader sends the data during the baseline resyc.
-    /// is_last_obj in in snapshot_data will be true once all the data has been transfered.
+    /// is_last_obj in in snapshot_obj will be true once all the data has been transfered.
     /// After this the raft on the follower side can do the incremental resync.
-    virtual void write_snapshot_data(shared< snapshot_context > context, shared< snapshot_data > snp_data) = 0;
+    virtual void write_snapshot_obj(shared< snapshot_context > context, shared< snapshot_obj > snp_obj) = 0;
 
-    /// @brief Free up user-defined context inside the snapshot_data that is allocated during read_snapshot_data.
+    /// @brief Free up user-defined context inside the snapshot_obj that is allocated during read_snapshot_obj.
     virtual void free_user_snp_ctx(void*& user_snp_ctx) = 0;
 
 private:
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -124,6 +124,8 @@ struct snapshot_obj {
     bool is_last_obj{false};
 };
 
+//HomeStore has some meta information to be transmitted during the baseline resync,
+//Although now only dsn needs to be synced, this structure is defined as a general message, and we can easily add data if needed in the future.
 struct snp_repl_dev_data {
     uint64_t magic_num{HOMESTORE_RESYNC_DATA_MAGIC};
     uint32_t protocol_version{HOMESTORE_RESYNC_DATA_PROTOCOL_VERSION_V1};
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -46,6 +46,10 @@ VENUM(journal_type_t, uint16_t,
       HS_CTRL_REPLACE = 3, // Control message to replace a member
 )
 
+// magic num comes from the first 8 bytes of 'echo homestore_resync_data | md5sum'
+static constexpr uint64_t HOMESTORE_RESYNC_DATA_MAGIC = 0xa65dbd27c213f327;
+static constexpr uint32_t HOMESTORE_RESYNC_DATA_PROTOCOL_VERSION_V1 = 0x01;
+
 struct repl_key {
     int32_t server_id{0}; // Server Id which this req is originated from
     uint64_t term;        // RAFT term number
@@ -118,6 +122,13 @@ struct snapshot_obj {
     sisl::io_blob_safe blob;
     bool is_first_obj{false};
     bool is_last_obj{false};
+};
+
+struct snp_repl_dev_data {
+    uint64_t magic_num{HOMESTORE_RESYNC_DATA_MAGIC};
+    uint32_t protocol_version{HOMESTORE_RESYNC_DATA_PROTOCOL_VERSION_V1};
+    uint32_t crc{0};
+    uint64_t dsn{0};
 };
 
 struct repl_journal_entry;
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -1495,7 +1495,7 @@ void RaftReplDev::create_snp_resync_data(raft_buf_ptr_t& data_out) {
     snp_repl_dev_data msg;
     auto msg_size = sizeof(snp_repl_dev_data);
     msg.dsn = m_next_dsn;
-    auto crc = crc32_ieee(0, reinterpret_cast< const unsigned char* >(&msg), msg_size);
+    auto crc = crc32_ieee(init_crc32, reinterpret_cast< const unsigned char* >(&msg), msg_size);
     RD_LOGD("create snapshot resync msg, dsn={}, crc={}", msg.dsn, crc);
     msg.crc = crc;
     data_out = nuraft::buffer::alloc(msg_size);
@@ -1511,8 +1511,9 @@ bool RaftReplDev::apply_snp_resync_data(nuraft::buffer& data) {
     }
     auto received_crc = msg->crc;
     RD_LOGD("received snapshot resync msg, dsn={}, crc={}, received crc={}", msg->dsn, msg->crc, received_crc);
+    // Clear the crc field before verification, because the crc value computed by leader doesn't contain it.
     msg->crc = 0;
-    auto computed_crc = crc32_ieee(0, reinterpret_cast< const unsigned char* >(msg),
+    auto computed_crc = crc32_ieee(init_crc32, reinterpret_cast< const unsigned char* >(msg),
                                    sizeof(snp_repl_dev_data));
     if (received_crc != computed_crc) {
         RD_LOGE("Snapshot resync data crc mismatch, received_crc={}, computed_crc={}", received_crc, computed_crc);
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -1491,6 +1491,37 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
     handle_commit(rreq, true /* recovery */);
 }
 
+void RaftReplDev::create_snp_resync_data(raft_buf_ptr_t& data_out) {
+    snp_repl_dev_data msg;
+    auto msg_size = sizeof(snp_repl_dev_data);
+    msg.dsn = m_next_dsn;
+    auto crc = crc32_ieee(0, reinterpret_cast< const unsigned char* >(&msg), msg_size);
+    RD_LOGD("create snapshot resync msg, dsn={}, crc={}", msg.dsn, crc);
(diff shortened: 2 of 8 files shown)


JacksonYao287 (commented):
if I understand correctly , this pr aims to update the m_next_dsn in the snapshot receiver side, right?

for the receiver , after baseline resync is completed, incremental resync will start, where new log entries will be recevied and committed.  according to the code here
https://github.com/eBay/HomeStore/blob/6a2dfd8e212a9a4b5c9b17c3d1de45acb3907ce2/src/lib/replication/repl_dev/raft_repl_dev.cpp#L933-L934

m_next_dsn will be updated by handle_commit (committing the received log entry), and thus m_next_dsn will eventually be  up-to-date as leader. so , I am not sure is it necessary to add a separate step to sync m_next_dsn.

pls correct me if I am wrong or miss something

xiaoxichen:
We discussed this case in the review,

The case is,

- Leader sends snapshot at LSN T1 to follower F1.
- F1 fully receives the snapshot and now at T1.
- Leader yield its leadership, F1 elected as leader.

In this sequence the incremental resync will not kicked in to update the m_next_dsn.

Yes now it only update the `m_next_dsn` but more generically ,  it is a channel for baseline resync at HS level.

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -297,14 +297,20 @@ void RaftStateMachine::create_snapshot(nuraft::snapshot& s, nuraft::async_result
 
 int RaftStateMachine::read_logical_snp_obj(nuraft::snapshot& s, void*& user_ctx, ulong obj_id, raft_buf_ptr_t& data_out,
                                            bool& is_last_obj) {
+    if ((obj_id & snp_obj_id_type_mask) == 0) {
```
xiaoxichen, line 300: any `obj id < 0x8000000000000000` will result this statement to be true.
I think you want is `obj_id & snp_obj_id_type_mask ==  snp_obj_id_type_mask`

src/lib/replication/repl_dev/raft_state_machine.h
```diff
@@ -86,6 +86,8 @@ class StateMachineStore;
 #define RD_LOGE(...) RD_LOG(ERROR, ##__VA_ARGS__)
 #define RD_LOGC(...) RD_LOG(CRITICAL, ##__VA_ARGS__)
 
+static constexpr uint64_t snp_obj_id_type_mask = 0x8000000000000000;
```
koujl, line 89: Let's add some comments to describe our design of obj_id encoding format here. Otherwise readers may be confused about the origin of this magic number.

And suggest using bit shifting instead of raw hex to represent these values much more clearly (same for `get_next_lsn`).

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1491,6 +1491,37 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
...
+bool RaftReplDev::apply_snp_resync_data(nuraft::buffer& data) {
+    auto msg = r_cast< snp_repl_dev_data* >(data.data_begin());
+    if (msg->magic_num != HOMESTORE_RESYNC_DATA_MAGIC || msg->protocol_version !=
+        HOMESTORE_RESYNC_DATA_PROTOCOL_VERSION_V1) {
+        RD_LOGE("Snapshot resync data validation failed, magic={}, version={}", msg->magic_num, msg->protocol_version);
+        return false;
+    }
+    auto received_crc = msg->crc;
+    msg->crc = 0;
+    RD_LOGD("received snapshot resync msg, dsn={}, crc={}, received crc={}", msg->dsn, msg->crc, received_crc);
```
koujl, line 1514: Why we're still printing `msg->crc` here as it's just set to 0?

src/include/homestore/replication/repl_dev.h
```diff
@@ -112,14 +116,21 @@ class nuraft_snapshot_context : public snapshot_context {
...
+struct snapshot_obj {
     void* user_ctx{nullptr};
-    int64_t offset{0};
+    uint64_t offset{0};
     sisl::io_blob_safe blob;
     bool is_first_obj{false};
     bool is_last_obj{false};
 };
 
+struct snp_repl_dev_data {
```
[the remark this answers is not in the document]
yuwmao: Yes, will add an explanation comment for this struct and PR.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1491,6 +1491,37 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
...
+    auto computed_crc = crc32_ieee(0, reinterpret_cast< const unsigned char* >(msg),
+                                   sizeof(snp_repl_dev_data));
+    if (received_crc != computed_crc) {
+        RD_LOGE("Snapshot resync data crc mismatch, received_crc={}, computed_crc={}", received_crc, computed_crc);
+        return false;
+    }
+    m_next_dsn = msg->dsn;
```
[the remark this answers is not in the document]
yuwmao: Ok.

src/lib/replication/repl_dev/raft_state_machine.cpp
yuwmao: For the first bit(type bit), 0 is for HomeStore and 1 is for Application(HomeObject).
So if obj id < 0x8000000000000000, it is expected to be handled by HS.

src/lib/replication/repl_dev/raft_state_machine.h
yuwmao: Sure

src/lib/replication/repl_dev/raft_repl_dev.cpp
yuwmao: Sorry, it should be put after the RD_LOGD. Setting to 0 aims at excluding it to compute crc.

JacksonYao287:
> Leader yield its leadership, F1 elected as leader.

yes , theoretically , this will happen.  moreover , we need to trigger a cp_flush after baseline resync is done to guarantee m_next_dsn is persisted to metaservice to prevent this from happening even if F1 restarts. cc @yuwmao

yuwmao:
> > Leader yield its leadership, F1 elected as leader.
> 
> yes , theoretically , this will happen. moreover , we need to trigger a cp_flush after baseline resync is done to guarantee m_next_dsn is persisted to metaservice to prevent this from happening even if F1 restarts. cc @yuwmao

Ok, in our design, the last obj is an empty message, let's call cp_flush in the last message.

xiaoxichen:
> > > Leader yield its leadership, F1 elected as leader.
> > 
> > 
> > yes , theoretically , this will happen. moreover , we need to trigger a cp_flush after baseline resync is done to guarantee m_next_dsn is persisted to metaservice to prevent this from happening even if F1 restarts. cc @yuwmao
> 
> Ok, in our design, the last obj is an empty message, let's call cp_flush in the last message.

I would suggest do that in apply_snapshot(), we also update lsn there.

yuwmao:
> > > > Leader yield its leadership, F1 elected as leader.
> > > 
> > > 
> > > yes , theoretically , this will happen. moreover , we need to trigger a cp_flush after baseline resync is done to guarantee m_next_dsn is persisted to metaservice to prevent this from happening even if F1 restarts. cc @yuwmao
> > 
> > 
> > Ok, in our design, the last obj is an empty message, let's call cp_flush in the last message.
> 
> I would suggest do that in apply_snapshot(), we also update lsn there.

Ok, make sense.

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -363,7 +365,10 @@ bool RaftStateMachine::apply_snapshot(nuraft::snapshot& s) {
     m_rd.set_last_commit_lsn(s.get_last_log_idx());
     m_rd.m_data_journal->set_last_durable_lsn(s.get_last_log_idx());
     auto snp_ctx = std::make_shared< nuraft_snapshot_context >(s);
-    return m_rd.m_listener->apply_snapshot(snp_ctx);
+    auto res = m_rd.m_listener->apply_snapshot(snp_ctx);
+    //make sure the changes are flushed.
+    hs()->cp_mgr().trigger_cp_flush(true /* force */);
```
xiaoxichen, line 370: .get() to wait till the cp finish.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1491,6 +1491,41 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
     handle_commit(rreq, true /* recovery */);
 }
 
+void RaftReplDev::create_snp_resync_data(raft_buf_ptr_t& data_out) {
+    snp_repl_dev_data msg;
+    auto msg_size = sizeof(snp_repl_dev_data);
+    msg.dsn = m_next_dsn;
+    auto crc = crc32_ieee(0, reinterpret_cast< const unsigned char* >(&msg), msg_size);
```
yamingk, line 1498: is there a reason to use `0` over `init_crc32`?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1491,6 +1491,41 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
...
+bool RaftReplDev::apply_snp_resync_data(nuraft::buffer& data) {
+    auto msg = r_cast< snp_repl_dev_data* >(data.data_begin());
+    if (msg->magic_num != HOMESTORE_RESYNC_DATA_MAGIC || msg->protocol_version !=
+        HOMESTORE_RESYNC_DATA_PROTOCOL_VERSION_V1) {
+        RD_LOGE("Snapshot resync data validation failed, magic={}, version={}", msg->magic_num, msg->protocol_version);
+        return false;
+    }
+    auto received_crc = msg->crc;
+    RD_LOGD("received snapshot resync msg, dsn={}, crc={}, received crc={}", msg->dsn, msg->crc, received_crc);
+    msg->crc = 0;
```
yamingk, line 1514: why it needs to be reset to 0 here?

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -297,14 +297,22 @@ void RaftStateMachine::create_snapshot(nuraft::snapshot& s, nuraft::async_result
...
-    auto snp_data = std::make_shared< snapshot_data >();
+    auto snp_data = std::make_shared< snapshot_obj >();
     snp_data->user_ctx = user_ctx;
     snp_data->offset = obj_id;
     snp_data->is_last_obj = is_last_obj;
 
     // Listener will read the snapshot data and we pass through the same.
-    int ret = m_rd.m_listener->read_snapshot_data(snp_ctx, snp_data);
+    int ret = m_rd.m_listener->read_snapshot_obj(snp_ctx, snp_data);
     if (ret < 0) return ret;
```
yamingk, line 316: Is anyone catching this error code?

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -320,8 +328,16 @@ int RaftStateMachine::read_logical_snp_obj(nuraft::snapshot& s, void*& user_ctx,
 
 void RaftStateMachine::save_logical_snp_obj(nuraft::snapshot& s, ulong& obj_id, nuraft::buffer& data, bool is_first_obj,
                                             bool is_last_obj) {
+    if ((obj_id & snp_obj_id_type_mask) == 0) {
```
yamingk, line 331: Can we move `(obj_id & snp_obj_id_type_mask) == 0` to a function, e.g. `is_hs_msg(obj_id)`

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -320,8 +328,16 @@ int RaftStateMachine::read_logical_snp_obj(nuraft::snapshot& s, void*& user_ctx,
...
+        // Homestore preserved msg
+        if (m_rd.apply_snp_resync_data(data)) {
+            obj_id = snp_obj_id_type_mask;
```
yamingk, line 334: I am afraid I don't quite follow the logic here, what is the purpose of this line to set obj_id to this type_mask, by reading the log, it says it will be next obj id, but now sure how we achieved that by assigning to the type mask?

src/lib/replication/repl_dev/raft_state_machine.cpp
yuwmao: NuRaft will handle this error and finish the snapshot. In HomeObject read logic, we add some retries for get_blob.
https://github.com/eBay/NuRaft/blob/cda7dffee79bc0899d51917a9765eb0392ae958f/src/handle_snapshot_sync.cxx#L207

src/lib/replication/repl_dev/raft_state_machine.cpp
yamingk: okay, should we rename `snp_obj_id_type_mask` to `snp_obj_id_type_app`, which seems to be the case here right?

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: I see the point. Can we add a brief comment here to explain it?

src/lib/replication/repl_dev/raft_state_machine.cpp
yuwmao: Ok, `snp_obj_id_type_app` is clearer.

yamingk:
A general comment: is there a plan to add unit test case (with flip) for the error path, such as read_snapshot_obj, apply_snapshot_obj, save_snapshot_obj, etc to verify the retry logic will be happening in the baseline resync flow?
It is okay to do it with another PR.

xiaoxichen (approved):

PR #597, opened 2024-11-26
Support flexible virtual device creation in HomeStore with num_chunks or chunk_size

Hooper9973:
Support flexible virtual device creation in `homestore::BlkDataService` with num_chunks or chunk_size. Prioritize `num_chunks` over `chunk_size` if both are provided.

### DIFF
--- a/src/lib/blkdata_svc/blkdata_service.cpp
+++ b/src/lib/blkdata_svc/blkdata_service.cpp
@@ -38,7 +38,7 @@ BlkDataService::~BlkDataService() = default;
 
 // first-time boot path
 void BlkDataService::create_vdev(uint64_t size, HSDevType devType, uint32_t blk_size, blk_allocator_type_t alloc_type,
-                                 chunk_selector_type_t chunk_sel_type, uint32_t num_chunks) {
+                                 chunk_selector_type_t chunk_sel_type, uint32_t num_chunks, uint32_t chunk_size) {
     hs_vdev_context vdev_ctx;
     vdev_ctx.type = hs_vdev_type_t::DATA_VDEV;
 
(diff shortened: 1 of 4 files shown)


src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -55,6 +55,24 @@ void BlkDataService::create_vdev(uint64_t size, HSDevType devType, uint32_t blk_
                                                         .context_data = vdev_ctx.to_blob()});
 }
 
+void BlkDataService::create_vdev_with_chunk_size(uint64_t size, HSDevType devType, uint32_t blk_size, blk_allocator_type_t alloc_type,
```
xiaoxichen, line 58: why we want separate function instead of passing both chunk_size and num_chunks in? 

in `create_vdev` we check one and only one is not zero

src/lib/homestore.cpp
```diff
@@ -224,13 +224,33 @@ void HomeStore::format_and_start(std::map< uint32_t, hs_format_params >&& format
...
         } else if ((svc_type & HS_SERVICE::DATA) && has_data_service()) {
-            m_data_service->create_vdev(pct_to_size(fparams.size_pct, fparams.dev_type), fparams.dev_type,
-                                        fparams.block_size, fparams.alloc_type, fparams.chunk_sel_type,
-                                        fparams.num_chunks);
+            if (fparams.num_chunks != 0) {
+                m_data_service->create_vdev_with_num_chunks(pct_to_size(fparams.size_pct, fparams.dev_type),
+                                                            fparams.dev_type, fparams.block_size, fparams.alloc_type,
+                                                            fparams.chunk_sel_type,
+                                                            fparams.num_chunks);
+            } else if (fparams.chunk_size != 0) {
```
JacksonYao287, line 232: when creating vdev, device_manager always prioritize num_chunks over chunk_size
https://github.com/eBay/HomeStore/blob/6a2dfd8e212a9a4b5c9b17c3d1de45acb3907ce2/src/lib/device/device_manager.cpp#L243-L244

if we want to create vdev according to num_chunks , we set it to non-zero, otherwist, we set num_chunks to zero ,and set chunk_size to what we want

Hooper9973: good idea

xiaoxichen (approved):

PR #599, opened 2024-12-03
Implement get_next_batch_size_hint_in_bytes()

xiaoxichen:
we use the `byte` as `cnt` as of now.

Also update the log_entries_ext() which will be called on leader,

if hint < 0 means follower want nothing, return an empty vector so that an empty append_entries_req will be sent, to carry the commit_index update and trigger the follower to commit.

if hint > 0, respect the cnt that the follower want,  this is useful when two logs within same batch has dependency, we can exclude the dependent one.

if hint = 0 means control by leader.

### DIFF
--- a/src/lib/replication/log_store/home_raft_log_store.cpp
+++ b/src/lib/replication/log_store/home_raft_log_store.cpp
@@ -223,14 +223,27 @@ nuraft::ptr< std::vector< nuraft::ptr< nuraft::log_entry > > > HomeRaftLogStore:
 
 nuraft::ptr< std::vector< nuraft::ptr< nuraft::log_entry > > >
 HomeRaftLogStore::log_entries_ext(ulong start, ulong end, int64_t batch_size_hint_in_bytes) {
-    // in nuraft , batch_size_hint_in_bytes < 0 indicats that follower is busy now and do not want to receive any more
-    // log entries ATM. here we just send one log entry if this happens which is helpful for nuobject case and no harm
-    // to other case.
-    if (batch_size_hint_in_bytes < 0) end = start + 1;
-
-    // for the case where batch_size_hint_in_bytes >= 0, we do not take any size check here for now.
-    // TODO: limit the size of the returned entries by batch_size_hint_in_bytes int the future if necessary
-    return log_entries(start, end);
+    // WARNING: we interpret batch_size_hint_in_bytes as count as of now.
+    auto batch_size_hint_cnt = batch_size_hint_in_bytes;
+    auto new_end = end;
+    // batch_size_hint_in_bytes < 0 indicats that follower is busy now and do not want to receive any more log entry.
+    if (batch_size_hint_cnt < 0)
+        new_end = start;
+    else if (batch_size_hint_cnt > 0) {
+        // limit to the hint, also prevent overflow by a huge batch_size_hint_cnt
+        if (sisl_unlikely(start + (uint64_t)batch_size_hint_cnt < start)) {
+            new_end = end;
+        } else {
+            new_end = start + (uint64_t)batch_size_hint_cnt;
+        }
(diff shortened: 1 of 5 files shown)


src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -249,6 +249,27 @@ void RaftStateMachine::rollback_ext(const nuraft::state_machine::ext_op_params&
...
+
+int64_t RaftStateMachine::inc_next_batch_size_hint() {
+    constexpr int64_t next_batch_size_hint_limit = 16;
+    // set to minimal if previous hint is negative (i.e do not want any log)
+    if (next_batch_size_hint < 0) {
+        next_batch_size_hint = 1;
+        return next_batch_size_hint;
+    }
+    // Exponential growth till next_batch_size_hint_limit,  set to 0 afterward means leader take control.
+    next_batch_size_hint = next_batch_size_hint * 2 > next_batch_size_hint_limit ? 0 : next_batch_size_hint;
```
JacksonYao287, line 264: should we change it to 

```
next_batch_size_hint = next_batch_size_hint * 2 > next_batch_size_hint_limit ? 0 : next_batch_size_hint * 2;
```

JacksonYao287 (commented):
this tcp-like traffic(log sending) control looks good to me

xiaoxichen:
thanks @JacksonYao287 , addressed.

JacksonYao287 (approved):
LGTM.

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -223,14 +223,27 @@ nuraft::ptr< std::vector< nuraft::ptr< nuraft::log_entry > > > HomeRaftLogStore:
...
+        // limit to the hint, also prevent overflow by a huge batch_size_hint_cnt
+        if (sisl_unlikely(start + (uint64_t)batch_size_hint_cnt < start)) {
+            new_end = end;
+        } else {
+            new_end = start + (uint64_t)batch_size_hint_cnt;
+        }
+        // limit to original end
+        new_end = std::min(new_end, end);
+    }
+    DEBUG_ASSERT(new_end <= end, "new end {} should be <= original end {}", new_end, end);
```
yamingk, line 242: nit: `DEBUG_ASSERT_LE(new_ned, end);`

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1261,6 +1261,13 @@ std::pair< bool, nuraft::cb_func::ReturnCode > RaftReplDev::handle_raft_event(nu
                 auto req = m_state_machine->localize_journal_entry_prepare(*entry);
                 if (req == nullptr) {
                     sisl::VectorPool< repl_req_ptr_t >::free(reqs);
+		    // The hint set here will be used by the next after next appendEntry, the next one
+		    // always go with -1 from NuRraft code.
+		    //
+                    // We are rejecting this log entry, meaning we can accept previous log entries.
+		    // If there is nothing we can accept(i==0), that maens we are waiting for commit
+		    // of previous lsn, set it to 1 in this case.
+                    m_state_machine->reset_next_batch_size_hint(std::max(1ul, i));
```
yamingk, line 1270: Can you help me understand if we are setting 1 here, how can we get `batch_size_hint_cnt` to be set as -1  at `HomeRaftLogStore::log_entries_ext`, so that we can apply empty append entries and only send commit to the follower?

xiaoxichen: https://github.com/eBay/NuRaft/blob/1adcc6282109c2ddf1121bbc374d48d303145e39/src/handle_append_entries.cxx#L703

PR #600, opened 2024-12-03
update is_proposer definition

Besroy:
Define is_proposer by comparing m_raft_server_id with rkey.server_id

### DIFF
--- a/src/lib/replication/repl_dev/raft_state_machine.cpp
+++ b/src/lib/replication/repl_dev/raft_state_machine.cpp
@@ -154,15 +154,9 @@ repl_req_ptr_t RaftStateMachine::localize_journal_entry_finish(nuraft::log_entry
 
     auto rreq = m_rd.repl_key_to_req(rkey);
     if ((rreq == nullptr) || (rreq->is_localize_pending())) {
-        rreq = localize_journal_entry_prepare(lentry);
-        if (rreq == nullptr) {
-            RELEASE_ASSERT(rreq != nullptr,
-                           "We get an linked data for rkey=[{}], jentry=[{}] not as part of Raft Append but "
-                           "indirectly through possibly unpack() and in those cases, if we are not able to alloc "
-                           "location to write the data, there is no recourse. So we must crash this system ",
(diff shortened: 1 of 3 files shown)


xiaoxichen:
please move the issue from google doc to gh issue
do not link any google doc from gh

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -154,6 +154,7 @@ repl_req_ptr_t RaftStateMachine::localize_journal_entry_finish(nuraft::log_entry
 
     auto rreq = m_rd.repl_key_to_req(rkey);
     if ((rreq == nullptr) || (rreq->is_localize_pending())) {
+        RD_LOGE("Should not happen in current implementation! rkey=[{}]", rkey.to_string());
```
xiaoxichen, line 157: //this path is for log pack apply only.  Now we fully disabled log pack/unpack with the new_learner optt, should not reaching here.

Besroy: Will gc hit this issue? For example:
t1: commit lsn 10 with dsn 20
t2: append lsn 11 with dsn 10
t3: in gc_repl_reqs, found lsn11->dsn is less than m_next_dsn and req is expired, remove req
t4: commit lsn11 but cannot find rreq in the map

JacksonYao287: this check happens before log is written to log store , so commit will not reach here.
but theoretically, if the append log thread is lagging too far, this rreq might be gc before the log is appended to log store(after the log is received but before it is appended to log store)

so I think this might happen

xiaoxichen: In general there is a lot of corner cases that we cannot handle, if gc_repl_reqs released a req that it should not to. Basically because we dont have a determine way to identify a req as "stale"

The `gc_repl_reqs` should be conservative as much as possible.
Other pieces of code is fine to assert or panic so next restart probably solve the issue, but should not corrupt data or hang.

PR #602, opened 2024-12-08
Add on_repl_devs_init_completed cb.

xiaoxichen:
A stable callback is needed regardless whether we have repl_dev created.

This CB is a nice place for upper layer to recover those depends on repl_dev.


JacksonYao287 (approved):
LGTM

PR #604, opened 2024-12-10
fix handle_error and update log

Besroy:
Also add volatile vs non-volatile to code more self-explain and easier to follow some corner cases (for example https://github.com/eBay/HomeStore/issues/607)

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -594,7 +594,8 @@ folly::Future< folly::Unit > RaftReplDev::notify_after_data_written(std::vector<
     });
 }
 
-bool RaftReplDev::wait_for_data_receive(std::vector< repl_req_ptr_t > const& rreqs, uint64_t timeout_ms) {
+bool RaftReplDev::wait_for_data_receive(std::vector< repl_req_ptr_t > const& rreqs, uint64_t timeout_ms,
+                                        std::vector< repl_req_ptr_t >* timeout_rreqs) {
     std::vector< folly::Future< folly::Unit > > futs;
     std::vector< repl_req_ptr_t > only_wait_reqs;
     only_wait_reqs.reserve(rreqs.size());
@@ -621,14 +622,23 @@ bool RaftReplDev::wait_for_data_receive(std::vector< repl_req_ptr_t > const& rre
 
     // We are yet to support reactive fetch from remote.
     if (is_resync_mode()) {
-        check_and_fetch_remote_data(std::move(only_wait_reqs));
+        check_and_fetch_remote_data(only_wait_reqs);
     } else {
-        m_repl_svc.add_to_fetch_queue(shared_from_this(), std::move(only_wait_reqs));
+        m_repl_svc.add_to_fetch_queue(shared_from_this(), only_wait_reqs);
     }
 
     // block waiting here until all the futs are ready (data channel filled in and promises are made);
-    auto all_futs = folly::collectAllUnsafe(futs).wait(std::chrono::milliseconds(timeout_ms));
-    return (all_futs.isReady());
+    auto all_futs_ready = folly::collectAllUnsafe(futs).wait(std::chrono::milliseconds(timeout_ms)).isReady();
+    if (!all_futs_ready && timeout_rreqs != nullptr) {
+        timeout_rreqs->clear();
+        for (size_t i{0}; i < futs.size(); ++i) {
+            if (!futs[i].isReady()) {
+                timeout_rreqs->emplace_back(only_wait_reqs[i]);
+            }
+        }
+        all_futs_ready = timeout_rreqs->empty();
+    }
+    return all_futs_ready;
 }
 
 void RaftReplDev::check_and_fetch_remote_data(std::vector< repl_req_ptr_t > rreqs) {
@@ -953,18 +963,25 @@ void RaftReplDev::handle_commit(repl_req_ptr_t rreq, bool recovery) {
 
 void RaftReplDev::handle_error(repl_req_ptr_t const& rreq, ReplServiceError err) {
     if (err == ReplServiceError::OK) { return; }
+    RD_LOGE("Raft Channel: Error in processing rreq=[{}] error={}", rreq->to_string(), err);
 
     if (!rreq->add_state_if_not_already(repl_req_state_t::ERRORED)) {
-        RD_LOGE("Raft Channel: Error in processing rreq=[{}] error={}", rreq->to_string(), err);
+        RD_LOGE("Raft Channel: Error has been added for rreq=[{}] error={}", rreq->to_string(), err);
         return;
     }
 
     // Remove from the map and thus its no longer accessible from applier_create_req
     m_repl_key_req_map.erase(rreq->rkey());
 
-    if (rreq->op_code() == journal_type_t::HS_DATA_INLINED) {
+    // Ensure non-volatile lsn not exist because handle_error should not be called after append entries.
+    auto exist_rreq = m_state_machine->lsn_to_req(rreq->lsn());
+    if (exist_rreq != nullptr && !exist_rreq->is_volatile()) {
+        HS_REL_ASSERT(false, "Unexpected: LSN={} is already ready to commit, exist_rreq=[{}]",
+                      rreq->lsn(), exist_rreq->to_string());
+    }
+
+    if (rreq->op_code() == journal_type_t::HS_DATA_LINKED) {
         // Free the blks which is allocated already
-        RD_LOGE("Raft Channel: Error in processing rreq=[{}] error={}", rreq->to_string(), err);
         if (rreq->has_state(repl_req_state_t::BLK_ALLOCATED)) {
             auto blkid = rreq->local_blkid();
             data_service().async_free_blk(blkid).thenValue([blkid](auto&& err) {
@@ -1288,8 +1305,9 @@ nuraft::cb_func::ReturnCode RaftReplDev::raft_event(nuraft::cb_func::Type type, 
             }
 
             // Wait till we receive the data from its originator for all the requests
-            if (!wait_for_data_receive(*reqs, HS_DYNAMIC_CONFIG(consensus.data_receive_timeout_ms))) {
-                for (auto const& rreq : *reqs) {
+            std::vector< repl_req_ptr_t > timeout_rreqs;
+            if (!wait_for_data_receive(*reqs, HS_DYNAMIC_CONFIG(consensus.data_receive_timeout_ms), &timeout_rreqs)) {
+                for (auto const& rreq : timeout_rreqs) {
                     handle_error(rreq, ReplServiceError::TIMEOUT);
                 }
                 ret = nuraft::cb_func::ReturnCode::ReturnNull;
@@ -1480,11 +1498,15 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
     RD_DBG_ASSERT(happened, "rreq already exists for rkey={}", rkey.to_string());
     uint32_t data_size{0u};
 
+    // If the data is linked and value_size is non-zero, it means blks have been allocated for data.
+    // Since the log is flushed after data is written, the data has already been received.
     if ((jentry->code == journal_type_t::HS_DATA_LINKED) && (jentry->value_size > 0)) {
         MultiBlkId entry_blkid;
         entry_blkid.deserialize(entry_to_val(jentry), true /* copy */);
         data_size = entry_blkid.blk_count() * get_blk_size();
         rreq->set_local_blkid(entry_blkid);
+        rreq->add_state(repl_req_state_t::BLK_ALLOCATED);
+        rreq->add_state(repl_req_state_t::DATA_RECEIVED);
     }
 
     rreq->set_lsn(repl_lsn);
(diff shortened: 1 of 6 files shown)


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -953,18 +953,20 @@ void RaftReplDev::handle_commit(repl_req_ptr_t rreq, bool recovery) {
...
     if (!rreq->add_state_if_not_already(repl_req_state_t::ERRORED)) {
-        RD_LOGE("Raft Channel: Error in processing rreq=[{}] error={}", rreq->to_string(), err);
+        RD_LOGE("Raft Channel: Error has been added for rreq=[{}] error={}", rreq->to_string(), err);
         return;
     }
 
     // Remove from the map and thus its no longer accessible from applier_create_req
     m_repl_key_req_map.erase(rreq->rkey());
+    // Remove from the lsn map and thus its no longger accessible for commit (is it really needed?)
+    m_state_machine->unlink_lsn_to_req(rreq->lsn(), rreq);
```
JacksonYao287, line 966: the lsn of  a rreq is set after the corresponding log is appended.  and handle_error will always happen before log is appended. so , IMO, it will not find the lsn to the rreq and will do nothing. we can add an assert here to check if it has a lsn when handl_error of a rreq

xiaoxichen (commented):
the changes in handle_error makes sense. Just pls add more comment in API signature to state the `handle_error` should not be called after append_entries.

And we need to fix the `handle_raft_event` as well to avoid calling handle_error on logs that already in logstore.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -953,18 +963,26 @@ void RaftReplDev::handle_commit(repl_req_ptr_t rreq, bool recovery) {
...
     }
 
     // Remove from the map and thus its no longer accessible from applier_create_req
     m_repl_key_req_map.erase(rreq->rkey());
 
-    if (rreq->op_code() == journal_type_t::HS_DATA_INLINED) {
+    // Remove from the lsn map, so that it is no longer accessible from the state machine
+    // Since handle_error and append_log may occur concurrently,
+    // this removal is effective only if append_log is already processed
+    if (m_state_machine->lsn_to_req(rreq->lsn()) != nullptr) {
```
JacksonYao287, line 979: >(m_state_machine->lsn_to_req(rreq->lsn()) != nullptr

this means the log has been appended to log store and pre_committed. if we want to discard this log, should we rollback it ?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -627,8 +628,17 @@ bool RaftReplDev::wait_for_data_receive(std::vector< repl_req_ptr_t > const& rre
...
 
     // block waiting here until all the futs are ready (data channel filled in and promises are made);
-    auto all_futs = folly::collectAllUnsafe(futs).wait(std::chrono::milliseconds(timeout_ms));
-    return (all_futs.isReady());
+    auto all_futs_ready = folly::collectAllUnsafe(futs).wait(std::chrono::milliseconds(timeout_ms)).isReady();
+    if (!all_futs_ready && timeout_rreqs != nullptr) {
+        timeout_rreqs->clear();
+        for (size_t i{0}; i < futs.size(); ++i) {
+            if (!futs[i].isReady()) {
+                timeout_rreqs->emplace_back(only_wait_reqs[i]);
```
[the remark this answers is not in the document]
Besroy: rreq saved here is in m_repl_key_req_map, if the rreq is overwritten with different data, means server_id/term/dsn is same but data is not same, will that be an unexpected error? 
If the rreq in state machine is overwritten, means we have appended one req with the same lsn before, here are two senarios (pls correct me if I'm wrong):
1. role is leader, after the leader append log, it was restarted, then the new leader send a new rreq with the same lsn : resolved by rollback
2. role is follower, will append same lsn with same log entry (leader resend due to timeout), which will be skipped from raft side.

src/lib/replication/repl_dev/raft_repl_dev.cpp
Besroy: Actually I think the current scenario and issue are not very clear (it might require an emergent GC as mentioned in https://github.com/eBay/HomeStore/issues/607). The only scenario I can think of ATM is: the log is appended to the log store, but the data write fails, causing it to get stuck at the end_of_append_batch (and not get persisted). In this case, a rollback might not be necessary; instead, data should be retrieved from the data channel. Additionally, if a rollback is needed, I haven't yet figured out a suitable method to determine an appropriate rollback point.

xiaoxichen: I would suggest lets carefully using `pre_commit_ext` in our current implementation,  only do whatever volatile.

The `pre_commit_ext` was called when log append to logstore,  (in our implementation it is memory only),  upon to this point,  (in our implementation) the log is not persisted to logstore, it is the `end_of_append_batch`,  ensures the persistency.   If a crash happened before `end_of_append_batch`,  the log can be volatilized from logstore,   no `rollback` will be called if the log get over-written.


Here,  we have to add the req to map(state-machine) to allow the `on_precommit` be executed,  but in fact the log is not persisted at all and no `commit` will be execute.

Better we differentiate "volatile log in SM" vs "non-volatile log in SM",  any log, after passing `end_of_append_batch` , it should NOT be touched from outside of SM as they can be committed already.

Technics can be done for example,
 
```
int64 ReplLogStore::to_volatile_lsn(uint64 lsn) {
  assert( lsn <= int64_max);
  return - int64_cast(lsn);
}

ReplLogStore::apend() {
  ......
  m_sm.link_lsn_to_req(rreq, to_volatile_lsn(lsn));
}
 ``` 
 
 ```
void ReplLogStore::end_of_append_batch(ulong start_lsn, ulong count) {
 ....
 for (int64_t lsn = int64_cast(start_lsn); lsn <= end_lsn; ++lsn) {
        auto volatile_lsn = to_volatile_lsn(lsn)
        auto rreq = m_sm.lsn_to_req(volatile_lsn);
        m_sm.link_lsn_to_req(rreq, lsn);
         m_sm.unlink_lsn_to_req(rreq, volatile_lsn);
        
 ```

Comment (Besroy) [reply src/lib/replication/repl_dev/raft_repl_dev.cpp:979]: Agree that differentiating volatile vs non-volatile is a good approach as it clarifies the behavior of rreq in the state machine when encountering errors. For example, during error handling, if the rreq is volatile, we can perform an unlink, and if it is non-volatile, we might use ASSERT. However, I have some doubts about its necessity for the following reasons:
1. Currently, handle_error is expected to only deal with volatile logs, and non-volatile logs should not reach the error handling logic. Do we really need to introduce this concept at this stage, especially considering it reduces the available range of LSNs (even though our code defaults LSN to be int64 ATM)
2. For volatile logs, since we cannot control the order of link_lsn_to_req and failure of 'async write', when handle_error triggered, the rreq might or might not exist in the state machine. So we might still need to further discuss how to handle drive failures in the future. For example, should we check for the existence of volatile rreq in the state machine when triggering emergent gc, and if it exists, remove it? Otherwise, we may need to consider whether it is safe for link_lsn_to_req to overwrite volatile rreq.

Comment (xiaoxichen) [reply src/lib/replication/repl_dev/raft_repl_dev.cpp:979]: 1.  The target is not trying to let non-volatile rreq reach handle_error but more on how to make the code more self-explain and easier to follow,  the feeling of uncertainty that a volatile rreq might be `unlink_lsn_to_req` is bad and every time we need to comfort ourselves there is no such case.

2. If a log is made to non-volatile at the end of `end_of_append_batch`, i.e passing the wait IO, even passing the log flush, for example L96 before `    sisl::VectorPool< repl_req_ptr_t >::free(reqs);`,  will that helps?

Comment (xiaoxichen) [reply src/lib/replication/repl_dev/raft_repl_dev.cpp:979]: uint64->int64 should be fine,  int64 can last for **2.92 Million Years** at 100K tps :)

Comment (Besroy) [reply src/lib/replication/repl_dev/raft_repl_dev.cpp]: Updated. Since differentiating logs makes it easier to follow, I also update logic in `handle_error` and `link_lsn_to_req` to gather more information, especially for corner cases we might not have thought of. This change can resolve the conflict between 'append' and 'commit', but there's still an issue with the following scenario. So it's probably best not to unlink in `handle_error` and instead add some logs to check if an overwrite really happens in the state machine:

T1: Handle the raft event and pass.
T2: Append log to the log store and save the volatile LSN into the state machine.
T3: Fail to write due to an IO error, and then remove it from the state machine.
T4: In the pre-commit phase, can't find the volatile LSN, leading to a nullptr exception.

Comment (xiaoxichen) [reply src/lib/replication/repl_dev/raft_repl_dev.cpp]: agree not to do unlink atm.  unlink volatile lsn from SM doesnt help us., we can avoid putting error lsn as non-volatile into SM easily by checking rreq state `repl_req_state_t::ERRORED`.  Although we will not reaching the case before we have better disk write failure  handing

Comment (Besroy) [reply src/lib/replication/repl_dev/raft_repl_dev.cpp]: Okay, added check for rreq state before convert lsn to non-volatile.

**Review by xiaoxichen — DISMISSED** [review]
lgtm

File: src/lib/replication/repl_dev/raft_repl_dev.cpp [hunk 4]
```diff
@@ -594,7 +594,8 @@ folly::Future< folly::Unit > RaftReplDev::notify_after_data_written(std::vector<
     });
 }
 
-bool RaftReplDev::wait_for_data_receive(std::vector< repl_req_ptr_t > const& rreqs, uint64_t timeout_ms) {
+bool RaftReplDev::wait_for_data_receive(std::vector< repl_req_ptr_t > const& rreqs, uint64_t timeout_ms,
```
JacksonYao287, line 597: 1 we have to places to call `wait_for_data_receive`, in `handle_raft_event` and  `end_of_append_batch` . now that we have guaranteed the data is received in handle_raft_event, can we remove wait_for_data_receive in notify_after_data_written(end_of_append_batch)? since there is no timeout_rreqs in that case.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -621,14 +622,23 @@ bool RaftReplDev::wait_for_data_receive(std::vector< repl_req_ptr_t > const& rre
...
     } else {
-        m_repl_svc.add_to_fetch_queue(shared_from_this(), std::move(only_wait_reqs));
+        m_repl_svc.add_to_fetch_queue(shared_from_this(), only_wait_reqs);
     }
 
     // block waiting here until all the futs are ready (data channel filled in and promises are made);
-    auto all_futs = folly::collectAllUnsafe(futs).wait(std::chrono::milliseconds(timeout_ms));
-    return (all_futs.isReady());
+    auto all_futs_ready = folly::collectAllUnsafe(futs).wait(std::chrono::milliseconds(timeout_ms)).isReady();
+    if (!all_futs_ready && timeout_rreqs != nullptr) {
```
JacksonYao287, line 632: if we do not call handle_error for any timeout rreq, and just return a false which indicate not appending the log, so all the rreq will be the same as I mentioned before , right?  so we can keep them in m_repl_key_req_map even if they have data time_out .

in this case, any rreq will never be erased from m_repl_key_req_map because of the time_out data.

src/lib/replication/repl_dev/raft_repl_dev.cpp
xiaoxichen: we can remove this based on no pack/unpack.

xiaoxichen:
a general comment from @yamingk is if we can do 
```
bool rreq::is_volatile() {
     if (rreq-> lsn >= raft_server()->get_last_log_idx() ) return false;
     return true
```
so that we dont need to touch the lsn

JacksonYao287 (approved):

xiaoxichen:
> a general comment from @yamingk is if we can do
> 
> ```
> bool rreq::is_volatile() {
>      if (rreq-> lsn >= raft_server()->get_last_log_idx() ) return false;
>      return true;
> }
> ```
> 
> so that we dont need to touch the lsn.
> 
> I think it makes sense, @Besroy what do you think?
`
@Besroy could you explain the thinking you put a atomic to store is_volatile instead of relying on `get_last_log_idx`?

Besroy:
> > a general comment from @yamingk is if we can do
> > ```
> > bool rreq::is_volatile() {
> >      if (rreq-> lsn >= raft_server()->get_last_log_idx() ) return false;
> >      return true;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > so that we dont need to touch the lsn.
> > I think it makes sense, @Besroy what do you think?
> > `@Besroy could you explain the thinking you put a atomic to store is_volatile instead of relying on`get_last_log_idx`?

@xiaoxichen The reason for using an atomic variable instead of `get_last_log_idx` is based on implementation considerations: the `req ctx` is defined in `include/replication/repl_dev.h`, whereas the `raft_server` is defined in `lib/replication/raft_repl_dev.h`. If we were to use `raft_server`, we would need to employ some tricky methods to make `repl_dev` aware of the elements in `raft_repl_dev`, which doesn't seem very elegant. Therefore, I chose not to use the raft server. Instead, I use an atomic boolean variable to indicate the state, which keeps the `req ctx` cleaner.

xiaoxichen (approved):

PR #605, opened 2024-12-11
disable restart for destroy-pending repl-dev

JacksonYao287:
when we try to destroy a repl_dev , we will first mark it to destroy-pending state,  and then a background gc thread will try to find it periodically and permanently destroy it. however,  if crash happens before it is permanently destroyed, then there will be some issue left.

1 a destroy-pending repl-dev will not be put into `m_rd_map` , so `raft_group_config_found` will return a nullptr for this repl_dev, and thus repl_dev->restart will cause a nullpointer fault(**fixed**).

2 when permanently destroy a repl_dev we will
```
    m_rd_sb.destroy();
    m_raft_config_sb.destroy();
    m_data_journal->remove_store();
    logstore_service().destroy_log_dev(m_data_journal->logdev_id());
```

if crash happens after `m_rd_sb.destroy()`, but before `m_data_journal->remove_store()` , we will have no chance to reclaim log related resource for this repl_dev.

this pr checks and reclaims the resource when start and destory repl_dev superblk only after all the related resource are reclaimed

### DIFF
--- a/src/lib/replication/service/raft_repl_service.cpp
+++ b/src/lib/replication/service/raft_repl_service.cpp
@@ -384,7 +384,26 @@ void RaftReplService::load_repl_dev(sisl::byte_view const& buf, void* meta_cooki
     }
 
     if (rd_sb->destroy_pending == 0x1) {
-        LOGINFOMOD(replication, "ReplDev group_id={} was destroyed, skipping the load", group_id);
+        LOGINFOMOD(replication, "ReplDev group_id={} was destroyed, reclaim the stale resource", group_id);
+        // if we do not add the repl_dev to m_rd_map, it will not be permanently destroyed since gc thread finds the
+        // pending destroy repl_dev only from m_rd_map. so, we should try to reclaim all the repl_dev stale resources
+        // here.
+
+        // 1 since we permanantly destroy the repl_dev here, it will not join_raft group where raft_server will be
+        // created. hence , no need to detroy it through nuraft_mesg, where raft_server will be shutdown.
+        // 2  m_raft_config_sb will be destroyed in raft_group_config_found() method if repl_dev is is not found, so
+        // skip it.
+
+        // 3 try to destroy logstore related resources.
+        logstore_id_t logstore_id = rd_sb->logstore_id;
+        logdev_id_t logdev_id = rd_sb->logdev_id;
+        LOGINFOMOD(replication, "remove logstore {} and logdev {} for group {}", logstore_id, logdev_id, group_id);
+        logstore_service().remove_log_store(logdev_id, logstore_id);
+        logstore_service().destroy_log_dev(logdev_id);
+
+        // 4 destroy the superblk, and after this,  the repl_dev will not be loaded and found again.
+        rd_sb.destroy();
+
         return;
     }
 
--- a/src/lib/replication/service/raft_repl_service.cpp
+++ b/src/lib/replication/service/raft_repl_service.cpp
@@ -128,7 +128,8 @@ void RaftReplService::start() {
     // We need to first load the repl_dev with its config and then attach the raft config to that repl dev.
     for (auto const& [buf, mblk] : m_config_sb_bufs) {
         auto rdev = raft_group_config_found(buf, voidptr_cast(mblk));
-        rdev->on_restart();
+        // if repl_dev is in destroy_pending state, it will not be loaded.
+        if (rdev) rdev->on_restart();
     }
     m_config_sb_bufs.clear();
     LOGINFO("Repl devs load completed, calling upper layer on_repl_devs_init_completed");
--- a/src/lib/replication/service/raft_repl_service.cpp
+++ b/src/lib/replication/service/raft_repl_service.cpp
@@ -394,12 +394,8 @@ void RaftReplService::load_repl_dev(sisl::byte_view const& buf, void* meta_cooki
         // 2  m_raft_config_sb will be destroyed in raft_group_config_found() method if repl_dev is is not found, so
         // skip it.
 
-        // 3 try to destroy logstore related resources.
-        logstore_id_t logstore_id = rd_sb->logstore_id;
-        logdev_id_t logdev_id = rd_sb->logdev_id;
-        LOGINFOMOD(replication, "remove logstore {} and logdev {} for group {}", logstore_id, logdev_id, group_id);
-        logstore_service().remove_log_store(logdev_id, logstore_id);
-        logstore_service().destroy_log_dev(logdev_id);
+        // 3 logdev will be destroyed in delete_unopened_logdevs() if we don't open it(create repl_dev) here, so skip
+        // it.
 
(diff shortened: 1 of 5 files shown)


src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -383,7 +384,26 @@ void RaftReplService::load_repl_dev(sisl::byte_view const& buf, void* meta_cooki
     }
 
     if (rd_sb->destroy_pending == 0x1) {
-        LOGINFOMOD(replication, "ReplDev group_id={} was destroyed, skipping the load", group_id);
+        LOGINFOMOD(replication, "ReplDev group_id={} was destroyed, reclaim the stale resource", group_id);
+        // if we do not add the repl_dev to m_rd_map, it will not be permanently destroyed since gc thread finds the
+        // pending destroy repl_dev only from m_rd_map. so, we should try to reclaim all the repl_dev stale resources
```
xiaoxichen, line 389: can we call `permanent_destroy` instead of duplicating here?

xiaoxichen (commented):
lgtm in general.

src/lib/replication/service/raft_repl_service.cpp
JacksonYao287: permanent_destroy is a member function a class RaftReplDev, in which the following two line is class specified

```
    logstore_service().destroy_log_dev(m_data_journal->logdev_id());
    m_stage.update([](auto* stage) { *stage = repl_dev_stage_t::PERMANENT_DESTROYED; });
```

here, we do not have a repl dev instance, so we can not call permanent_destroy directly

xiaoxichen: move this block down to L412 after rdev constructed?  There is no much cost in RaftReplDev constructor.

JacksonYao287: there are some corner cases. for example, 
```
void RaftReplDev::permanent_destroy() {
    RD_LOGI("Permanent destroy for raft repl dev group_id={}", group_id_str());
    m_raft_config_sb.destroy();
    m_data_journal->remove_store();
    logstore_service().destroy_log_dev(m_data_journal->logdev_id());
    m_stage.update([](auto* stage) { *stage = repl_dev_stage_t::PERMANENT_DESTROYED; });
    // we should destroy repl_dev superblk only after all the resources are cleaned up, so that is crash recovery
    // occurs, we have a chance to find the stale repl_dev and reclaim all the stale resources.
    m_rd_sb.destroy();
}
```

if we crash after `m_data_journal->remove_store()` but before  `logstore_service().destroy_log_dev(m_data_journal->logdev_id());`, then log store is removed and ReplLogStore can not be created, [this line](https://github.com/eBay/HomeStore/blob/fba9034a2e212948970dedf617bb8b86ddf116de/src/lib/replication/repl_dev/raft_repl_dev.cpp#L40-L41) is expected to fail. as a result, the repl_dev can not be created since the log store of it is lost

xiaoxichen (commented):
also one thing i am not clear ,   
```
void LogDev::handle_unopened_log_stores(bool format) {
```
this seems trying to similar target to GC the leaked logstores, it doesnt take care the logdev but it is not hard to add that when a logdev has zero open log store we GC the logdev? 

I think better to have one clear solution,  if we believe we have handled the GC here,  the `handle_unopened_log_stores` should be change to `ASSERT` or `LOG_ERROR`

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -383,7 +384,26 @@ void RaftReplService::load_repl_dev(sisl::byte_view const& buf, void* meta_cooki
...
+        // 1 since we permanantly destroy the repl_dev here, it will not join_raft group where raft_server will be
+        // created. hence , no need to detroy it through nuraft_mesg, where raft_server will be shutdown.
+        // 2  m_raft_config_sb will be destroyed in raft_group_config_found() method if repl_dev is is not found, so
+        // skip it.
+
+        // 3 try to destroy logstore related resources.
+        logstore_id_t logstore_id = rd_sb->logstore_id;
+        logdev_id_t logdev_id = rd_sb->logdev_id;
+        LOGINFOMOD(replication, "remove logstore {} and logdev {} for group {}", logstore_id, logdev_id, group_id);
+        logstore_service().remove_log_store(logdev_id, logstore_id);
```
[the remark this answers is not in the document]
JacksonYao287: yes, in the current master, it will. but int this PR, it has been changed to a warn
https://github.com/eBay/HomeStore/blob/fba9034a2e212948970dedf617bb8b86ddf116de/src/lib/logstore/log_dev.cpp#L641

xiaoxichen (commented):
lgtm

feel free to merge if want to create a ticket and work on enhancement later

xiaoxichen (dismissed):

JacksonYao287:
> I think better to have one clear solution

this is a good question. after go through the code, the logic is as follows.

**1 destroy log store**
for a particular log dev , a log store will be put into pending gc state(in `m_garbage_store_ids`) in two cases:
1 if not been opened before logdev starting
2 be removed(remove_log_store) at runtime.

when log truncation happens , if all the log entries of this log store is truncated, it will be permanently destroyed.

**2 destroy logdev** 
it will release all the chunks of this logdev, and delete the metablk of it. and as a result , all the log store of this logdev is also detroyed , since all the logstore id of this logdev in the the metablk of this logdev.

coming to repl_dev case, we have only one logstore for one logdev, so we can skip destroying logstore, we can destroy logdev directly.

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -383,7 +384,22 @@ void RaftReplService::load_repl_dev(sisl::byte_view const& buf, void* meta_cooki
...
+        // pending destroy repl_dev only from m_rd_map. so, we should try to reclaim all the repl_dev stale resources
+        // here.
+
+        // 1 since we permanantly destroy the repl_dev here, it will not join_raft group where raft_server will be
+        // created. hence , no need to detroy it through nuraft_mesg, where raft_server will be shutdown.
+        // 2  m_raft_config_sb will be destroyed in raft_group_config_found() method if repl_dev is is not found, so
+        // skip it.
+
+        // 3 logdev will be destroyed in delete_unopened_logdevs() if we don't open it(create repl_dev) here, so skip
+        // it.
```
JacksonYao287, line 398: we need do nothing here,  since if we do not create the repl_dev , [the related log_dev will not be opened](https://github.com/eBay/HomeStore/blob/4933dfb18abf6e0490e87ff280d8e80b72856d1d/src/lib/replication/log_store/home_raft_log_store.cpp#L100), as a result , the log dev will be eventually detroyed at 

https://github.com/eBay/HomeStore/blob/4933dfb18abf6e0490e87ff280d8e80b72856d1d/src/lib/replication/service/raft_repl_service.cpp#L176

xiaoxichen (approved):

PR #606, opened 2024-12-11
Calling pre_commit for lsn > dc_lsn.

xiaoxichen:
driven by nuraft implementation

https://github.com/eBay/NuRaft/blob/1adcc6282109c2ddf1121bbc374d48d303145e39/src/handle_append_entries.cxx#L847-L852

the pre_commit is called once log appened to log store, even before persist.

For logs recovered from log store, it should call pre-commit with no condition.

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -1261,12 +1261,12 @@ std::pair< bool, nuraft::cb_func::ReturnCode > RaftReplDev::handle_raft_event(nu
                 auto req = m_state_machine->localize_journal_entry_prepare(*entry);
                 if (req == nullptr) {
                     sisl::VectorPool< repl_req_ptr_t >::free(reqs);
-		    // The hint set here will be used by the next after next appendEntry, the next one
-		    // always go with -1 from NuRraft code.
-		    //
+                    // The hint set here will be used by the next after next appendEntry, the next one
+                    // always go with -1 from NuRraft code.
+                    //
                     // We are rejecting this log entry, meaning we can accept previous log entries.
-		    // If there is nothing we can accept(i==0), that maens we are waiting for commit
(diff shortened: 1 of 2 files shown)


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1485,16 +1485,20 @@ void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
     rreq->add_state(repl_req_state_t::LOG_FLUSHED);
     RD_LOGD("Replay log on restart, rreq=[{}]", rreq->to_string());
 
+    // 2. Pre-commit the log entry as in nuraft pre-commit was called once log appended to logstore.
+    m_listener->on_pre_commit(rreq->lsn(), rreq->header(), rreq->key(), rreq);
```
JacksonYao287, line 1489: any issue caused by this? 
for log entries with repl_lsn <= m_rd_sb->durable_commit_lsn, which means they are already definietly commited, why should we call on_precommit for them?

> Pre-commit the log entry as in nuraft pre-commit was called once log appended to logstore.

in nuraft, `on_commit` or `on_rollback` is expected after `on_precommit`. if we only call `on_precommit` for log entries with repl_lsn <= m_rd_sb->durable_commit_lsn, no later `on_commit` will be called for these log entries, not sure this is safe

xiaoxichen: why you think there is no on_commit() be called?

JacksonYao287: see line 1494 ，if repl_lsn > m_rd_sb->durable_commit_lsn , it will return and not call handle _commit

JacksonYao287: sorry , you are right, i misread line 1494

JacksonYao287 (approved):
LG

PR #608, opened 2024-12-13
register logstore callbacks in on_log_store_found

JacksonYao287:


xiaoxichen:
lode lgtm, however i hope the commit message contains enough background regarding why we want to do this

JacksonYao287:
> lode lgtm, however i hope the commit message contains enough background regarding why we want to do this

ok , will force push with a more meaningful commit message

xiaoxichen (approved):

PR #609, opened 2024-12-13
Fix Index recovery path for split(put)

shosseinimotlagh:
1- fix old root change when new root buffer is being discarded and old root is dirtied
    2- fix the wait down when the down_wait of a node for children is 0 and up buffer of the should not wait for this node to be repairedase enter the commit message for your changes. Lines starting


koujl (approved):
lgtm

PR #611, opened 2024-12-17
Duplication Handling

yuwmao:
This commit addresses duplication issues on the follower side caused by resync from the leader, it mainly happens when resend snapshot mesg during baseline resync and apply log after snapshot completion. This helps avoid unnecessary GC due to duplicated data.

Key Changes:
- Utilize allocation hints to check data existence via the application listener.
- Introduce `committed_blk_id` in `blk_alloc_hints` to indicate already allocated and committed blocks and pass it from application to HS, preventing reallocation and recommitment.
- In `alloc_local_blks()`, if `committed_blk_id` is returned, also add states `DATA_RECEIVED`, `DATA_WRITTEN`, and `DATA_COMMITTED` to skip async_write() and commit_blk().

On the leader side (`RaftReplDev::async_alloc_write`), duplication is treated as an error, as the leader should not propose duplicate data, which may result from mistakes.

FYI, to make the idea clearer, here is the adaptive changes in HO https://github.com/eBay/HomeObject/pull/244/files.

### DIFF
--- a/src/lib/replication/repl_dev/common.cpp
+++ b/src/lib/replication/repl_dev/common.cpp
@@ -6,11 +6,12 @@
 #include <common/homestore_config.hpp>
 #include "replication/repl_dev/common.h"
 #include <libnuraft/nuraft.hxx>
+#include <iomgr/iomgr_flip.hpp>
 
 namespace homestore {
 
-void repl_req_ctx::init(repl_key rkey, journal_type_t op_code, bool is_proposer, sisl::blob const& user_header,
-                        sisl::blob const& key, uint32_t data_size) {
+ReplServiceError repl_req_ctx::init(repl_key rkey, journal_type_t op_code, bool is_proposer, sisl::blob const& user_header,
+                        sisl::blob const& key, uint32_t data_size, cshared< ReplDevListener >& listener) {
     m_rkey = std::move(rkey);
 #ifndef NDEBUG
     if (data_size > 0) {
@@ -24,6 +25,14 @@ void repl_req_ctx::init(repl_key rkey, journal_type_t op_code, bool is_proposer,
     m_header = user_header;
     m_key = key;
     m_is_jentry_localize_pending = (!is_proposer && (data_size > 0)); // Pending on the applier and with linked data
+    if (has_linked_data()) {
+        auto alloc_status = alloc_local_blks(listener, data_size);
+        if (alloc_status != ReplServiceError::OK) {
+            LOGERROR("Allocate blk for rreq failed error={}", alloc_status);
+        }
+        return alloc_status;
+    }
+    return ReplServiceError::OK;
 }
 
 repl_req_ctx::~repl_req_ctx() {
@@ -93,6 +102,7 @@ ReplServiceError repl_req_ctx::alloc_local_blks(cshared< ReplDevListener >& list
 
     if (hints_result.value().committed_blk_id.has_value()) {
         //if the allocted_blk_id is already present, use it and skip allocation
+        LOGINFO("For Repl_key=[{}] data already exists, skip", rkey().to_string());
         m_local_blkid = hints_result.value().committed_blk_id.value();
         add_state(repl_req_state_t::BLK_ALLOCATED);
         add_state(repl_req_state_t::DATA_RECEIVED);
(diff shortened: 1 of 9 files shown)


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -501,29 +504,23 @@ repl_req_ptr_t RaftReplDev::applier_create_req(repl_key const& rkey, journal_typ
     // We need to allocate the block, since entry doesn't exist or if it exist, two threads are trying to do the same
     // thing. So take state mutex and allocate the blk
     std::unique_lock< std::mutex > lg(rreq->m_state_mtx);
-    rreq->init(rkey, code, false /* is_proposer */, user_header, key, data_size);
-
-    // There is no data portion, so there is not need to allocate
+    auto status = rreq->init(rkey, code, false /* is_proposer */, user_header, key, data_size, m_listener);
     if (!rreq->has_linked_data()) { return rreq; }
-    if (rreq->has_state(repl_req_state_t::BLK_ALLOCATED)) { return rreq; }
```
xiaoxichen, line 508: this line is trying to skip the allocation if it is already allocated, for example , data channel comes first, the `happened == true`, however raft channel comes in with `happened = false` but  finished the allocation, then data channel should skip the allocation.

similar logic is needed for rreq->init as you move the allocation into that.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -292,25 +292,28 @@ void RaftReplDev::async_alloc_write(sisl::blob const& header, sisl::blob const&
...
-        push_data_to_all_followers(rreq, data);
-
-        // Step 1: Alloc Blkid
-        auto const status = rreq->alloc_local_blks(m_listener, data.size);
-        if (status != ReplServiceError::OK) {
-            RD_LOGD("Allocating blks failed error={}, failing this req", status);
-            handle_error(rreq, status);
+        if (rreq->is_proposer() && rreq->has_state(repl_req_state_t::DATA_COMMITTED)) {
+            RD_LOGD("data blks has already been allocated and committed, failing this req");
+            handle_error(rreq, ReplServiceError::DATA_DUPLICATED);
```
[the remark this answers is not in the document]
yuwmao: No concrete case found currently, just a safety check, to avoid duplication happening on the leader size, e.g. bsn is not managed correctly in HO.

src/lib/replication/repl_dev/raft_repl_dev.cpp
yuwmao: Yes, should add this logic in rreq->init.

src/lib/replication/repl_dev/common.cpp
```diff
@@ -91,6 +100,19 @@ ReplServiceError repl_req_ctx::alloc_local_blks(cshared< ReplDevListener >& list
...
+        //if the allocted_blk_id is already present, use it and skip allocation
+        LOGINFO("For Repl_key=[{}] data already exists, skip", rkey().to_string());
+        m_local_blkid = hints_result.value().committed_blk_id.value();
+        add_state(repl_req_state_t::BLK_ALLOCATED);
+        add_state(repl_req_state_t::DATA_RECEIVED);
+        add_state(repl_req_state_t::DATA_WRITTEN);
+        add_state(repl_req_state_t::DATA_COMMITTED);
+        m_data_received_promise.setValue();
+        m_data_written_promise.setValue();
+        return ReplServiceError::OK;
```
JacksonYao287, line 113: if we return here, we will not set m_local_blkid in line 117, it is very critical for log entry.

for every member(leader or follower), if a log entry is data_linked, then the log entry will have a member of local_blk_id to indicate the data associated with this log entry.

when leader send log entry to follower, the local_blk_id will be considered as remote_blk_id in the follower side , used for fetch data.  when follower append log to local log store, it will change the local_blk_id to its own local_blk_id, so that when this follower becomes leader, it can send this log to other follower , and other follower can fetch data from this new leader according to the local_blk_id in this log entry.

so here, we need to change m_local_blkid to the commited_blk_id

yuwmao: Yes, I have already set `m_local_blkid` to the `commited_blk_id` at line 106.

xiaoxichen:
@yuwmao pls rebase

JacksonYao287 (dismissed):
LGTM

xiaoxichen (approved):

PR #612, opened 2024-12-17
Add lock for log dev read api's.

sanebay:
Avoid concurrent access to journal vdev for truncate, write and read api's from different threads.


JacksonYao287 (approved):
LGTM. let`s revisit here if any performance issue surfaces in performance test

PR #614, opened 2024-12-18
fix twice call of leave

JacksonYao287:
Removed From cluster will be called twice when committing config_change log and journal_type_t::HS_CTRL_DESTROY in the moved out member, so the destroy future will be setvalue twice , which will lead to an error. this PR fixes this bug

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -1183,7 +1183,7 @@ nuraft::ptr< nuraft::log_store > RaftReplDev::load_log_store() { return m_data_j
 
 int32_t RaftReplDev::server_id() { return m_raft_server_id; }
 
-bool RaftReplDev::is_destroy_pending() const { return (m_rd_sb->destroy_pending == 0x1); }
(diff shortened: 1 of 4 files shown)


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1207,6 +1207,19 @@ void RaftReplDev::permanent_destroy() {
...
+    //  2. it is removed from the cluster and the new config(excluding this node) is being committed on this node
+    //  3. it is removed from the cluster , but the node is down and new config log(excluding this node) is not
+    //  replicated to this removed node. when the node restart, leader will not send any append entry to this node,
+    //  since it is not a member of the raft group. it will become a condidate and send request-vote request to other
+    //  members of this raft group. a member will send RemovedFromCluster to the node if this member finds the node
+    //  is no longer a member of the raft group.
+
+    // leave() will never be called concurrently, since config change and journal_type_t::HS_CTRL_DESTROY are all log
+    // entry, which will be committed sequentially.
+    if (is_destroy_pending()) return;
```
JacksonYao287, line 1221: this avoids line 1237 be called twice

xiaoxichen (approved):
lgtm

PR #615, opened 2024-12-19
Refine variant_node::put return value

koujl:
Replace the generic put_failed error status with more specific error messages to enhance clarity and prevent confusion during log analysis.

### DIFF
--- a/src/include/homestore/btree/detail/btree_mutate_impl.ipp
+++ b/src/include/homestore/btree/detail/btree_mutate_impl.ipp
@@ -56,7 +56,7 @@ retry:
         const auto matched = my_node->match_range(req.working_range(), start_idx, end_idx);
         if (!matched) {
             BT_NODE_LOG_ASSERT(false, my_node, "match_range returns 0 entries for interior node is not valid pattern");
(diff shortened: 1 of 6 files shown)


xiaoxichen (approved):

src/include/homestore/btree/detail/btree_mutate_impl.ipp
```diff
@@ -56,7 +56,7 @@ retry:
         const auto matched = my_node->match_range(req.working_range(), start_idx, end_idx);
         if (!matched) {
             BT_NODE_LOG_ASSERT(false, my_node, "match_range returns 0 entries for interior node is not valid pattern");
-            ret = btree_status_t::put_failed;
+            ret = btree_status_t::not_found;
```
shosseinimotlagh, line 59: @xiaoxichen It is unclear to me. It needs some explanation for the different cases. A caller wants to insert but get into a range that may be wrong, what does not_found mean? it is not supposed not to be found.

xiaoxichen: What do you think?  in which case we will hit into this?

PR #616, opened 2024-12-19
Only call cp_flush for those consumer paticipated in this cp.

xiaoxichen:
If a consumer registered after a cp goes to flushing state, the on_switchover_cp cb will not be called for this consumer. In this CP, the ctx for this consumer is nullptr as the consumer never participant in the cp.

Previous code calling cp_flush for every consumer, leaving the duty of properly handle the nullptr returned by cp->context(svc_id) to consumer. However, none of the existing consumer handled the case.

As a result, we hit an occurance that Index generate a CP sololy, but before the cp fully flushed, other consumer registered and be called into cp_flush(), the replication service, doesnt properly handled the nullptr like below, `get_repl_dev_ctx` was called with this_ptr is null, it is dangerous as invalid memory get accessed.

This change is a breaking change for consumer like HO so bump up the version. HomeObject participant the CP as CLIENT, current implementation of HO always returns nullptr for `on_switchover_cp` which will result the CLIENT be excluded from cp_flush after this commit merged.

callstack:
...

code:

```
 auto cp_ctx = s_cast< ReplSvcCPContext* >(cp->context(cp_consumer_t::REPLICATION_SVC));
 ...
 auto dev_ctx = cp_ctx->get_repl_dev_ctx(repl_dev.get());
```

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.6.0"
+    version = "6.6.1"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/lib/checkpoint/cp_mgr.cpp
+++ b/src/lib/checkpoint/cp_mgr.cpp
@@ -187,7 +187,8 @@ folly::Future< bool > CPManager::do_trigger_cp_flush(bool force, bool flush_on_s
     // sealer should be the first one to switch over
     auto& sealer_cp = m_cp_cb_table[(size_t)cp_consumer_t::SEALER];
     if (sealer_cp) {
-        new_cp->m_contexts[(size_t)cp_consumer_t::SEALER] = std::move(sealer_cp->on_switchover_cp(cur_cp.get(), new_cp));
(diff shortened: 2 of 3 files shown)


JacksonYao287:
only index service(wb_cache) register_consumer at constructor, so can we change [this line](https://github.com/eBay/HomeStore/blob/6695df205d1b579f76a07734a9bc5c6eec2d86c4/src/lib/index/index_service.cpp#L92)  to 
```
hs()->cp_mgr().trigger_cp_flush(true /* force */).wait();

```

this will trigger a sync cp_flush. at that time, no other cp consumer is registered.

xiaoxichen:
> only index service(wb_cache) register_consumer at constructor, so can we change [this line](https://github.com/eBay/HomeStore/blob/6695df205d1b579f76a07734a9bc5c6eec2d86c4/src/lib/index/index_service.cpp#L92) to
> 
> ```
> hs()->cp_mgr().trigger_cp_flush(true /* force */).wait();
> ```
> 
> this will trigger a sync cp_flush. at that time, no other cp consumer is registered.

This line will be changed with other PR.

JacksonYao287 (approved):
the issue here is that a consumer register

JacksonYao287 (approved):

PR #617, opened 2024-12-20
read_logical_snp_obj: pass user_ctx to prevent memleak

koujl:
Ensure user_ctx is passed up to NuRaft regardless of the return value to enable the cleanup of the allocated context object.


yuwmao (approved):
LGTM

PR #619, opened 2024-12-23
fix HomeRaftLogStore::last_entry

JacksonYao287:
according to the contract, we should return a dummy log entry if the index is out of range.

### DIFF
--- a/src/include/homestore/logstore/log_store.hpp
+++ b/src/include/homestore/logstore/log_store.hpp
@@ -231,6 +231,7 @@ public:
     bool rollback(logstore_seq_num_t to_lsn);
 
     auto start_lsn() const { return m_start_lsn.load(std::memory_order_acquire); }
+    auto tail_lsn() const { return m_tail_lsn.load(std::memory_order_acquire); }
 
     nlohmann::json dump_log_store(const log_dump_req& dump_req = log_dump_req());
 
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -39,7 +39,10 @@ RaftReplDev::RaftReplDev(RaftReplService& svc, superblk< raft_repl_dev_superblk 
         m_data_journal = std::make_shared< ReplLogStore >(
             *this, *m_state_machine, m_rd_sb->logdev_id, m_rd_sb->logstore_id,
             [this](logstore_seq_num_t lsn, log_buffer buf, void* key) { on_log_found(lsn, buf, key); },
-            [this](std::shared_ptr< HomeLogStore > hs, logstore_seq_num_t lsn) { m_log_store_replay_done = true; });
+            [this](std::shared_ptr< HomeLogStore > hs, logstore_seq_num_t lsn) {
+                m_log_store_replay_done = true;
+                set_log_store_last_durable_lsn(hs->tail_lsn());
+            });
         m_next_dsn = m_rd_sb->last_applied_dsn + 1;
         m_commit_upto_lsn = m_rd_sb->durable_commit_lsn;
         m_last_flushed_commit_lsn = m_commit_upto_lsn;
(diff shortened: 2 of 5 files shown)


Besroy:
lgtm

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -39,7 +39,10 @@ RaftReplDev::RaftReplDev(RaftReplService& svc, superblk< raft_repl_dev_superblk
         m_data_journal = std::make_shared< ReplLogStore >(
             *this, *m_state_machine, m_rd_sb->logdev_id, m_rd_sb->logstore_id,
             [this](logstore_seq_num_t lsn, log_buffer buf, void* key) { on_log_found(lsn, buf, key); },
-            [this](std::shared_ptr< HomeLogStore > hs, logstore_seq_num_t lsn) { m_log_store_replay_done = true; });
+            [this](std::shared_ptr< HomeLogStore > hs, logstore_seq_num_t lsn) {
+                m_log_store_replay_done = true;
+                set_last_durable_lsn(hs->tail_lsn());
```
JacksonYao287, line 44: apart from applying snapshot, last_durable_lsn is only updated when log is flushed (end_of_log_batch). so if we do not schedule any log flush and try to get the last_durable_lsn, then we will get -1, which is not intended.

so, we need to set it to the tail_lsn after all the log entry is replayed.

src/include/homestore/logstore/log_store.hpp
```diff
@@ -231,6 +231,7 @@ class HomeLogStore : public std::enable_shared_from_this< HomeLogStore > {
     bool rollback(logstore_seq_num_t to_lsn);
 
     auto start_lsn() const { return m_start_lsn.load(std::memory_order_acquire); }
+    auto tail_lsn() const { return m_tail_lsn.load(std::memory_order_acquire); }
```
xiaoxichen, line 234: so this line is the actual fix？
thedummy log is a right fix but we don't expect to hit?

Besroy: Yes, this line and` set_last_durable_lsn(hs->tail_lsn());` is the actual fix to avoid hitting 'dummy log at last_entry()'

JacksonYao287: 1 according the [nuraft logstore#last_entry()](https://github.com/eBay/NuRaft/blob/50e2f949503081262cb21923e633eaa8dacad8fa/include/libnuraft/log_store.hxx#L56), a dummy log should return If no log entry exists, and should not throw an exception. 

2 set_last_durable_lsn(hs->tail_lsn() makes sure HomeRaftLogStore#m_last_durable_lsn is set to up-to-date we restart. now , m_last_durable_lsn is only updated when log is flushed (end_of_log_batch), which means if we do not append any log after restart , the m_last_durable_lsn will be a default value(-1), which might cause a issue.

in sm_long running issue 19, after restart, no log entry is appended. and this node starts handing vote, where ogstore#last_entry() will be called, but at this moment , m_last_durable_lsn is -1 , which shuold be the last lsn in log store(tail_lsn). this will cause an error that log entry not found

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -39,7 +39,10 @@ RaftReplDev::RaftReplDev(RaftReplService& svc, superblk< raft_repl_dev_superblk
         m_data_journal = std::make_shared< ReplLogStore >(
             *this, *m_state_machine, m_rd_sb->logdev_id, m_rd_sb->logstore_id,
             [this](logstore_seq_num_t lsn, log_buffer buf, void* key) { on_log_found(lsn, buf, key); },
-            [this](std::shared_ptr< HomeLogStore > hs, logstore_seq_num_t lsn) { m_log_store_replay_done = true; });
+            [this](std::shared_ptr< HomeLogStore > hs, logstore_seq_num_t lsn) {
+                m_log_store_replay_done = true;
+                set_log_store_last_durable_lsn(hs->tail_lsn());
```
JacksonYao287, line 44: this line is the actual fix , which will set m_last_durable_lsn to the tail_lsn(last lsn) of log store after all the logs are replayed. @xiaoxichen

xiaoxichen (approved):

PR #620, opened 2024-12-24
Remove log store truncation from resource mg

Besroy:
Currently both resource_mgr and raft can call log store's `truncate`, but resource_mgr will not truncate logs whose lsn less than compact lsn. That means resource_mgr just re-truncate logs which will be / has been truncated in `compact`. But if resource_mgr and raft call truncate concurrently, crash will happen. 
For example, raft truncate logs upto compact_lsn and execute 
```
m_records.truncate(upto_lsn);
m_start_lsn.store(upto_lsn + 1);
```
Then resource_mgr truncate truncate_lsn (which is no large than compact_lsn) and execute `m_trunc_ld_key = m_records.at(upto_lsn).m_trunc_key; ` Since truncate_lsn has been truncated, an exception thrown.

This pr remove truncation from resource_mgr to avoid concurrency.

### DIFF
--- a/src/lib/replication/log_store/home_raft_log_store.cpp
+++ b/src/lib/replication/log_store/home_raft_log_store.cpp
@@ -49,6 +49,9 @@ static uint64_t extract_term(const log_buffer& log_bytes) {
     return (*r_cast< uint64_t const* >(raw_ptr));
 }
 
+#if 0
+// Since truncate_lsn can not accross compact_lsn passed down by raft server
+// and compact will truncate logs upto compact_lsn, we don't need to re-truncate in this function now.
 void HomeRaftLogStore::truncate(uint32_t num_reserved_cnt, repl_lsn_t compact_lsn) {
     auto const last_lsn = last_index();
     auto const start_lsn = start_index();
@@ -79,6 +82,7 @@ void HomeRaftLogStore::truncate(uint32_t num_reserved_cnt, repl_lsn_t compact_ls
         m_log_store->truncate(truncate_lsn);
     }
 }
+#endif
 
 HomeRaftLogStore::HomeRaftLogStore(logdev_id_t logdev_id, logstore_id_t logstore_id, log_found_cb_t const& log_found_cb,
                                    log_replay_done_cb_t const& log_replay_done_cb) :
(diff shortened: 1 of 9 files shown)


src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -54,7 +54,9 @@ void HomeRaftLogStore::truncate(uint32_t num_reserved_cnt, repl_lsn_t compact_ls
     auto const start_lsn = start_index();
 
     // compact_lsn will be zero on first time boot, so we should not truncate in that case.
-    if (compact_lsn == 0 || (start_lsn + num_reserved_cnt >= last_lsn)) {
+    // if comapct_lsn > last_lsn, it means compact will/has truncate all logs, so we should not truncate in that case.
+    // FIXME: cannot resolve the concurrency of trucate and comapct (if not first time boot), maybe need lock in future.
+    if (compact_lsn == 0 || compact_lsn > s_cast< repl_lsn_t >(last_lsn) || (start_lsn + num_reserved_cnt >= last_lsn)) {
```
yuwmao, line 59: Can we just add condition `|| to_store_lsn(compact_lsn)<start_lsn` to simplify line 75~81?

Besroy: We might consider setting truncate_lsn as compact_lsn and adding this condition to simplify the code. However, I'm uncertain if this approach is feasible. In the current implementation, it seems we cannot truncate any log larger than compact_lsn, yet the compact operation will truncate logs. Do you know if there were any prior considerations regarding this?

Otherwise, lines 75-81 in home_raft_log_store are useful in the scenario where `last_lsn > compact_lsn > start_lsn` and` last_lsn - num_reserved_cnt < compact_lsn`.

JacksonYao287:
@yamingk @sanebay  this is another concurrent issue at logstore level, cause by two concurrent truncation(one is from nuraft, the other is from resource manager). do we indeed need truncation in resource manager? pls check if this will affect nublox case.

xiaoxichen (dismissed):
lgtm

xiaoxichen (approved):

PR #621, opened 2024-12-30
minor fix for homeobject's homestore_test

Besroy:


xiaoxichen (approved):
i am not sure this is a fix but as the `override` is needed merging this pr

Besroy:
> i am not sure this is a fix but as the `override` is needed merging this pr

@xiaoxichen Yes you are right, it is not useful for ho ut, the reason may be https://github.com/eBay/HomeStore/pull/622

PR #622, opened 2024-12-31
update last_commit_lsn in commit_config

Besroy:
HomeObject call HS `is_ready_for_traffic` before put/get/del, and if it return false, HO will reject requests. In HomeStore, we use `m_commit_upto_lsn>=m_traffic_ready_lsn` to determine if it is ready, so we need to update `m_commit_upto_lsn` every time raft commit a log. Currently we only update it in app log, so this pr add the update in config log.

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -959,6 +959,20 @@ void RaftReplDev::handle_commit(repl_req_ptr_t rreq, bool recovery) {
     if (!rreq->is_proposer()) { rreq->clear(); }
 }
 
+void RaftReplDev::handle_config_commit(const repl_lsn_t lsn, raft_cluster_config_ptr_t& new_conf) {
+    // when reaching here, the new config has already been applied to the cluster.
+    // since we didn't create repl req for config change, we just need to update m_commit_upto_lsn here.
+
+    // keep this variable in case it is needed later
+    (void) new_conf;
+    auto prev_lsn = m_commit_upto_lsn.load(std::memory_order_relaxed);
+    RD_DBG_ASSERT_GT(lsn, prev_lsn,
+                     "Out of order commit of lsns, it is not expected in RaftReplDev. cur_lsns={}, prev_lsns={}",
+                     lsn, prev_lsn);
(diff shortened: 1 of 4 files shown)


xiaoxichen (changes requested):
pls add a ` void RaftReplDev::handle_config_commit()` and update the commit_lsn via `compare_and_set`

xiaoxichen:
it is not for correctness as the commit of state machine is always sequential, but just to make the code style more align.

xiaoxichen (approved):

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -959,6 +959,20 @@ void RaftReplDev::handle_commit(repl_req_ptr_t rreq, bool recovery) {
...
+    // when reaching here, the new config has already been applied to the cluster.
+    // since we didn't create repl req for config change, we just need to update m_commit_upto_lsn here.
+
+    // keep this variable in case it is needed later
+    (void) new_conf;
+    auto prev_lsn = m_commit_upto_lsn.load(std::memory_order_relaxed);
+    RD_DBG_ASSERT_GT(lsn, prev_lsn,
+                     "Out of order commit of lsns, it is not expected in RaftReplDev. cur_lsns={}, prev_lsns={}",
+                     lsn, prev_lsn);
+    RD_DBG_ASSERT(m_commit_upto_lsn.compare_exchange_strong(prev_lsn, lsn),
```
JacksonYao287, line 972: not sure whether it will take effect if in release or releaseWithDebug mode.  will it only work when debug mode? pls check this.

if we want to guarantee `m_commit_upto_lsn.compare_exchange_strong(prev_lsn, lsn) ` is called, pls use a separate line for this code.

JacksonYao287 (commented):
LGTM , only a minor comment

src/lib/replication/repl_dev/raft_repl_dev.cpp
xiaoxichen: it will be execute as normal
```
#ifndef NDEBUG
#define DEBUG_ASSERT(cond, msg, ...) RELEASE_ASSERT(cond, msg, ##__VA_ARGS__)
#else
#define DEBUG_ASSERT(cond, msg, ...)
```

PR #623, opened 2025-01-06
Support async snapshot io config

yuwmao:
Default snapshot read logic is in a sync manner which may block other response handling and make the heartbeat timeout  . As a result, the leader will yield its leadership after check_leadership_validity.


xiaoxichen (approved):
lgtm, this parameters only affecting snapshot read on leader side

PR #624, opened 2025-01-08
skip append dummy logs to log dev

Besroy:
During the baseline resynchronization process, when a follower successfully receives a snapshot, it calls the compact function to truncate all logs that are no larger than snapshot.last_log_idx. In the current implementation, compact initially appends dummy logs up to snapshot.last_log_idx before truncating them. This operation is resource-intensive, and the size of the data waiting to be flushed can exceed the journal vdev's chunk size, posing a risk of system crashes.

This change eliminates the need to append dummy logs to the log device by directly updating certain indices. It also removes the dependency of log dev truncation on log store's LSN.

### DIFF
--- a/src/lib/logstore/log_dev.cpp
+++ b/src/lib/logstore/log_dev.cpp
@@ -135,6 +135,7 @@ void LogDev::stop() {
     m_log_idx.store(0);
     m_pending_flush_size.store(0);
     m_last_flush_idx = -1;
+    m_last_flush_ld_key = logdev_key{0, 0};
     m_last_truncate_idx = -1;
     m_last_crc = INVALID_CRC32_VALUE;
 
@@ -501,6 +502,7 @@ void LogDev::on_flush_completion(LogGroup* lg) {
     free_log_group(lg);
     m_log_records->truncate(upto_indx);
     m_last_flush_idx = upto_indx;
+    m_last_flush_ld_key = logdev_key{from_indx, dev_offset};
 
     // since we support out-of-order lsn write, so no need to guarantee the order of logstore write completion
     for (auto const& [idx, req] : req_map) {
@@ -530,20 +532,18 @@ uint64_t LogDev::truncate() {
         auto lstore = store.log_store;
         if (lstore == nullptr) { continue; }
         auto const [trunc_lsn, trunc_ld_key, tail_lsn] = lstore->truncate_info();
-        if (trunc_lsn == tail_lsn) {
-            THIS_LOGDEV_LOG(DEBUG, "Store_id={} didn't have any writes since last truncation, skipping ", store_id);
-            m_logdev_meta.remove_all_rollback_records(store_id, m_stopped /* persist_now */);
-            continue;
-        }
-        HS_DBG_ASSERT_GE(trunc_ld_key.idx, m_last_truncate_idx, "Trying to truncate logid which is already truncated");
         m_logdev_meta.update_store_superblk(store_id, logstore_superblk(trunc_lsn + 1), m_stopped /* persist_now */);
-
         // We found a new minimum logdev_key that we can truncate to
-        if (trunc_ld_key.idx > 0 && trunc_ld_key.idx < min_safe_ld_key.idx) { min_safe_ld_key = trunc_ld_key; }
+        if (trunc_ld_key.idx < min_safe_ld_key.idx) { min_safe_ld_key = trunc_ld_key; }
+    }
+
+    // All log stores are empty, we can truncate logs depends on the last flushed logdev_key
+    if (min_safe_ld_key == logdev_key::out_of_bound_ld_key()) {
+        min_safe_ld_key = m_last_flush_ld_key;
     }
 
     // There are no writes or no truncation called for any of the store, so we can't truncate anything
-    if (min_safe_ld_key == logdev_key::out_of_bound_ld_key() || min_safe_ld_key.idx <= m_last_truncate_idx) return 0;
+    if (min_safe_ld_key.idx <= 0 || min_safe_ld_key.idx <= m_last_truncate_idx) return 0;
 
     uint64_t const num_records_to_truncate = uint64_cast(min_safe_ld_key.idx - m_last_truncate_idx);
 
--- a/src/lib/logstore/log_store.cpp
+++ b/src/lib/logstore/log_store.cpp
@@ -189,12 +189,27 @@ void HomeLogStore::truncate(logstore_seq_num_t upto_lsn, bool in_memory_truncate
 
 #endif
 
+    // In normal write and compact path, upto_lsn is expected to be no larger than m_tail_lsn after the flush.
+    // So upto_lsn > m_tail_lsn is expected to exist only in baseline resync path.
+    // In baseline resync path, we truncate all entries up to upto_lsn, and update m_tail_lsn and m_next_lsn
+    // to make sure logstore's idx is always = raft's idx - 1.
     if (upto_lsn > m_tail_lsn) {
         THIS_LOGSTORE_LOG(WARN,
-                          "Truncating issued on lsn={} which is greater than tail_lsn={}, truncating upto tail_lsn",
+                          "Truncating issued on lsn={} which is greater than tail_lsn={}",
                           upto_lsn, m_tail_lsn.load(std::memory_order_relaxed));
-        m_trunc_ld_key = m_records.at(m_tail_lsn).m_trunc_key;
-        upto_lsn = m_tail_lsn;
+        // update m_tail_lsn if it is less than upto_lsn
+        auto current_tail_lsn = m_tail_lsn.load(std::memory_order_relaxed);
+        while (current_tail_lsn < upto_lsn &&
+               !m_tail_lsn.compare_exchange_weak(current_tail_lsn, upto_lsn, std::memory_order_relaxed)) {}
+
+        // update m_next_lsn if it is less than upto_lsn + 1
+        auto current_next_lsn = m_next_lsn.load(std::memory_order_relaxed);
+        while (current_next_lsn < upto_lsn + 1 &&
+               !m_next_lsn.compare_exchange_weak(current_next_lsn, upto_lsn + 1, std::memory_order_relaxed)) {}
+
+        // insert an empty record to make sure m_records has enough size to truncate
+        logdev_key empty_ld_key;
+        m_records.create_and_complete(upto_lsn, logstore_record(empty_ld_key, empty_ld_key));
     } else {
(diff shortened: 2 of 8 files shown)


src/lib/logstore/log_store.cpp
```diff
@@ -189,17 +189,35 @@ void HomeLogStore::truncate(logstore_seq_num_t upto_lsn, bool in_memory_truncate
...
+    // So upto_lsn > m_tail_lsn is expected to exist only in baseline resync path.
+    // In baseline resync path, we truncate all entries up to upto_lsn, and set m_trunc_ld_key to the ld_key of
+    // m_tail_lsn so that log dev can truncate all log records.
+    // We also need to update m_tail_lsn and m_next_lsn to make sure logstore's idx is always = raft's idx - 1.
     if (upto_lsn > m_tail_lsn) {
         THIS_LOGSTORE_LOG(WARN,
-                          "Truncating issued on lsn={} which is greater than tail_lsn={}, truncating upto tail_lsn",
+                          "Truncating issued on lsn={} which is greater than tail_lsn={}",
                           upto_lsn, m_tail_lsn.load(std::memory_order_relaxed));
         m_trunc_ld_key = m_records.at(m_tail_lsn).m_trunc_key;
```
Besroy, line 201: @xiaoxichen I use the trunc key of tail_lsn rather than an empty key here to completely decouple the relationship between the log store and log dev. The trunc key corresponding to the last lsn seen by the log store is precisely where the log dev needs to truncate. I want this approach to avoid potential conflicts between the log store's logic for using empty or out-of-range keys and the logic of the log dev in the future.

src/lib/logstore/log_dev.cpp
```diff
@@ -530,9 +530,10 @@ uint64_t LogDev::truncate() {
         auto lstore = store.log_store;
         if (lstore == nullptr) { continue; }
         auto const [trunc_lsn, trunc_ld_key, tail_lsn] = lstore->truncate_info();
-        if (trunc_lsn == tail_lsn) {
-            THIS_LOGDEV_LOG(DEBUG, "Store_id={} didn't have any writes since last truncation, skipping ", store_id);
-            m_logdev_meta.remove_all_rollback_records(store_id, m_stopped /* persist_now */);
+        if (trunc_ld_key.idx <= m_last_truncate_idx) {
```
JacksonYao287, line 533: in what case will `trunc_ld_key.idx < m_last_truncate_idx`?
I think this will never happen. we can add an assert before this line. and change this line to `trunc_ld_key.idx == m_last_truncate_idx`

```
ASSERT(trunc_ld_key.idx >= m_last_truncate_idx)
```

src/lib/logstore/log_store.cpp
JacksonYao287: one question. if current tail_lsn of the logstore is 100.  
first we truncate it to 200, then we can , then here we use the m_trunc_ld_key of tail_lsn(lsn 100). it exists since it is a tail_lsn.  after this truncation , the tail_lsn will be set to 200.

now we truncate it again to 300,  if we use m_trunc_ld_key of tail_lsn(200) now, then it probably fail since actually tail_lsn(200) is an empty lsn at this moment

src/lib/logstore/log_store.cpp
```diff
@@ -189,17 +189,35 @@ void HomeLogStore::truncate(logstore_seq_num_t upto_lsn, bool in_memory_truncate
...
+        auto current_tail_lsn = m_tail_lsn.load(std::memory_order_relaxed);
+        while (current_tail_lsn < upto_lsn &&
+               !m_tail_lsn.compare_exchange_weak(current_tail_lsn, upto_lsn, std::memory_order_relaxed)) {}
+
+        // update m_next_lsn if it is less than upto_lsn + 1
+        auto current_next_lsn = m_next_lsn.load(std::memory_order_relaxed);
+        while (current_next_lsn < upto_lsn + 1 &&
+               !m_next_lsn.compare_exchange_weak(current_next_lsn, upto_lsn + 1, std::memory_order_relaxed)) {}
+
+        // insert an empty record to make sure m_records has enough size to truncate
```
JacksonYao287, line 213: NIT : use `logstore#fill_gap`

src/lib/logstore/log_store.cpp
Besroy: Do you mean SM receives 2 snapshots with different LSN? The first snapshot wants to compact upto 200 (but there are only 100 logs in sm), but it fails to complete after the m_tail_lsn is reset at 200, then the second snapshot is going to compact upto 300, since the trunc key is empty for LSN 200, logs 0-100 will not be truncated.
How about using `m_records.at(m_tail_lsn)`  to create and complete `m_records.at(upto_lsn)`?

src/lib/logstore/log_dev.cpp
Besroy: Will trunc_ld_key.idx < m_last_truncate_idx in this case:
T1: log dev truncate upto idx 100
T2: restart, set m_last_truncate_idx = 100 while loading logs
T3: trigger log dev truncate, in this time m_trunc_ld_key in log store is not set, so it is the default idx 0, 0 < 100

src/lib/logstore/log_store.cpp
Besroy: Thanks! It seems that fill_gap is used when there are holes NOT OUT OF RANGE in m_records, and this function will fill holes with invalid value, but I cannot find where the function is used (seems only used in test case). Can I update this function to permit set an out_of_range seq_num to enlarge m_records? Also I need to set a specific value. If not, maybe I need to write a new function

src/lib/logstore/log_dev.cpp
JacksonYao287: if do not miss anything, we never trigger logdev#truncate directly. we always trigger logstore#truncate , which will then trigger logdev#truncate.

when triggering logstore#truncate , m_trunc_ld_key will be definitely updated. see logstore#truncate.
```
    if (upto_lsn > m_tail_lsn) {
        THIS_LOGSTORE_LOG(WARN,
                          "Truncating issued on lsn={} which is greater than tail_lsn={}, truncating upto tail_lsn",
                          upto_lsn, m_tail_lsn.load(std::memory_order_relaxed));
        m_trunc_ld_key = m_records.at(m_tail_lsn).m_trunc_key;
        upto_lsn = m_tail_lsn;
    } else {
        m_trunc_ld_key = m_records.at(upto_lsn).m_trunc_key;
        THIS_LOGSTORE_LOG(TRACE, "Truncating logstore upto lsn={} , m_trunc_ld_key index {} offset {}", upto_lsn,
                          m_trunc_ld_key.idx, m_trunc_ld_key.dev_offset);
    }
```

Besroy: Yes, the case only exists when restart SM and trigger truncation on resource_mgr. Since the m_trunc_ld_key is not persisted on disk, so it will be {0,0} after restart, which is less than m_last_truncate_idx (updated on loading logs).

src/lib/logstore/log_store.cpp
Besroy: Updated to return out_of_bound_ld_key to let log dev truncate freely.

JacksonYao287:
@sanebay pls also take a look

src/lib/logstore/log_store.cpp
```diff
@@ -207,7 +222,10 @@ void HomeLogStore::truncate(logstore_seq_num_t upto_lsn, bool in_memory_truncate
 
 std::tuple< logstore_seq_num_t, logdev_key, logstore_seq_num_t > HomeLogStore::truncate_info() const {
     auto const trunc_lsn = m_start_lsn.load(std::memory_order_relaxed) - 1;
-    return std::make_tuple(trunc_lsn, m_trunc_ld_key, m_tail_lsn.load(std::memory_order_relaxed));
+    auto const tail_lsn = m_tail_lsn.load(std::memory_order_relaxed);
+
+    return (trunc_lsn == tail_lsn) ? std::make_tuple(trunc_lsn, logdev_key::out_of_bound_ld_key(), tail_lsn)
```
xiaoxichen, line 227: some comment to explain we return `logdev_key::out_of_bound_ld_key()` in case of empty

src/tests/test_log_dev.cpp
```diff
@@ -337,6 +337,101 @@ TEST_F(LogDevTest, ReTruncate) {
     read_all_verify(log_store);
 }
 
+TEST_F(LogDevTest, TruncateWithOverlappingLSN) {
```
xiaoxichen, line 340: what does overlapping means here ?  non-overlapping I suppose?

src/tests/test_log_dev.cpp
```diff
@@ -337,6 +337,101 @@ TEST_F(LogDevTest, ReTruncate) {
...
+
+    LOGINFO("Step 8: Append 500 entries")
+    cur_lsn = log_store->next_lsn();
+    kickstart_inserts(log_store, cur_lsn, 500);
+    ASSERT_EQ(log_store->next_lsn(), 2000500);
+
+    LOGINFO("Step 9: Read and verify all entries");
+    read_all_verify(log_store);
+}
+
```
xiaoxichen, line 383: can we add an UT to test the logdev truncation logic:
3 logstores share same logdev,
store1: insert 100
store2: insert 200
store3: insert 200
store2: truncate 100
store3: truncate 500
//now store1 [0,100), store2[100, 200), store3[500,500), check trun up to 0
store1: truncate 100
//now store1 [100,100), store2[100, 200), store3[500,500), check trun up to 100
store2: truncate 300
//now store1 [100,100), store2[300, 300), store3[500,500), all empty, check trun up to 500

Besroy: Thanks! Added ut and also udpated according to comments.

xiaoxichen (approved):

JacksonYao287 (commented):
LGTM, one small question

src/lib/logstore/log_dev.cpp
```diff
@@ -501,6 +502,7 @@ void LogDev::on_flush_completion(LogGroup* lg) {
     free_log_group(lg);
     m_log_records->truncate(upto_indx);
     m_last_flush_idx = upto_indx;
+    m_last_flush_ld_key = logdev_key{from_indx, dev_offset};
```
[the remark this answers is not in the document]
Besroy: Both `upto_indx` and `from_indx` work for truncating, the purpose using `from_indx` as the `min_safe_ld_key` is to keep things consistent with our usual truncate logic. In our normal flush or log loading processes, we always use `from_indx` as the trunc key. This approach might be more readable for others.

JacksonYao287 (approved):

PR #629, opened 2025-01-09
use log group lower bound as flush key

Besroy:
Ensure that the logic for setting the truncation key is consistent with the `on_write_completion` process, so that the log records saved in the log device completely match the records in `m_records`.


src/lib/logstore/log_dev.cpp
```diff
@@ -204,8 +204,7 @@ void LogDev::do_load(off_t device_cursor) {
         // Loop through each record within the log group and do a callback
         decltype(header->nrecords()) i{0};
         HS_REL_ASSERT_GT(header->nrecords(), 0, "nrecords greater then zero");
-        const auto flush_ld_key =
-            logdev_key{header->start_idx() + header->nrecords(), group_dev_offset + header->total_size()};
+        const auto flush_ld_key = logdev_key{header->start_idx(), group_dev_offset};
```
JacksonYao287, line 207: can you explain with more details about this ? I am not quite clear

Besroy: That corresponding to [issue 22](https://docs.google.com/document/d/16DMqv9J-JuNs5c25IBuX01QXe5WbxPm4B4hAfz4JNZM/edit?pli=1&tab=t.0#heading=h.zbaklzrlcs8t).
In the test case "restart all members simultaneously", after all members restart and a leader is elected, the leader updates config and stores a log with LSN N+1. However, the leader fails to read and send this new log to followers because it hasn't been flushed.

For example:
T1: SM flushes logs in the range [0-99] and [100-199] (corresponding raft lsn [1-100] and [101,200]).
T2: Restart SM. After the restart, logs in [0,99] are set to [a truncate_key_index of 100] due to this code line. 
T3: Truncate logs upto 49. Since log 49's truncate_key_index is 100, all logs below 100 in log dev are truncated. And truncate_lsn in log store updates to 49, m_start_lsn updates to 50.
T4: Restart SM again. Then m_start_lsn is loaded as 50, initializing m_records from 50.  However, since logs from 50 to 99 were truncated and can't be recreated during replay, there's a gap in the range [50,100). This results in m_records.active_upto() returning 49 instead of 199.
T5: SM is elected as leader, update config with raft lsn 201 and log store lsn 200, and attempts to flush the new log [by raft](https://github.com/eBay/NuRaft/blob/master/src/raft_server.cxx#L1768), but [fails due to an incorrect active_upto_lsn 49](https://github.com/eBay/HomeStore/blob/348e05d65112dcf92104492a79eb994103051454/src/lib/logstore/log_store.cpp#L280), leaving logstore_lsn 200 incomplete. 
Thus, when  the leader tries to retrieve raft log 201 (log store lsn 200) using log_entries_ext, which searches logs via m_records.foreach_all_completed, it cannot access it.

xiaoxichen (dismissed):

xiaoxichen (approved):

PR #631, opened 2025-01-16
Ensure end_of_append_batch is Called for All Raft Log Types

Besroy:
Previously, HomeRaftLogStore::end_of_append_batch was only invoked for app_logs, which required requests in m_lsn_req_map. This behavior caused issues when only non-app raft logs (e.g., conf logs) were appended, as the function would not be called, leaving m_last_durable_lsn outdated. Consequently, next_slot() could return incorrect values based on the stale m_last_durable_lsn. This update ensures that HomeRaftLogStore::end_of_append_batch is called for all raft log types, guaranteeing that all logs are flushed and m_last_durable_lsn is consistently updated everytime log_store's end_of_append_batch is executed

### DIFF
--- a/src/lib/replication/log_store/repl_log_store.cpp
+++ b/src/lib/replication/log_store/repl_log_store.cpp
@@ -44,11 +44,8 @@ void ReplLogStore::end_of_append_batch(ulong start_lsn, ulong count) {
     auto proposer_reqs = sisl::VectorPool< repl_req_ptr_t >::alloc();
     for (int64_t lsn = int64_cast(start_lsn); lsn <= end_lsn; ++lsn) {
         auto rreq = m_sm.lsn_to_req(lsn);
-        // Skip this call in proposer, since this method will synchronously flush the data, which is not required for
-        // leader. Proposer will call the flush as part of commit after receiving quorum, upon which time, there is a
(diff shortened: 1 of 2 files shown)


src/lib/replication/log_store/repl_log_store.cpp
```diff
@@ -92,6 +92,11 @@ void ReplLogStore::end_of_append_batch(ulong start_lsn, ulong count) {
         for (auto const& rreq : *proposer_reqs) {
             if (rreq) { rreq->add_state(repl_req_state_t::LOG_FLUSHED); }
         }
+    } else {
+        // Ensure log flush even if reqs and proposer_reqs are empty, as this is the case for raft config entries.
+        RD_LOGT("Raft Channel: end_of_append_batch, no data to be written, only flush logs from {} , count {}",
+                start_lsn, count);
+        HomeRaftLogStore::end_of_append_batch(start_lsn, count);
```
JacksonYao287, line 99: NIT, we can move this line out of judging area to make sure end_of_append_batch will be called for any case

JacksonYao287 (dismissed):
LGTM

xiaoxichen (approved):

xiaoxichen:
thanks it looks much cleaner.

PR #632, opened 2025-01-21
Bypass dependency bugs in snapshot transmission

koujl:
Bypass some NuRaft bugs in specific corner cases:

* Create a snapshot after adding a new member to prevent transmitting a snapshot with outdated configuration.
* Trigger cp_flush on last_obj in case apply_snapshot() is skipped dueto crash.

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -178,6 +178,15 @@ AsyncReplResult<> RaftReplDev::replace_member(const replica_member_info& member_
             RD_LOGI("Replace member added member={} to group_id={}", boost::uuids::to_string(member_in.id),
                     group_id_str());
 
+            // If enabled, create a snapshot here to ensure the new member will use the latest snapshot with itself in the config
+            if (raft_server()->get_current_params().snapshot_distance_ > 0) {
+                if (auto idx = raft_server()->create_snapshot(); idx > 0) {
+                    RD_LOGI("Created snapshot idx={} after adding member", idx);
+                } else {
+                    RD_LOGW("Failed to create snapshot after adding member");
+                }
+            }
+
             // Step 3. Append log entry to mark the old member is out and new member is added.
             auto rreq = repl_req_ptr_t(new repl_req_ctx{});
             replace_members_ctx members;
--- a/src/lib/replication/repl_dev/raft_state_machine.cpp
+++ b/src/lib/replication/repl_dev/raft_state_machine.cpp
@@ -367,11 +367,14 @@ void RaftStateMachine::save_logical_snp_obj(nuraft::snapshot& s, ulong& obj_id, 
     snp_data->is_last_obj = is_last_obj;
 
     // We are doing a copy here.
-    sisl::io_blob_safe blob{s_cast< size_t >(data.size())};
+    sisl::io_blob_safe blob{static_cast<uint32_t>(data.size())};
     std::memcpy(blob.bytes(), data.data_begin(), data.size());
     snp_data->blob = std::move(blob);
 
     m_rd.m_listener->write_snapshot_obj(snp_ctx, snp_data);
+    if (is_last_obj) {
+        hs()->cp_mgr().trigger_cp_flush(true).wait(); // ensure DSN is flushed to disk
+    }
 
     // Update the object offset.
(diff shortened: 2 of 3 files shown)


src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -367,11 +367,15 @@ void RaftStateMachine::save_logical_snp_obj(nuraft::snapshot& s, ulong& obj_id,
...
-    sisl::io_blob_safe blob{s_cast< size_t >(data.size())};
+    sisl::io_blob_safe blob{static_cast<uint32_t>(data.size())};
     std::memcpy(blob.bytes(), data.data_begin(), data.size());
     snp_data->blob = std::move(blob);
 
     m_rd.m_listener->write_snapshot_obj(snp_ctx, snp_data);
+    if (is_last_obj) {
+        m_rd.set_last_commit_lsn(s.get_last_log_idx());
+        m_rd.m_data_journal->set_last_durable_lsn(s.get_last_log_idx());
+    }
```
yuwmao, line 378: Please also invoke cp_flush `hs()->cp_mgr().trigger_cp_flush(true /* force */).get();`, otherwise some data might be lost if restart happens after log truncation.

yuwmao (commented):
LGTM

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -178,6 +178,11 @@ AsyncReplResult<> RaftReplDev::replace_member(const replica_member_info& member_
             RD_LOGI("Replace member added member={} to group_id={}", boost::uuids::to_string(member_in.id),
                     group_id_str());
 
+            // If enabled, create a snapshot here to ensure the new member will use the latest snapshot with itself in the config
+            if (raft_server()->get_current_params().snapshot_distance_ > 0) {
+                raft_server()->create_snapshot();
```
JacksonYao287, line 183: should we check the returned value here.?  return 0 means fail to create_snapshot

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -367,11 +367,15 @@ void RaftStateMachine::save_logical_snp_obj(nuraft::snapshot& s, ulong& obj_id,
...
     // We are doing a copy here.
-    sisl::io_blob_safe blob{s_cast< size_t >(data.size())};
+    sisl::io_blob_safe blob{static_cast<uint32_t>(data.size())};
     std::memcpy(blob.bytes(), data.data_begin(), data.size());
     snp_data->blob = std::move(blob);
 
     m_rd.m_listener->write_snapshot_obj(snp_ctx, snp_data);
+    if (is_last_obj) {
+        // flush DSN
+        hs()->cp_mgr().trigger_cp_flush(true);
```
JacksonYao287, line 377: it`s better to wait this until completed , which guarantee dsn is flushed.
```
hs()->cp_mgr().trigger_cp_flush(true).wait();
```

JacksonYao287 (commented):
genrally LGTM, only two minor comments , PTAL!

src/lib/replication/repl_dev/raft_repl_dev.cpp
koujl: Good point, added some warning logs in the latest push since I think returning an error may be too strict for a workaround.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -178,6 +178,15 @@ AsyncReplResult<> RaftReplDev::replace_member(const replica_member_info& member_
             RD_LOGI("Replace member added member={} to group_id={}", boost::uuids::to_string(member_in.id),
                     group_id_str());
 
+            // If enabled, create a snapshot here to ensure the new member will use the latest snapshot with itself in the config
+            if (raft_server()->get_current_params().snapshot_distance_ > 0) {
+                if (auto idx = raft_server()->create_snapshot(); idx > 0) {
+                    RD_LOGI("Created snapshot idx={} after adding member", idx);
+                } else {
+                    RD_LOGW("Failed to create snapshot after adding member");
```
JacksonYao287, line 186: we can add a TODO for handling error case and revisit here if necessary in the future

JacksonYao287 (approved):
LGTM

PR #633, opened 2025-01-22
Flush log store meta when compacting log

yuwmao:
In baseline resync scenario, log truncation won't flush the truncate result immediately(periodically flushed by resource manager), thus if durable_commit_lsn has been flushed but start_lsn of the logstore hasn't been flushed, it will trigger the [issue6](https://docs.google.com/document/d/1U2LZ_wIou_w9PLgEcGo3DF9MZtQH44GRJ5MVC5xhOcI/edit?tab=t.0#heading=h.y491romdh9x).
Add a flush meta function in logstore, and manually trigger flush in log truncation.

### DIFF
--- a/src/lib/replication/log_store/home_raft_log_store.cpp
+++ b/src/lib/replication/log_store/home_raft_log_store.cpp
@@ -366,7 +366,7 @@ bool HomeRaftLogStore::compact(ulong compact_lsn) {
         // we directly compact and truncate up to compact_lsn assuming there are dummy logs.
         REPL_STORE_LOG(DEBUG, "Compact with log holes from {} to={}", cur_max_lsn + 1, to_store_lsn(compact_lsn));
     }
-    m_log_store->truncate(to_store_lsn(compact_lsn));
+    m_log_store->truncate(to_store_lsn(compact_lsn), false);
     return true;
 }
 
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.6.13"
+    version = "6.6.14"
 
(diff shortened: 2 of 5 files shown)


src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -366,7 +366,15 @@ bool HomeRaftLogStore::compact(ulong compact_lsn) {
...
         REPL_STORE_LOG(DEBUG, "Compact with log holes from {} to={}", cur_max_lsn + 1, to_store_lsn(compact_lsn));
     }
+#ifdef _PRERELEASE
+    auto delay = iomgr_flip::instance()->get_test_flip< long >("simulate_log_compact_delay");
+    if (delay) {
+        REPL_STORE_LOG(DEBUG, "Simulating log compaction with delay, delay:{}", delay.get());
+        std::this_thread::sleep_for(std::chrono::milliseconds(delay.get()));
+    }
+#endif
     m_log_store->truncate(to_store_lsn(compact_lsn));
```
JacksonYao287, line 376: IMO, we can change this line to 
```
m_log_store->truncate(to_store_lsn(compact_lsn), false);
```
this will be enough.

home_raft_log_store will have one seperate logdev,  and there will be only one logstore on this logdev.

this line will change `in_memory_truncate_only` to false and as result it will trigger logdev truncation, where the metadata of logstore will be updated

JacksonYao287 (commented):
LGTM. only a minor comment

src/lib/replication/log_store/home_raft_log_store.cpp
yuwmao: Thanks, that make it even simpler!

JacksonYao287 (approved):
LGTM

koujl (approved):

PR #635, opened 2025-01-24
Persist logstore superblock when logdev truncation is unnecessary

Besroy:
This change ensures that the logstore can get the correct start LSN during recovery. Avoid the following scenario:

T1: Follower1 appends logs up to 100, then is stopped by a sigkill.
T2: Upon restart, since the leader's log range is 1000-2500, a baseline resync is triggered using snapshot 2000.
T3: Follower1 completes the baseline resync (start_lsn=2001, tail_lsn=2000), but m_trunc_ld_key is not updated since we cannot get a valid device offset for LSN 2000.
T4: Follower1 appends logs from 2001 to 2500, making tail_lsn greater than 2000.
T5: During logdev truncation, the truncation info is found at first. Since trunc_lsn < tail_lsn, it returns m_trunc_ld_key (still {0,0}), then exits without persist the logstore sb.
T6: Follower1 is killed again, and upon restart, its start index in the store superblock remains 0, incorrectly interpreting the range as [1,2500].

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.6.16"
+    version = "6.6.17"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/lib/logstore/log_dev.cpp
+++ b/src/lib/logstore/log_dev.cpp
@@ -542,7 +542,18 @@ uint64_t LogDev::truncate() {
     }
 
     // There are no writes or no truncation called for any of the store, so we can't truncate anything
-    if (min_safe_ld_key.idx <= 0 || min_safe_ld_key.idx <= m_last_truncate_idx) return 0;
+    if (min_safe_ld_key.idx <= 0 || min_safe_ld_key.idx <= m_last_truncate_idx) {
+        // Persist the logstore superblock to ensure correct start LSN during recovery. Avoid such scenario:
+        // 1. Follower1 appends logs up to 100, then is stopped by a sigkill.
+        // 2. Upon restart, a baseline resync is triggered using snapshot 2000.
+        // 3. Baseline resync completed with start_lsn=2001, but m_trunc_ld_key remains {0,0} since we cannot get a valid
+        //    device offset for LSN 2000 to update it.
+        // 4. Follower1 appends logs from 2001 to 2500, making tail_lsn > 2000.
+        // 5. Get m_trunc_ld_key={0,0}, goto here and return 0 without persist.
+        // 6. Follower1 is killed again, after restart, its start index remains 0, misinterpreting the range as [1,2500].
+        m_logdev_meta.persist();
+        return 0;
+    }
 
     uint64_t const num_records_to_truncate = uint64_cast(min_safe_ld_key.idx - m_last_truncate_idx);
 


src/lib/logstore/log_store.cpp
```diff
@@ -193,10 +193,13 @@ void HomeLogStore::truncate(logstore_seq_num_t upto_lsn, bool in_memory_truncate
     // So upto_lsn > m_tail_lsn is expected to exist only in baseline resync path.
     // In baseline resync path, we truncate all entries up to upto_lsn, and update m_tail_lsn and m_next_lsn
     // to make sure logstore's idx is always = raft's idx - 1.
+    // Additionally, m_trunc_ld_key should be updated to match the truncation key of tail_lsn.
+    // This ensures that logdev can obtain the correct trunc index in scenarios where new logs appended after ls truncate.
     if (upto_lsn > m_tail_lsn) {
         THIS_LOGSTORE_LOG(WARN,
                           "Truncating issued on lsn={} which is greater than tail_lsn={}",
                           upto_lsn, m_tail_lsn.load(std::memory_order_relaxed));
+        m_trunc_ld_key = m_records.at(m_tail_lsn).m_trunc_key;
```
JacksonYao287, line 202: not sure should we move this line after line 215 and change it to 
```
m_trunc_ld_key = m_records.at(upto_lsn).m_trunc_key;
```

zhiteng:
@Besroy The PR description should be in commit message as well.

xiaoxichen (commented):
I think the proper fix is let `logdev::truncate()` update store superblock even if truncation key is zero here (https://github.com/eBay/HomeStore/blob/707c11137a8de7dc859a6d00e933b8b33eee59e8/src/lib/logstore/log_dev.cpp#L546) ,  reasons below

The fundamental issue is,  if we truncate to LSN 2000,  and the LSN 2000 never exists,  we cannot get a valid device offset (tunc_key).    Previous PR we fixed partial of the issue that if the store is empty then we return out_of_bound_key means free to truncate,  this PR you are fixing the counter part that the logstore is not empty. 

But you still having the same issue that we cannot have a proper trunc_key for LSN 2000(or anything below 2000),  and returning the offset of LSN 2001 is not right.This PR tries to find a trunc_key for lsn 2000, which I get the intent but not sure how it works,  seems L202 and L215 will both return an zero key.

 The proper behavior for this logstore should be `DO NOT TRUNCATE for now as I dont know where i can truncate to`.  We dont have issue with skipping truncation instead the issues comes from we skip persisting the superblock.   Then lets fix this.

src/lib/logstore/log_store.cpp
Besroy: Updated the logic: will not update `m_trunc_ld_key` if we don't have a valid offset for a 'dummy lsn'

Besroy:
> I think the proper fix is let `logdev::truncate()` update store superblock even if truncation key is zero here (
> 
> https://github.com/eBay/HomeStore/blob/707c11137a8de7dc859a6d00e933b8b33eee59e8/src/lib/logstore/log_dev.cpp#L546
> 
> ) , reasons below
> The fundamental issue is, if we truncate to LSN 2000, and the LSN 2000 never exists, we cannot get a valid device offset (tunc_key). Previous PR we fixed partial of the issue that if the store is empty then we return out_of_bound_key means free to truncate, this PR you are fixing the counter part that the logstore is not empty.
> 
> But you still having the same issue that we cannot have a proper trunc_key for LSN 2000(or anything below 2000), and returning the offset of LSN 2001 is not right.This PR tries to find a trunc_key for lsn 2000, which I get the intent but not sure how it works, seems L202 and L215 will both return an zero key.
> 
> The proper behavior for this logstore should be `DO NOT TRUNCATE for now as I dont know where i can truncate to`. We dont have issue with skipping truncation instead the issues comes from we skip persisting the superblock. Then lets fix this.

Thanks @xiaoxichen ! That would make code clearer and simpler.

xiaoxichen (dismissed):
lgtm except suggesting copy/summary the scenario  into L546.

xiaoxichen (approved):

PR #636, opened 2025-01-27
graceful shutdown

JacksonYao287:
this PR introduce graceful shutdown to homestore. the basic idea is to introduce a counter and flag for each service or component to make sure all the pending requests are handled and reject later api calls from upper layer before stop.


JacksonYao287:
https://docs.google.com/document/d/11SP84tI9AwB_FhRSFJoC8oZIHECOn4o2mbooZf4W0EY/edit?tab=t.0#heading=h.gitenqkosfz6

related doc

src/include/homestore/blkdata_service.hpp
```diff
@@ -224,6 +226,19 @@ class BlkDataService {
     std::unique_ptr< BlkReadTracker > m_blk_read_tracker;
     std::shared_ptr< ChunkSelector > m_custom_chunk_selector;
     uint32_t m_blk_size;
+
+private:
+    // graceful shutdown related
+    std::atomic_bool m_stopping{false};
+    mutable std::atomic_uint64_t pending_request_num{0};
```
szmyd, line 233: Can we use a [Counting Semaphore](https://en.cppreference.com/w/cpp/thread/counting_semaphore) so we can make it event driven instead of check->sleep->check logic?

src/lib/replication/service/raft_repl_service.h
```diff
@@ -54,6 +54,7 @@ class RaftReplService : public GenericReplService,
 
 public:
     RaftReplService(cshared< ReplApplication >& repl_app);
+    ~RaftReplService();
```
szmyd, line 57: Since this is a derived class we should set this to `override` so we enforce the base classes have declared their destructors `virtual` yes?

szmyd (commented):
Some comments.

src/include/homestore/blkdata_service.hpp
JacksonYao287: thanks for the suggestion, I got your point here. well , I think the scenario here is a little different from the one suitable for counting_semaphore.

1 counting_semaphore is used to set a limitation to the max concurrency of actions simultaneously. in the scenario here, I do not want set any limitation. all the api can be called as much as possible as long as the service is not stopped.

2 when starting shutting down, we should exactly know whether all the pending requests are completed. however , counting_semaphore does not provide an api to see if all the action that acquired the semaphore has release it.  with counting_semaphore, what we can do it only to acquire or try_acquire.

pls correct me if I am wrong @szmyd

src/lib/replication/service/raft_repl_service.h
JacksonYao287: yes，we need declare a virtual destructor for the base class and set override in the derived clase.  thanks!

src/include/homestore/blkdata_service.hpp
szmyd: You're right; I am wrong, I had it inverted in my head. What we'd want is a shared_mutex that everyone takes a read lock (increment the lock count), and the "waiter" takes a write_lock which will only unlock once the readers are done.

This solution you have currently is fine since it only is exercised during shutdown though. I will continue to review and ignore this part.

szmyd:
I thought about this more; and I'm fine with this approach for now. I'd love to come up with something a little more clever that can't be error prone if someone misses a `incr`/`decr` call somewhere, but I'll approve for now. Please resync and I'll approve the result asap.

szmyd (approved):
lgtm

PR #643, opened 2025-02-11
Clean up old logs at the beginning of baseline resync.

Besroy:
If the follower restarts during baseline resync, it will replay remaining logs first. However, we have already cleared shard info at the beginning of resync, so we cannot get shard while replaying logs, which will raise errors.
This change clean up old logs in log store at the beginning of baseline resync to avoid this situation.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.6.17"
+    version = "6.6.18"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -477,6 +477,9 @@ public:
     /// @return true if ready, false otherwise
     virtual bool is_ready_for_traffic() const = 0;
 
+    /// @brief Clean up resources on this repl dev.
+    virtual void purge() = 0;
+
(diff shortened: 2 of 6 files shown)


xiaoxichen (commented):
my 0.02

The truncate is misleading and as far as we spent effort on removing this part, do not added back,    a explicit `HomeRaftLogStore::purge_logs()` or `HomeRaftLogStore::purge_all_logs()` might have better readability.

Another suggestion is do not do it in  `save_logical_snp_obj` as it will be occur in every snapshot batch,  better to add it to `HomeObject::pg_destroy` then call down to `repl_dev_::purge()` , in the implementation of RaftReplDev we clean up logs.

Besroy:
> my 0.02
> 
> The truncate is misleading and as far as we spent effort on removing this part, do not added back, a explicit `HomeRaftLogStore::purge_logs()` or `HomeRaftLogStore::purge_all_logs()` might have better readability.
> 
> Another suggestion is do not do it in `save_logical_snp_obj` as it will be occur in every snapshot batch, better to add it to `HomeObject::pg_destroy` then call down to `repl_dev_::purge()` , in the implementation of RaftReplDev we clean up logs.

Thanks! updated to `purge` and `purge_all_logs` at HS side.  `save_logical_snp_obj` will be executed at the begin of snapshot, then obj type will be set with snp_obj_id_type_app, wh

xiaoxichen (approved):

xiaoxichen:
Yes you are right...lets merge this as is

JacksonYao287 (approved):
LGTM

PR #644, opened 2025-02-11
Remove snapshot creation in add_member

yuwmao:
Previously, since add_member is only handled by leader, thus only leader triggers a latest snapshot. If leader restarts, the new leader will use the old snapshot for baseline resync. 
Since the nuraft reconfigure issue has been fixed, so we don't need this work around.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.6.17"
+    version = "6.6.18"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -178,15 +178,6 @@ AsyncReplResult<> RaftReplDev::replace_member(const replica_member_info& member_
             RD_LOGI("Replace member added member={} to group_id={}", boost::uuids::to_string(member_in.id),
                     group_id_str());
 
-            // If enabled, create a snapshot here to ensure the new member will use the latest snapshot with itself in the config
-            if (raft_server()->get_current_params().snapshot_distance_ > 0) {
-                if (auto idx = raft_server()->create_snapshot(); idx > 0) {
-                    RD_LOGI("Created snapshot idx={} after adding member", idx);
-                } else {
-                    RD_LOGW("Failed to create snapshot after adding member");
-                }
-            }
-
             // Step 3. Append log entry to mark the old member is out and new member is added.
             auto rreq = repl_req_ptr_t(new repl_req_ctx{});
             replace_members_ctx members;


xiaoxichen:
Intent is correct, implementation wise, can we move this to config_commit?
https://github.com/eBay/HomeStore/pull/644/files#diff-c9fac27171101b8113ad29c31e1f4f34bcd0eba5c284a3709e6369da935e1827L234?

yuwmao:
> Intent is correct, implementation wise, Is that better to move this to config_commit? https://github.com/eBay/HomeStore/pull/644/files#diff-c9fac27171101b8113ad29c31e1f4f34bcd0eba5c284a3709e6369da935e1827L234?

In the `config_commit`, we can only get `new_config`, it seems hard for us to know what's the action/changes.

xiaoxichen:
> > Intent is correct, implementation wise, Is that better to move this to config_commit? https://github.com/eBay/HomeStore/pull/644/files#diff-c9fac27171101b8113ad29c31e1f4f34bcd0eba5c284a3709e6369da935e1827L234?
> 
> In the `config_commit`, we can only get `new_config`, it seems hard for us to know what's the action/changes. Additionally, we cannot compare the new_config with the prior config, as [the new config has been saved](https://github.com/eBay/NuRaft/blob/cf80969e7725def8df157bc15b77c3dafc510a00/src/handle_commit.cxx#L416) before the [commit_config callback](https://github.com/eBay/NuRaft/blob/cf80969e7725def8df157bc15b77c3dafc510a00/src/handle_commit.cxx#L428).

make sense ,  taking a snapshot on every config_commit is wrong as well..

My thinking is the BR started once the new_server sent the `join_cluster_response`,  the commit of the config change(lsn 110) ,  is the only sync point before the BR, using the old snapshot on leader , say LSN 100

If we do that in the pre-commit/commit of HS_CTRL_REPLACE,  the snapshot is taken on LSN 111 then it doesnt help in this case.

Maybe we need to rethink if we need explicitly take a snapshot in this case?

yuwmao:
> 
> My thinking is the BR started once the new_server sent the `join_cluster_response`, the commit of the config change(lsn 110) , is the only sync point before the BR, using the old snapshot on leader , say LSN 100
> 
> If we do that in the pre-commit/commit of HS_CTRL_REPLACE, the snapshot is taken on LSN 111 then it doesnt help in this case.
> 
> Maybe we need to rethink if we need explicitly take a snapshot in this case?

In our case, since we enable new_joiner type, the baseline resync is trigggered in [`request_append_entries`](https://github.com/eBay/NuRaft/blob/3008beff7782f9a49c5e429f81135c13d919186d/src/handle_append_entries.cxx#L301) and [`create_append_entries_req`](https://github.com/eBay/NuRaft/blob/3008beff7782f9a49c5e429f81135c13d919186d/src/handle_append_entries.cxx#L542). 
However the first `create_append_entries_req` won't trigger baseline resync because [the next_log_idx has been set to the leader's cur_next_idx](https://github.com/eBay/NuRaft/blob/3008beff7782f9a49c5e429f81135c13d919186d/src/handle_append_entries.cxx#L452) (maybe it is intended to ask for the real next_log_id). So the new member will reject the first req and return the actual next_log_idx=1. Then the 2nd `create_append_entries_req` will trigger the baseline resync.

Since there is an invalid remote call and the precommit is executed locally on the leader side, so new snapshot creation is more likely to happen before baseline resync. 
But the case you mentioned could happen as well, in that case, a full baseline resync re-execution is unavoidable.
I think the mechanism still can work without this change(perform a new baseline resync using the old snapshot), but this change is a nice-to-have optimization which most resumption cases may benefit.

xiaoxichen:
If we dont take the new snapshot on leader side,  all snapshot should be consistent on all members as they have same snapshot distance right?
So I am wondering if this is a negative optimization.  IIRC we added https://github.com/eBay/HomeStore/pull/644/files#diff-e5ba64483f8c347aa367fa3c61121269fcdd74dabfe8f66f5a95bf6cf3939876L181-L189 to work around the nuraft issue,  if the nuraft issue resolved and we are seeing bad impact introduced by this, maybe thinking about remove it?

Or thinking a bit wild,  if we can take a consistent snapshot (through  `HS_CTRL_SNAPSHOT` for example) before the add_member,  is that even better?

xiaoxichen:
should we merge this, or wait for your fix to persist snapshot?

yuwmao:
> should we merge this, or wait for your fix to persist snapshot?

Let‘s merge it as it won't cause bug, just do BR twice when leader restarts.

PR #646, opened 2025-02-11
Fix the migration path for pre-3.8 LogDevs.

szmyd:
Log Devices pre-3.8 did not contain a superblock for the rollback metadata. We were not handling this absense correctly and crashing when loading a <=3.7.x homestore drive.


raakella1 (approved):
LG!

PR #647, opened 2025-02-12
OOM anf metrics fixes

raakella1:
This PR has 2 fixes

- set the wb req intrusive ptr to null in the completion callback to allow it to be destroyed
- put the unmap counter in the NDEBUG preprocessor scope to disable it for non debug builds.


szmyd (approved):
lgtm

PR #652, opened 2025-02-24
fix dead lock in graceful shutdown

JacksonYao287:
m_rd_map_mtx will be locked when cp_flush repl_service, so we need to release it ASAP in case of other component triggers cp, which will also cp_flush repl_service

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.6.22"
+    version = "6.6.23"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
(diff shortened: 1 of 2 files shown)


JacksonYao287:
pending_request_num mechanism is only useful for leader. for a follower, when we call RaftReplService::stop(), it is might still receiving install_snapshot req from leader , [where a cp_flush will be triggered](https://github.com/yuwmao/HomeObject/blob/0898dcbf4dac179181663b56c637ac54ac9166d8/src/lib/homestore_backend/snapshot_receive_handler.cpp#L369). in cp flush,  [cp context of repl_service will be prepared](https://github.com/eBay/HomeStore/blob/14b4ab11c755447f21a66f6b3c9b41b0ad445c90/src/lib/replication/service/raft_repl_service.cpp#L602). here, iterate_repl_devs will be called and try to [lock m_rd_map_mtx](https://github.com/eBay/HomeStore/blob/14b4ab11c755447f21a66f6b3c9b41b0ad445c90/src/lib/replication/service/generic_repl_svc.cpp#L60), but now this follower is being stopped and m_rd_map_mtx is locked by the stop caller thread.

this PR fixes this issue

szmyd (approved):

PR #653, opened 2025-02-25
Minor fixs

Besroy:
1. fix char* copy logic, add \0 at the end of array
 2. update nuraft_mesg version which has https://github.com/eBay/nuraft_mesg/pull/117

### DIFF
--- a/src/lib/device/device.h
+++ b/src/lib/device/device.h
@@ -36,6 +36,7 @@ VENUM(vdev_multi_pdev_opts_t, uint8_t, // Indicates the style of vdev when multi
 struct vdev_info {
     static constexpr size_t size = 512;
     static constexpr size_t user_private_size = 256;
+    static constexpr size_t max_name_len = 64;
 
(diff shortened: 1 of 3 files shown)


src/lib/device/device.h
```diff
@@ -59,7 +59,10 @@ struct vdev_info {
     uint32_t get_vdev_id() const { return vdev_id; }
     uint64_t get_size() const { return vdev_size; }
 
-    void set_name(const std::string& n) { std::strncpy(charptr_cast(name), n.c_str(), 63); }
+    void set_name(const std::string& n) {
+        std::strncpy(charptr_cast(name), n.c_str(), 63);
+        name[63] = '\0';
```
koujl, line 64: Suggest adding a constexpr variable to replace these `63/64` magic numbers.

Besroy: Thanks! updated.

xiaoxichen (approved):

PR #654, opened 2025-02-26
Fix missing fields in RaftReplDev::save_state

koujl:
Newly added fields in `nuraft::srv_state` should also be persisted/loaded.

Reference: https://github.com/eBay/NuRaft/blob/2724344f530c7d7d8551fb8d2c7bb5ed75e9e8e1/include/libnuraft/srv_state.hxx

https://github.com/eBay/NuRaft/commit/da4a1aa23ec98271347c6cda50c521a290cf14d6
https://github.com/eBay/NuRaft/commit/3008beff7782f9a49c5e429f81135c13d919186d

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.6.24"
+    version = "6.6.25"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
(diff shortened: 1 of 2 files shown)


Besroy (approved):

xiaoxichen (commented):
1.  what is the reason that we try to serialize by ourself into json, instead of relying on the  serialize member function of srv_state?

2.   We can ignore backward compatibility now as no serious deployment. But we need to thinnk about it. It is out of the scope of this PR but most of our superblk structure do not include a version.

koujl:
Issue raised for replacing JSON: https://github.com/eBay/HomeStore/issues/656

PR #655, opened 2025-02-26
Expose new joiner in replication status

yuwmao:


src/include/homestore/replication/repl_decls.h
```diff
@@ -78,6 +78,7 @@ struct peer_info {
     uint64_t replication_idx_;
     // The elapsed time since the last successful response from this peer, set to 0 on leader
     uint64_t last_succ_resp_us_;
+    bool is_new_joiner{false};
```
xiaoxichen, line 81: Lets do it in the grpc spec clearly state which field is trustable.

Now ever member in PG report status,  the replication_idx only valid on leader (other replica are all zero) and seems like the is_new_joiner only valid for local node?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1117,10 +1117,14 @@ replica_id_t RaftReplDev::get_leader_id() const {
...
     auto rep_status = m_repl_svc_ctx->get_raft_status();
+    auto srv_cfg = raft_server()->get_config()->get_server(raft_server()->get_id());
     for (auto const& pinfo : rep_status) {
-        pi.emplace_back(peer_info{.id_ = boost::lexical_cast< replica_id_t >(pinfo.id_),
-                                  .replication_idx_ = pinfo.last_log_idx_,
-                                  .last_succ_resp_us_ = pinfo.last_succ_resp_us_});
+        auto peer = peer_info{.id_ = boost::lexical_cast< replica_id_t >(pinfo.id_),
+                              .replication_idx_ = pinfo.last_log_idx_,
+                              .last_succ_resp_us_ = pinfo.last_succ_resp_us_};
+        if (pinfo.id_ == my_replica_id_str()) { peer.is_new_joiner = srv_cfg->is_new_joiner(); }
```
xiaoxichen, line 1125: 1. better if we can get is_new_joiner for all member, at least on leader.
2. if this change can be push down to nuraf_mesg?

yuwmao: As new_joiner flag is in raft srv_config, technically all members can get new joiner state. 
It would be better if it can be pushed down to nuraf_mesg.

xiaoxichen (approved):

JacksonYao287 (approved):
LGTM

PR #657, opened 2025-03-05
Add event callbacks into listener for upper layer

JacksonYao287:
some events need to be handled by upper layer. This PR add three event: 

1 fetch_data: upper layer can decide which data to be returned

2 no_space_left: this error should be handled by upper layer if
  necessary

3 on_log_replay_done: after log replay is done and before joining raft
  group, upper layer might do something

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.6.25"
+    version = "6.6.26"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.6.26"
+    version = "6.7.0"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -370,11 +370,10 @@ public:
     /// @brief ask upper layer to decide which data should be returned.
     // @param header - header of the log entry.
     // @param blkid - original blkid of the log entry
-    // @param sgs_vec - vector of sgs to be filled with data
+    // @param sgs - sgs to be filled with data
     // @param lsn - lsn of the log entry
     virtual folly::Future< std::error_code > on_fetch_data(const int64_t lsn, const sisl::blob& header,
-                                                           const MultiBlkId& blkid,
-                                                           std::vector< sisl::sg_list >& sgs_vec) {
+                                                           const MultiBlkId& blkid, sisl::sg_list& sgs) {
         return folly::makeFuture< std::error_code >(std::error_code{});
     }
 
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -11,6 +11,7 @@
 #include <sisl/grpc/generic_service.hpp>
 #include <sisl/grpc/rpc_client.hpp>
 #include <homestore/replication/repl_decls.h>
+#include <homestore/blkdata_service.hpp>
 #include <libnuraft/snapshot.hxx>
 
 namespace nuraft {
@@ -374,7 +375,8 @@ public:
     // @param lsn - lsn of the log entry
     virtual folly::Future< std::error_code > on_fetch_data(const int64_t lsn, const sisl::blob& header,
                                                            const MultiBlkId& blkid, sisl::sg_list& sgs) {
-        return folly::makeFuture< std::error_code >(std::error_code{});
+        // default implementation is reading by blkid directly
+        return data_service().async_read(blkid, sgs, sgs.size);
     }
 
     /// @brief ask upper layer to handle no_space_left event
@@ -384,10 +386,6 @@ public:
 
     /// @brief when restart, after all the logs are replayed and before joining raft group, notify the upper layer
     virtual void on_log_replay_done(const group_id_t& group_id){};
-
-    /// @brief decide whether a particular type of event need to be handled by upper layer
-    // TODD: if we have many events to handle, we can put all the judgement in one function
-    virtual bool need_to_handle_fetch_data() { return false; }
 
 private:
     std::weak_ptr< ReplDev > m_repl_dev;
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -47,6 +47,20 @@ VENUM(journal_type_t, uint16_t,
       HS_CTRL_REPLACE = 3, // Control message to replace a member
 )
 
+// some events need to be handled by upper layer
+VENUM(event_type_t, uint8_t,
+      RD_FETCH_DATA = 0,      // fetch data
+      RD_NO_SPACE_LEFT = 1,   // no space left error
+      RD_LOG_REPLAY_DONE = 2, // log replay done
+                              // add more if needed for other applications.
+)
+
+using fetch_data_handler_t = folly::Future< std::error_code > (*)(int64_t lsn, const sisl::blob& header,
+                                                                  std::vector< sisl::sg_list >& sgs_vec);
+using no_space_left_handler_t = folly::Future< folly::Unit > (*)(uint32_t pdev_id, chunk_num_t chunk_id);
+
+using log_replay_done_handler_t = void (*)();
+
 // magic num comes from the first 8 bytes of 'echo homestore_resync_data | md5sum'
 static constexpr uint64_t HOMESTORE_RESYNC_DATA_MAGIC = 0xa65dbd27c213f327;
 static constexpr uint32_t HOMESTORE_RESYNC_DATA_PROTOCOL_VERSION_V1 = 0x01;
@@ -367,6 +381,10 @@ public:
     /// @brief Free up user-defined context inside the snapshot_obj that is allocated during read_snapshot_obj.
     virtual void free_user_snp_ctx(void*& user_snp_ctx) = 0;
 
+    /// @brief some repl_dev event might be handle by upper layer. This is where homestore gets the callbacks for these
+    /// events.
+    virtual void* get_event_handler(event_type_t event_type) { return nullptr; }
+
 private:
     std::weak_ptr< ReplDev > m_repl_dev;
 };
@@ -449,7 +467,7 @@ public:
     /// @brief Clean up resources on this repl dev.
     virtual void purge() = 0;
 
-    virtual std::shared_ptr<snapshot_context> deserialize_snapshot_context(sisl::io_blob_safe &snp_ctx) = 0;
+    virtual std::shared_ptr< snapshot_context > deserialize_snapshot_context(sisl::io_blob_safe& snp_ctx) = 0;
 
     virtual void attach_listener(shared< ReplDevListener > listener) { m_listener = std::move(listener); }
 
@@ -459,6 +477,8 @@ public:
             m_listener.reset();
         }
     }
+
+    virtual shared< ReplDevListener > get_listener() { return m_listener; }
 
     // we have no shutdown for repl_dev, since shutdown repl_dev is done by repl_service
     void stop() {
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -385,7 +385,7 @@ public:
     virtual void on_log_replay_done(const group_id_t& group_id){};
 
     /// @brief decide whether a particular type of event need to be handled by upper layer
-    // TODD: if we have many events to handle, we can all the judgement in one function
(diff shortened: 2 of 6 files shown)


src/include/homestore/replication/repl_dev.h
```diff
@@ -367,6 +381,10 @@ class ReplDevListener {
     /// @brief Free up user-defined context inside the snapshot_obj that is allocated during read_snapshot_obj.
     virtual void free_user_snp_ctx(void*& user_snp_ctx) = 0;
 
+    /// @brief some repl_dev event might be handle by upper layer. This is where homestore gets the callbacks for these
+    /// events.
+    virtual void* get_event_handler(event_type_t event_type) { return nullptr; }
```
xiaoxichen, line 386: This is anti-pattern with existing CB.   Suggesting follow the member function pattern for all the CB . If giving upper layer a choice not to implement such CB is important,  we can make it non-virtual.

xiaoxichen (commented):
aside from the comments regarding the implementation.  
Would love to see the design doc turned into a set of GH issues/tasks with estimation. That makes the project tracking easier.

JacksonYao287:
homestore, as a storage engine, may be used by different upper layer, which are interesting in different homestore event。
if we add a callback for each type of event, every upper layer need to implement the corresponding callback, even if they don`t care about these event. moreover, if we add a new event, all the upper layer need to implement this callback and recompile.

the intent of the interface here is that 
1 upper layer only handle those event it cares about.
2 a new event type will not result in recompiling and code changing, so that the impaction area is limited.

JacksonYao287:
@szmyd @yamingk  pls take a look and left your comments, what do you think? or do you have any other suggestion?. I am also good with @xiaoxichen `s idea.

conanfile.py
```diff
@@ -9,7 +9,7 @@
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.6.25"
+    version = "6.6.26"
```
szmyd, line 12: Regardless of how we implement this; will require at least a MINOR version change to `6.7.0` to signify the added call that a downstream consumer can begin to use starting here.

src/include/homestore/replication/repl_dev.h
```diff
@@ -47,6 +47,20 @@ VENUM(journal_type_t, uint16_t,
       HS_CTRL_REPLACE = 3, // Control message to replace a member
 )
 
+// some events need to be handled by upper layer
+VENUM(event_type_t, uint8_t,
+      RD_FETCH_DATA = 0,      // fetch data
+      RD_NO_SPACE_LEFT = 1,   // no space left error
+      RD_LOG_REPLAY_DONE = 2, // log replay done
```
yamingk, line 54: Can you also share some more insight with a use case (probably GC) when this log replay done will be used?

src/include/homestore/replication/repl_dev.h
```diff
@@ -47,6 +47,20 @@ VENUM(journal_type_t, uint16_t,
       HS_CTRL_REPLACE = 3, // Control message to replace a member
 )
 
+// some events need to be handled by upper layer
+VENUM(event_type_t, uint8_t,
+      RD_FETCH_DATA = 0,      // fetch data
+      RD_NO_SPACE_LEFT = 1,   // no space left error
```
yamingk, line 53: Shouldn't we change the signature of `ReplDev::aysnc_alloc_write` to return no space left in return value (since block alloc is being done in sync way) so that HomeObject have a context when no space left is happening. Registering event handler is also okay, but seems unnecessary, IMHO. 

This signature change will also be helpful for NuBlox 2.0.

conanfile.py
JacksonYao287: ack

src/include/homestore/replication/repl_dev.h
JacksonYao287: pls see my comments

on_log_replay_done: when resetart , after log is replayed and before joining raft group, we might need some other steps, for example, if emergency gc crash, when restart, we need wait for it to be completed before joining the raft group.

JacksonYao287:
@yamingk 
>in the PR description, push_data ==> fetch_data, correct?

yes, it`s a typo, will change it.

>Basically every originator who received a fetch data request from its peer needs to go through HomeObject with GC's presence. It should have nothing todo with push data path?

yes, it has nothing todo with push data path. for Homeobject case, we needs to go through HomeObject for every fetch_data request and I provide a callback for this. for other use case, if the handler is not provided, it will go throught the homestore default code path

src/include/homestore/replication/repl_dev.h
yamingk: I didn't find where the reasoning for emergency gc (EGC) is mentioned in code. 
If the logic is HO relies on this event to delaying PG to join raft group, before EGC is done, can we simplify a little bit by persiting a ECG flag in PG SB and persist it before EGC happens and clear the flag after EGC finishes. If EGC crashes, on reboot HO knows from PG SB that EGC is in the middle and delaying the PG from jioning the raft group before ECD finishes, right?

xiaoxichen: There is no hook-point between log replay and joining raft group , both controlled by HS. A hook point is needed


https://github.com/eBay/HomeStore/blob/6756b810a0d8ab136c4f99ffafd2ac75a243f6f1/src/lib/replication/service/raft_repl_service.cpp#L150-L166

src/include/homestore/replication/repl_dev.h
JacksonYao287: `ReplDev::aysnc_alloc_write` is only called by raft leader, and emergency gc only happens to follower when it gets a no_space_left. If leader gets a no_space_left error from `ReplDev::aysnc_alloc_write` , it will directly return failure to client to let it retry.

so this change is not necessary for nuobject. if it is helpful to you, pls make this change in a seperate PR

src/include/homestore/replication/repl_dev.h
JacksonYao287: yes , xiaoxi is right

src/include/homestore/replication/repl_dev.h
JacksonYao287: I have change the interfaces ， ptal

src/include/homestore/replication/repl_dev.h
yamingk: What will HO do on receipt of log replay done for a particular PG in presence of EGC meta blk? Will it do sync operation to finish EGC until the callback finishes? From code snip above, looks like it has to wait for this PG's EGC to finish in order to move on with other PG in the m_rd_map to join group, is it expected? 

Can we persist something in the raft dev sb in homestore to consume to skip this PG's joining of the raft if there is a EGC flag in it, and let HO to handle the PG SB and EGC during recovery that only affects this PG and after EGC is done, it do something to join the raft specificially? Does it make sense?

JacksonYao287: >From code snip above, looks like it has to wait for this PG's EGC to finish in order to move on with other PG in the m_rd_map to join group, is it expected?

yes , since different repl_dev is independent from others, as an optimization, we can make them join raft group concurrently. so that your concern can be resolved.

>Alternatively, does it make sense to persist EGC flag in the correspoinding PG's raft dev sb in homestore to consume to skip this PG's joining of the raft if there is a EGC flag in it, and let HO to handle the PG SB and EGC during recovery that only affects this PG and after EGC is done, it do something to join the raft specificially?

I do get your point. however, from the perspective of homestore itselft, it is unaware of PG, which is a homeobject specific concept. PG is created only in homeobject depending on repl_dev.  in other words,  even if homestore finds  this flag,  it has no idea about what does it mean and how to handle it. only homeobject knows that it means emergency GC.

Actually, when we create pg, we will
1 create a repl_dev, then we have a raft group
2 replicate a create_pg message across the raft group. when on_commit of this message is called, pg superblk and index table will be built.

so pg is not equal to a repl_dev, it is only a application of repl_dev, and repl_dev probably have many different kind of application

JacksonYao287:
@yamingk  I have change the code to join raft group concurrently, so that one repl_dev now no-longer need to wait for the completion of the previous one. PTAL!

pls note that, according to the design of chunk level data copy, gc recovery will be very fast since what recovery does is only updating the index table again, no data copy will be involved.

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -156,17 +156,20 @@ void RaftReplService::start() {
...
-        auto rdev = std::dynamic_pointer_cast< RaftReplDev >(it->second);
-        rdev->wait_for_logstore_ready();
-        if (!rdev->join_group()) {
-            HS_REL_ASSERT(false, "FAILED TO JOIN GROUP, PANIC HERE");
-            it = m_rd_map.erase(it);
-        } else {
-            ++it;
-        }
-    }
+    // Step 6: Iterate all the repl devs and ask each one of them to join the raft group concurrently.
```
yamingk, line 159: Looks fine to me to change to join raft group concurrently. Since this is some kind of behavior change, it probably worth to mention to the team and be aware of any testing outcomes.

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -156,17 +156,20 @@ void RaftReplService::start() {
...
+    std::vector< std::future< bool > > join_group_futures;
+    for (const auto& [_, repl_dev] : m_rd_map)
```
yamingk, line 161: nit: would be nice to put {...} for the `for` loop even thought it is one line. We had a discussion long back that data path code would like to establish this coding convention to avoid making mistakes when more lines needs to be added inside the for loop in the future. 

Similar comment for line: 171

src/include/homestore/replication/repl_dev.h
yamingk: Below comment is irrelevant to the concurrent joint raft group feature you added. E.g. we can still have that regardless of this discussion.

| >>> I do get your point. however, from the perspective of homestore itselft, it is unaware of PG, which is a homeobject specific concept. 

Valid point, my suggestion is actually aligned with your thought here about homestore is not aware of PG. It just enables homestore consumers, here in this case, the PG (again not aware to HomeStore), by providing consumer a way to inform HomeStore that: 
--- Hey, I (HomeObj) am doing some unfinished business with my upper object (PG), I want to tell you that please skip this object's underlaying resource (in this case the raft repl dev), and I will let you know when it is ready to continue the business. 

This would provide HomeStore consumer's to control the way how they want to boot HomeStore. 

Imagine we have a malfunctioning PG, and we want to move that out or this PG somehow hit some corruption that it can not be able to remove from the raft, we can set a flag (offline, read-only, skip, whatever we can call it), but HomeStore still is not aware of PG, it is just some additional functionality that Concumer can use for flexibility and error handling. 

Does it make sense?

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -156,17 +156,20 @@ void RaftReplService::start() {
...
-        rdev->wait_for_logstore_ready();
-        if (!rdev->join_group()) {
-            HS_REL_ASSERT(false, "FAILED TO JOIN GROUP, PANIC HERE");
-            it = m_rd_map.erase(it);
-        } else {
-            ++it;
-        }
-    }
+    // Step 6: Iterate all the repl devs and ask each one of them to join the raft group concurrently.
+    std::vector< std::future< bool > > join_group_futures;
```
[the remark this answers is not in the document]
JacksonYao287: if we use folly future, we need to create a thread pool for the execution of these futures.
 
 1  folly::InlineExecutor::instance() will run all the futures in the caller threads, which means all the futures will be executed sequentially, not asynchronously.
 
 2 folly::IOThreadPoolExecutor and CPUThreadPoolExecutor will create a separate thread pool. since we will use them only once, so it is not worth to create a thread pool for the one-time shot.
 
 so, I use std::async, which will run the future asynchronously and not create a thread pool for the one-time shot

src/lib/replication/service/raft_repl_service.cpp
JacksonYao287: ack

src/lib/replication/service/raft_repl_service.cpp
JacksonYao287: ack

src/include/homestore/replication/repl_dev.h
JacksonYao287: right , it makes sense. can we make this change in a separate PR?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -335,7 +335,15 @@ void RaftReplDev::async_alloc_write(sisl::blob const& header, sisl::blob const&
             handle_error(rreq, ReplServiceError::DATA_DUPLICATED);
             return;
         }
+
+#ifdef _PRERELEASE
+        if (iomgr_flip::instance()->test_flip("disable_leader_push_data")) {
+            RD_LOGD("Simulating push data failure, so that all the follower will have to fetch data");
+        } else
```
yamingk, line 342: please also put a {} around else

yamingk (approved):

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: ack， will do it in later pr

PR #661, opened 2025-03-13
[TODO] used if there is a concurrency issues between gc and flush

Besroy:
In previous implementation, there may be a concurrency issue between 'permanent_destroy' and 'cp_flush/flush_durable_commit_lsn': if m_rd_sb is destroyed at T1 then is written at T2, a NPE will occur. 
This pr is a fix way to avoid this issue by adding m_sb_mtx for destroy.

In addition to the current fix, there are two other potential approaches to address this issue:
1. support triggering specific CP: so that HO's pg_destroy won't trigger repl_dev's cp_flush, which acquiring m_rd_map's lock
2. refine leave_group func: modify the leave_group within the nuraft_mesg component to only shut down servers. so that gc can call leave_group and upper destroy without lock, call permanent_destroy and remove it from map with lock


xiaoxichen (approved):

xiaoxichen:
thanks for the effort to refine the edge case.

xiaoxichen:
@Besroy can you check if this was fixed or do we still need this?

Besroy:
> @Besroy can you check if this was fixed or do we still need this?

We still need this, but I haven't encountered this corner case. Let me close this PR and reopen it if it occurs again

PR #662, opened 2025-03-17
Send response if failed to save pushed_data

Besroy:
The previous implementation missed sending a response when storing pushed_data failed. As a result, a large amount of RPC data may be held in memory and only released when the connection timed out, leading to increased memory usage. This change fixes the issue.

Here is the background:
The issue occurs in test case `full pg recovery in 5 replicas env, blob_size=512KB`, when a batch of old shards sealed and a batch of new shards created, a lot of 'pushed_data' failed to save due to shards not committed. Then the memory increased and the pod is OOM in the test.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.7.2"
+    version = "6.7.3"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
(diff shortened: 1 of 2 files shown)


xiaoxichen (approved):

xiaoxichen:
Could you also share more background on which testcases help you identify this isue.

Besroy:
> Could you also share more background on which testcases help you identify this isue.

Sure. The issue occurs when I testing "full pg recovery in 5 replicas env, blob_size=512KB", when a batch of old shards sealed and a batch of new shards created,  a lot of 'pushed_data' failed to save due to shards not committed. Then the memory increased and the pod is OOM in my test. More details can be seen in [issue38](https://docs.google.com/document/d/16DMqv9J-JuNs5c25IBuX01QXe5WbxPm4B4hAfz4JNZM/edit?tab=t.0#heading=h.ry5ucurpeiez)

PR #672, opened 2025-03-21
Async IO metrics for physical dev.

xiaoxichen:


src/lib/device/physical_dev.cpp
```diff
@@ -35,6 +35,8 @@ namespace homestore {
 static std::mutex s_cached_dev_mtx;
 static std::unordered_map< std::string, iomgr::io_device_ptr > s_cached_opened_devs;
 
+__attribute__((no_sanitize_address)) static auto get_current_time() { return Clock::now(); }
```
JacksonYao287, line 38: why we need this attribute?

xiaoxichen: I dont know, it was there in existing code. I just move it up.

JacksonYao287:
LGTM. @yamingk  can you also take a look?

src/lib/device/physical_dev.cpp
yamingk: I guess it is to make sanitize run happy?

yamingk (approved):

yamingk:
LGTM

PR #674, opened 2025-03-24
Wait for the crash only if the flip is set and triggered in index crash test

raakella1:


src/lib/common/crash_simulator.hpp
```diff
@@ -38,8 +38,12 @@ class CrashSimulator {
         }
     }
 
+    bool will_crash() const { return m_will_crash.load(); }
+    void set_will_crash(bool crash) { m_will_crash.store(crash); }
+
 private:
     std::function< void(void) > m_restart_cb{nullptr};
+    std::atomic<bool> m_will_crash{false};
```
shosseinimotlagh, line 46: set it to false as default makes other tests not to crash and restart if they suppose to.

raakella1: Any test that is supposed to crash should call set_will_crash(true).

shosseinimotlagh (approved):
LGTM

p.s: I am not sure if other services use this crash simulator. I assume that you already checked them.

raakella1:
> LGTM
> 
> p.s: I am not sure if other services use this crash simulator. I assume that you already checked them.

Currently, index crash is the only test waiting on the crash

shosseinimotlagh (approved):

PR #676, opened 2025-03-26
 Fix baseline resync corner cases.

Besroy:
1. Leader side: Deny snapshot read if there are uncommitted logs in the snapshot. This prevents the following scenario:
 If a crash occurs during snapshot creation, the snapshot might be persisted while the rd sb is not. This means the durable_commit_lsn is less than the snapshot's log_idx. Upon restart, the changes in uncommitted logs may or may not be included in the snapshot data sent by the leader, depending on the race condition between commit and snapshot read, leading to data inconsistency.
 2. Follower side: Skip replay and commit when BR is in progress and purge logs is no longer supported.
 Purging logs can cause issues such as the commit thread being unable to access logs if they are purged. This change removes the purge logic and adds a new flag to skip replay/commit to avoid log ops accessing unavailable resources after the PG is destroyed by BR.

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -69,6 +69,7 @@ RaftReplDev::RaftReplDev(RaftReplService& svc, superblk< raft_repl_dev_superblk 
         m_rd_sb->logstore_id = m_data_journal->logstore_id();
         m_rd_sb->last_applied_dsn = 0;
         m_rd_sb->destroy_pending = 0x0;
+        m_rd_sb->last_snapshot_lsn = 0;
         m_rd_sb->group_ordinal = s_next_group_ordinal.fetch_add(1);
         m_rdev_name = fmt::format("rdev{}", m_rd_sb->group_ordinal);
 
@@ -1541,6 +1542,11 @@ void RaftReplDev::set_log_store_last_durable_lsn(store_lsn_t lsn) { m_data_journ
 
 void RaftReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx) {
     auto repl_lsn = to_repl_lsn(lsn);
+    if (need_skip_processing(repl_lsn)) {
+        RD_LOGI("Raft Channel: Log {} is outdated and will be handled by baseline resync. Ignoring replay.", lsn);
+        return;
+    }
+
     // apply the log entry if the lsn is between checkpoint lsn and durable commit lsn
     if (repl_lsn <= m_rd_sb->checkpoint_lsn) { return; }
 
@@ -1635,7 +1641,7 @@ void RaftReplDev::create_snp_resync_data(raft_buf_ptr_t& data_out) {
     std::memcpy(data_out->data_begin(), &msg, msg_size);
 }
 
-bool RaftReplDev::save_snp_resync_data(nuraft::buffer& data) {
+bool RaftReplDev::save_snp_resync_data(nuraft::buffer& data, nuraft::snapshot& s) {
     auto msg = r_cast< snp_repl_dev_data* >(data.data_begin());
     if (msg->magic_num != HOMESTORE_RESYNC_DATA_MAGIC ||
         msg->protocol_version != HOMESTORE_RESYNC_DATA_PROTOCOL_VERSION_V1) {
@@ -1651,6 +1657,14 @@ bool RaftReplDev::save_snp_resync_data(nuraft::buffer& data) {
     if (received_crc != computed_crc) {
         RD_LOGE("Snapshot resync data crc mismatch, received_crc={}, computed_crc={}", received_crc, computed_crc);
         return false;
+    }
+    {
+        // Save last_snapshot_lsn, so that we can skip the replay/commit operation for logs included in baseline resync.
+        // The reason is baseline resync will clear existing resources on the upper layer, skipping replay/commit
+        // operations can avoid accessing unavailable resources
+        std::unique_lock lg{m_sb_mtx};
+        m_rd_sb->last_snapshot_lsn = s_cast< repl_lsn_t >(s.get_last_log_idx());
+        m_rd_sb.write();
     }
     if (msg->dsn > m_next_dsn) {
         m_next_dsn = msg->dsn;
--- a/src/lib/replication/repl_dev/raft_repl_dev.h
+++ b/src/lib/replication/repl_dev/raft_repl_dev.h
@@ -25,6 +25,7 @@ struct raft_repl_dev_superblk : public repl_dev_superblk {
     uint8_t is_timeline_consistent; // Flag to indicate whether the recovery of followers need to be timeline consistent
     uint64_t last_applied_dsn;      // Last applied data sequence number
     uint8_t destroy_pending;        // Flag to indicate whether the group is in destroy pending state
+    repl_lsn_t last_snapshot_lsn;   // Last snapshot LSN follower received from leader
 
     uint32_t get_raft_sb_version() const { return raft_sb_version; }
 };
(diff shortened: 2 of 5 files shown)


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -283,6 +283,12 @@ void RaftReplDev::use_config(json_superblk raft_config_sb) { m_raft_config_sb =
 
 void RaftReplDev::on_create_snapshot(nuraft::snapshot& s, nuraft::async_result< bool >::handler_type& when_done) {
     RD_LOG(DEBUG, "create_snapshot last_idx={}/term={}", s.get_last_log_idx(), s.get_last_log_term());
+    // Ensure all metadata is flushed to disk before creating the snapshot to prevent the following scenario:
+    // If a crash occurs during snapshot creation, the snapshot might be persisted while the rd's sb is not.
+    // This means the durable_commit_lsn is less than the snapshot's log_idx. Upon restart, uncommitted logs may be
```
xiaoxichen, line 288: may not be? Not sure I understand correctly.
The snapshot@ 150 while dc_lsn=100. Upon restart, we commit up to 100 then start the statemachine. BR will restart however 100-150 may not included in the BR, for example all are in the first shard.

Besroy: Unfortunately, BR will use snapshot @150. Currently, BR saves the snapshot context to disk, and at the beginning of BR, the leader will try to reload it first. If snapshot @150 has been persisted to disk, even though dc_lsn is at 100, entries from 101 to 150 will be included in the BR. 
However, in my test last night, adding a cp_flush in create_snapshot resulted in a delay of approximately 1-3 seconds, which could slow down the commit process. I'm uncertain if this will affect user TPS. Therefore, this approach might not be ideal. I may need to discuss and find a more suitable solution. Also cc @yuwmao @koujl

xiaoxichen: I think we are taking about same thing? 

if snapshot @150 is used, that snapshot ,should contains data at least up to LSN 150 , meaning,  LSN 1-150 should be present, LSN >150 can be present.

The problem is LSN 100-150 is not yet committed after restart, so BR when reading the snapshot,  it can missed the 100-150 as they are not yet added to index.

Besroy: Yes, but if lsn 90 is put as blob=1, then blob=1's index is added to the index. When lsn 110 attempts to delete blob=1, since lsn 110 is not committed, blob=1's index is not removed from the index table. After the snapshot, the follower starts committing from lsn 151, so the index doesn't get a chance to be deleted. That is why [SH test read deleted blob successfully](https://docs.google.com/document/d/16DMqv9J-JuNs5c25IBuX01QXe5WbxPm4B4hAfz4JNZM/edit?tab=t.0#heading=h.pnlm9c2x64c2). 
Additionally, if lsn 100-150 have not yet been committed after a restart, they might not be available on follower since they haven't had the chance to be committed as well🤔

xiaoxichen: yes,  can the statement be " the changes in LSN 100-150 may or may not included in the snapshot data sent by leader, depending on the racing of commit vs snapshot read".

xiaoxichen:
If we think cp in create_snapshot is too heavy (as most of the snapshot will not be used).  Can we do that when we read the snapshot checking the ` committed_lsn >= snapshot_lsn`?

src/lib/replication/repl_dev/raft_repl_dev.cpp
Besroy: Updated, thanks for the correction~ Besides, since cp_flush includes a delay of approximately 1-3 seconds, which could slow down the commit process, do we need to find another way to resolve this issue?

Besroy:
> If we think cp in create_snapshot is too heavy (as most of the snapshot will not be used). Can we do that when we read the snapshot checking the ` committed_lsn >= snapshot_lsn`?

Thanks I think this is a good resolution. Let me try to change and test it in SH.

src/lib/replication/repl_dev/raft_repl_dev.h
```diff
@@ -231,9 +232,9 @@ class RaftReplDev : public ReplDev,
         if (!ready) { RD_LOGD("Not yet ready for traffic, committed to {} but gate is {}", committed_lsn, gate); }
         return ready;
     }
+    // purge all resources (e.g., logs in logstore) is a very dangerous operation, it is not supported yet.
     void purge() override {
-        // clean up existing logs in log store
-        m_data_journal->purge_all_logs();
+        RD_LOGW("Not Supported yet");
```
xiaoxichen, line 237: better assert :)

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1652,6 +1658,10 @@ bool RaftReplDev::save_snp_resync_data(nuraft::buffer& data) {
         RD_LOGE("Snapshot resync data crc mismatch, received_crc={}, computed_crc={}", received_crc, computed_crc);
         return false;
     }
+    // Set baseline_resyncing to 1, so that we can skip the replay/commit operation during baseline resync.
+    // The reason is baseline resync will clear existing resources on the upper layer, skipping replay/commit operations
+    // can avoid accessing unavailable resources
+    m_rd_sb->baseline_resyncing = 0x1;
```
xiaoxichen, line 1664: What we need to guarantee is before we destroying pg , this flag has been persisted.
It *seems*  satisfied as pg_destroy will call cp and that cp , upon success, will flush this superblock. 

However I am not sure this is the right call.  As CP is not consistent across component ,  it is possible some of the subsystem finished cp but others are not. If unfortunately crashed before repl_dev cp, will there be any issue?

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -373,6 +394,8 @@ void RaftStateMachine::save_logical_snp_obj(nuraft::snapshot& s, ulong& obj_id,
 
     m_rd.m_listener->write_snapshot_obj(snp_ctx, snp_data);
     if (is_last_obj) {
+        // reset baseline_resyncing flag since all snapshot data has been saved
+        m_rd.m_rd_sb->baseline_resyncing = 0x0;
```
yuwmao, line 398: I am considering it would be better to clear it when compact log after snapshot, because if it crashes between handling last obj and compact log, the outdated logs still have chance to be replayed.

Besroy: Yeah I'm thinking about it as well 🤔 If the last object has been saved, outdated replays might not encounter any problems, possibly due to the shard meta being rebuilt. However, if the flag is reset after compaction, there's a risk that it might not get reset after a crash.

src/lib/replication/repl_dev/raft_repl_dev.cpp
Besroy: Yes you're right. There is a risk that shard meta be deleted before the flag is persisted. I added a write op after the flag is set to ensure it is persisted.

src/lib/replication/repl_dev/raft_state_machine.cpp
yuwmao: Yes, if it crashed after last obj is saved, outdated replays won't encounter assert failure, but I'm thinking about this case:
If a blob is written at lsn 10 and deleted at lsn15, snapshot lsn=20, lsn[8-11] are not committed before snapshot.
During recovery, put blob will be replayed and then the blob is leaked.

Besroy: Thanks for the corner case, I updated flag to snapshot lsn and all logs less/equal to it will be skipped replay/commit.

yuwmao:
LGTM

xiaoxichen (approved):

PR #680, opened 2025-04-02
Support blk reservation

yuwmao:
Application has an option to reserve some blks in a chunk when allocation.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.7.7"
+    version = "6.7.8"
 
     homepage = "https://github.com/eBay/Homestore"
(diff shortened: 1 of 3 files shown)


xiaoxichen (approved):

Besroy:
Sorry, I have a small question. If space is reserved at the HS level, does this mean that followers also reserve this space? In that case, when a follower has garbage, would it still trigger a NO_SPACE_LEFT condition, causing the PG state to become unclean? Or will there be some modifications at the upper layer to prevent this issue?

xiaoxichen:
It has to be done in HS as from HO's point of view,  if it get a multiBlk it has no way to know how much space still available on the chunk ---  this is specific to allocator. 

But what you said makes sense, the reservation is done through allocation_hint CB.  We aware we should set no reservation for seal shard,  thanks for reminding we need to consider the role here

PR #681, opened 2025-04-02
Fix solo repl dev log flush and graceful shutdown.

sanebay:
Add flush mode to logdev as nublocks uses timer,
nuobject uses explicit log flush mode. Flush mode has to be stored in superblk to support recovery. Enable solo repl dev UT. Add graceful shutdown for UT to work.

### DIFF
--- a/src/lib/logstore/log_store_service.cpp
+++ b/src/lib/logstore/log_store_service.cpp
@@ -142,12 +142,12 @@ logdev_id_t LogStoreService::get_next_logdev_id() {
     return id;
 }
 
-logdev_id_t LogStoreService::create_new_logdev() {
+logdev_id_t LogStoreService::create_new_logdev(flush_mode_t flush_mode) {
     if (is_stopping()) return 0;
     incr_pending_request_num();
     folly::SharedMutexWritePriority::WriteHolder holder(m_logdev_map_mtx);
     logdev_id_t logdev_id = get_next_logdev_id();
-    auto logdev = create_new_logdev_internal(logdev_id);
+    auto logdev = create_new_logdev_internal(logdev_id, flush_mode);
     logdev->start(true /* format */, m_logdev_vdev);
     COUNTER_INCREMENT(m_metrics, logdevs_count, 1);
     HS_LOG(INFO, logstore, "Created log_dev={}", logdev_id);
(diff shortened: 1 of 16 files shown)


src/lib/logstore/log_store_service.cpp
```diff
@@ -189,19 +189,19 @@ void LogStoreService::delete_unopened_logdevs() {
...
+std::shared_ptr< LogDev > LogStoreService::create_new_logdev_internal(logdev_id_t logdev_id, flush_mode_t flush_mode) {
+    auto logdev = std::make_shared< LogDev >(logdev_id, flush_mode);
     const auto it = m_id_logdev_map.find(logdev_id);
     HS_REL_ASSERT((it == m_id_logdev_map.end()), "logdev id {} already exists", logdev_id);
     m_id_logdev_map.insert(std::make_pair<>(logdev_id, logdev));
     return logdev;
 }
 
-void LogStoreService::open_logdev(logdev_id_t logdev_id) {
+void LogStoreService::open_logdev(logdev_id_t logdev_id, flush_mode_t flush_mode) {
```
yamingk, line 200: Isn't we persisted flush_mode in logdev's metablk, is there a reason for open_logdev to take this input, can we get it from the metablk, it would be simpflied if we can get it from one source, right?

src/lib/replication/log_store/home_raft_log_store.cpp
```diff
@@ -101,7 +101,7 @@ HomeRaftLogStore::HomeRaftLogStore(logdev_id_t logdev_id, logstore_id_t logstore
         m_logdev_id = logdev_id;
         m_logstore_id = logstore_id;
         LOGDEBUGMOD(replication, "Opening existing home log_dev={} log_store={}", m_logdev_id, logstore_id);
-        logstore_service().open_logdev(m_logdev_id);
+        logstore_service().open_logdev(m_logdev_id, flush_mode_t::EXPLICIT);
```
yamingk, line 104: Is there a reason we can't get flush_mode from the meta blk when create_new_logdev is called? If so, we should be able to avoid change to this open_logdev for all places, yeah?

yamingk:
LGTM, except the comment around open_logdev to see if we can even simplify, thanks!

src/lib/logstore/log_store_service.cpp
sanebay: The reason was logdev_superblk_found should be called earlier and create a logdev based on the metablk as you mentioned. But here open_logdev will also create if it couldnt find. I guess this was done due to RaftReplDev, SoloReplDev and other modules open logdev in constructor during bootstrap while the metablk callbacks are being called during start.

yamingk: I see your point. Ideally open_log_dev should only be called after logdev_superblk_found. I guess we should apply metablk dependency around raftrepldev, e.g. raftrepl dev metablk should only be received after logdev metablk has been callbacked. If you don't have cycles to try this, could you help open an issue and move on?

yamingk (approved):

src/lib/logstore/log_store_service.cpp
xiaoxichen: There are two part of this IMO,

1.  the logdev is opened when we constructing HomeRaftLogStore, which happens when we recovering the raft_repl_dev.   Those un_opened  (meaning, not used) logdev will de zapped afterward.  I believe this is to capture those leaked logdev during creating/deleting repl_dev.

2. Have we ever hit into the situation that when `open_logdev` the logdev is not exists?   I am not 100% sure but I believe there is no such case. As said in #1  the open_logdev is called after all metasvc recovery. ----  we can add an debug_assert and see.

xiaoxichen (commented):
I am curious why the flush_mode needs to be persisted? 

The flush mode controlled the flushing behavior when appending logs,  once the log dev closed then re-opened,  the previous flush_mode should not take any forward impact right?  Meaning that  nublox can use explicit then in one tick we change the config file to use timer,  after a restart the timer can works and effect on new appending logs. 

What is the benefit of remember "this log dev was created with explicit flushing" ?   

And if we believe there is a racing between `logdev_superblk_found` and  `open_logdev`,  then there is a ambiguity whether the mode provided by `open_logdev` take effect, or the mode in sb take effect.

src/lib/logstore/log_store_service.cpp
sanebay: I will create an issue for dependency for raft repldev after logdev.

src/lib/replication/log_store/home_raft_log_store.cpp
sanebay: As discussed in previous comment, created an issue for it.

sanebay:
> I am curious why the flush_mode needs to be persisted?
> 
> The flush mode controlled the flushing behavior when appending logs, once the log dev closed then re-opened, the previous flush_mode should not take any forward impact right? Meaning that nublox can use explicit then in one tick we change the config file to use timer, after a restart the timer can works and effect on new appending logs.
> 
> What is the benefit of remember "this log dev was created with explicit flushing" ?
> 
> And if we believe there is a racing between `logdev_superblk_found` and `open_logdev`, then there is a ambiguity whether the mode provided by `open_logdev` take effect, or the mode in sb take effect.

This is to support logdev_superblk_found to open logdev with flush mode.

yamingk (approved):

PR #682, opened 2025-04-02
Support handling no_space_left error in raft channel

JacksonYao287:
implementation details, pls refer to 

https://docs.google.com/document/d/1lgmWrixw5jOQz1oy0EI8LoLVMRwi_1l-6RkIRCDeJwU/edit?tab=t.0#heading=h.ymzurj2bm995

### DIFF
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -142,7 +142,6 @@ public:
     repl_journal_entry const* journal_entry() const { return m_journal_entry; }
     uint32_t journal_entry_size() const;
     bool is_localize_pending() const { return m_is_jentry_localize_pending; }
-    bool is_data_inlined() const { return (m_op_code == journal_type_t::HS_DATA_INLINED); }
     bool has_linked_data() const { return (m_op_code == journal_type_t::HS_DATA_LINKED); }
 
     raft_buf_ptr_t& raft_journal_buf();
@@ -336,7 +335,8 @@ public:
     /// @return Expected to return blk_alloc_hints for this write. If the hints are not available, then return the
     /// error. It is to be noted this method should return error only in very abnornal cases as in some code flow, an
     /// error would result in a crash or stall of the entire commit thread.
-    virtual ReplResult< blk_alloc_hints > get_blk_alloc_hints(sisl::blob const& header, uint32_t data_size, cintrusive< homestore::repl_req_ctx >& hs_ctx) = 0;
+    virtual ReplResult< blk_alloc_hints > get_blk_alloc_hints(sisl::blob const& header, uint32_t data_size,
+                                                              cintrusive< homestore::repl_req_ctx >& hs_ctx) = 0;
 
     /// @brief Called when the repl_dev is being destroyed. The consumer is expected to clean up any related resources.
     /// However, it is expected that this call be idempotent. It is possible in rare scenarios that this can be called
@@ -383,12 +383,12 @@ public:
     }
 
     /// @brief ask upper layer to handle no_space_left event
-    virtual folly::Future< std::error_code > on_no_space_left(uint32_t pdev_id, chunk_num_t chunk_id) {
-        return folly::makeFuture< std::error_code >(std::error_code{});
-    }
+    // @param lsn - on which repl_lsn no_space_left happened
+    // @param chunk_id - on which chunk no_space_left happened
+    virtual void on_no_space_left(repl_lsn_t lsn, chunk_num_t chunk_id) { return; }
 
     /// @brief when restart, after all the logs are replayed and before joining raft group, notify the upper layer
-    virtual void on_log_replay_done(const group_id_t& group_id) {};
+    virtual void on_log_replay_done(const group_id_t& group_id){};
 
 private:
     std::weak_ptr< ReplDev > m_repl_dev;
@@ -469,6 +469,10 @@ public:
     /// @return last_commit_lsn
     virtual repl_lsn_t get_last_commit_lsn() const = 0;
 
+    /// @brief Gets the repl lsn of the last log in log store
+    /// @return last_append_repl_lsn
+    virtual repl_lsn_t get_last_append_lsn() = 0;
+
     /// @brief if this replica is ready for accepting client IO.
     /// @return true if ready, false otherwise
     virtual bool is_ready_for_traffic() const = 0;
@@ -499,6 +503,16 @@ public:
             std::this_thread::sleep_for(std::chrono::milliseconds(1000));
         }
     }
+
+    // pause/resume statemachine(commiting thread)
+    virtual void pause_statemachine() = 0;
+    virtual void resume_statemachine() = 0;
+
+    virtual void enter_emergency() = 0;
+    virtual void leave_emergency() = 0;
+
+    // clear reqs that has allocated blks on the given chunk.
+    virtual void clear_chunk_req(chunk_num_t chunk_id) = 0;
 
 protected:
     shared< ReplDevListener > m_listener;
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -508,8 +508,11 @@ public:
     virtual void pause_statemachine() = 0;
     virtual void resume_statemachine() = 0;
 
-    virtual void enter_emergency() = 0;
-    virtual void leave_emergency() = 0;
+    // complete all the requests that are in progress and start refusing new reqs
+    virtual void quiesce_reqs() = 0;
+
+    // start accepting new reqs
+    virtual void resume_accepting_reqs() = 0;
 
     // clear reqs that has allocated blks on the given chunk.
     virtual void clear_chunk_req(chunk_num_t chunk_id) = 0;
--- a/src/lib/replication/repl_dev/common.cpp
+++ b/src/lib/replication/repl_dev/common.cpp
@@ -10,8 +10,9 @@
 
 namespace homestore {
 
-ReplServiceError repl_req_ctx::init(repl_key rkey, journal_type_t op_code, bool is_proposer, sisl::blob const& user_header,
-                        sisl::blob const& key, uint32_t data_size, cshared< ReplDevListener >& listener) {
+ReplServiceError repl_req_ctx::init(repl_key rkey, journal_type_t op_code, bool is_proposer,
+                                    sisl::blob const& user_header, sisl::blob const& key, uint32_t data_size,
+                                    cshared< ReplDevListener >& listener) {
     m_rkey = std::move(rkey);
 #ifndef NDEBUG
     if (data_size > 0) {
@@ -26,17 +27,34 @@ ReplServiceError repl_req_ctx::init(repl_key rkey, journal_type_t op_code, bool 
     m_key = key;
     m_is_jentry_localize_pending = (!is_proposer && (data_size > 0)); // Pending on the applier and with linked data
 
-    // We need to allocate the block if the req has data linked, since entry doesn't exist or if it exist, two threads(data channel and raft channel) are trying to do the same
-    // thing. So take state mutex and allocate the blk
+    // We need to allocate the block if the req has data linked, since entry doesn't exist or if it exist, two
+    // threads(data channel and raft channel) are trying to do the same thing. So take state mutex and allocate the blk
     std::unique_lock< std::mutex > lg(m_state_mtx);
     if (has_linked_data() && !has_state(repl_req_state_t::BLK_ALLOCATED)) {
-        auto alloc_status = alloc_local_blks(listener, data_size);
+        ReplServiceError alloc_status;
+#ifdef _PRERELEASE
+        if (iomgr_flip::instance()->test_flip("fake_no_space_left") && !is_proposer) {
+            LOGERROR("Simulate no space left on follower for testing purposes");
+            // TODO: support `fake_no_space_left` for the leader, do not throw exception in on-error in the test
+            // framework, it will cause the leader to fail and exit.
+            alloc_status = ReplServiceError::NO_SPACE_LEFT;
+        } else {
+            alloc_status = alloc_local_blks(listener, data_size);
+            if (alloc_status != ReplServiceError::OK) {
+                LOGERRORMOD(replication, "[traceID={}] Allocate blk for rreq failed error={}", m_rkey.traceID,
+                            alloc_status);
+            }
+        }
+#else
+        alloc_status = alloc_local_blks(listener, data_size);
         if (alloc_status != ReplServiceError::OK) {
             LOGERRORMOD(replication, "[traceID={}] Allocate blk for rreq failed error={}", m_rkey.traceID,
                         alloc_status);
         }
+#endif
         return alloc_status;
     }
+
     return ReplServiceError::OK;
 }
 
@@ -107,8 +125,9 @@ ReplServiceError repl_req_ctx::alloc_local_blks(cshared< ReplDevListener >& list
     if (hints_result.hasError()) { return hints_result.error(); }
 
     if (hints_result.value().committed_blk_id.has_value()) {
-        //if the committed_blk_id is already present, use it and skip allocation and commitment
-        LOGINFOMOD(replication, "[traceID={}] For Repl_key=[{}] data already exists, skip", rkey().traceID, rkey().to_string());
+        // if the committed_blk_id is already present, use it and skip allocation and commitment
+        LOGINFOMOD(replication, "[traceID={}] For Repl_key=[{}] data already exists, skip", rkey().traceID,
+                   rkey().to_string());
         m_local_blkid = hints_result.value().committed_blk_id.value();
         add_state(repl_req_state_t::BLK_ALLOCATED);
         add_state(repl_req_state_t::DATA_RECEIVED);
@@ -122,7 +141,8 @@ ReplServiceError repl_req_ctx::alloc_local_blks(cshared< ReplDevListener >& list
     auto status = data_service().alloc_blks(sisl::round_up(uint32_cast(data_size), data_service().get_blk_size()),
                                             hints_result.value(), m_local_blkid);
     if (status != BlkAllocStatus::SUCCESS) {
-        LOGWARNMOD(replication, "[traceID={}] block allocation failure, repl_key=[{}], status=[{}]", rkey().traceID, rkey(), status);
+        LOGWARNMOD(replication, "[traceID={}] block allocation failure, repl_key=[{}], status=[{}]", rkey().traceID,
+                   rkey(), status);
         DEBUG_ASSERT_EQ(status, BlkAllocStatus::SUCCESS, "Unable to allocate blks");
         return ReplServiceError::NO_SPACE_LEFT;
     }
--- a/src/lib/replication/repl_dev/common.cpp
+++ b/src/lib/replication/repl_dev/common.cpp
@@ -33,9 +33,9 @@ ReplServiceError repl_req_ctx::init(repl_key rkey, journal_type_t op_code, bool 
     if (has_linked_data() && !has_state(repl_req_state_t::BLK_ALLOCATED)) {
         ReplServiceError alloc_status;
 #ifdef _PRERELEASE
-        if (iomgr_flip::instance()->test_flip("fake_no_space_left") && !is_proposer) {
+        if (iomgr_flip::instance()->test_flip("simulate_no_space_left") && !is_proposer) {
             LOGERROR("Simulate no space left on follower for testing purposes");
-            // TODO: support `fake_no_space_left` for the leader, do not throw exception in on-error in the test
+            // TODO: support `simulate_no_space_left` for the leader, do not throw exception in on-error in the test
             // framework, it will cause the leader to fail and exit.
             alloc_status = ReplServiceError::NO_SPACE_LEFT;
         } else {
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -15,7 +15,6 @@
 
 #include "common/homestore_assert.hpp"
 #include "common/homestore_config.hpp"
-// #include "common/homestore_flip.hpp"
 #include "replication/service/raft_repl_service.h"
 #include "replication/repl_dev/raft_repl_dev.h"
 #include "device/chunk.h"
@@ -200,18 +199,27 @@ AsyncReplResult<> RaftReplDev::replace_member(const replica_member_info& member_
             members.replica_in = member_in;
 
             sisl::blob header(r_cast< uint8_t* >(&members), sizeof(replace_members_ctx));
-            rreq->init(repl_key{.server_id = server_id(),
-                                .term = raft_server()->get_term(),
-                                .dsn = m_next_dsn.fetch_add(1),
-                                .traceID = trace_id},
-                       journal_type_t::HS_CTRL_REPLACE, true, header, sisl::blob{}, 0, m_listener);
+            auto status = init_req_ctx(rreq,
+                                       repl_key{.server_id = server_id(),
+                                                .term = raft_server()->get_term(),
+                                                .dsn = m_next_dsn.fetch_add(1),
+                                                .traceID = trace_id},
+                                       journal_type_t::HS_CTRL_REPLACE, true, header, sisl::blob{}, 0, m_listener);
 
-            auto err = m_state_machine->propose_to_raft(std::move(rreq));
-            if (err != ReplServiceError::OK) {
-                RD_LOGE(trace_id, "Replace member propose to raft failed {}", err);
+            if (status != ReplServiceError::OK) {
+                // Failed to initialize the repl_req_ctx for replace member.
+                RD_LOGE(trace_id, "Failed to initialize repl_req_ctx for replace member, error={}", status);
                 reset_quorum_size(0, trace_id);
                 decr_pending_request_num();
-                return make_async_error<>(std::move(err));
+                return make_async_error<>(std::move(status));
+            }
+
+            status = m_state_machine->propose_to_raft(std::move(rreq));
+            if (status != ReplServiceError::OK) {
+                RD_LOGE(trace_id, "Replace member propose to raft failed {}", status);
+                reset_quorum_size(0, trace_id);
+                decr_pending_request_num();
+                return make_async_error<>(std::move(status));
             }
 
             RD_LOGI(trace_id, "Replace member proposed to raft group_id={}", group_id_str());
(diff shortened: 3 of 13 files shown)


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -538,23 +555,54 @@ repl_req_ptr_t RaftReplDev::applier_create_req(repl_key const& rkey, journal_typ
...
+
+            // the chunk_id is the chunk on which no_space_left happens
+            const auto& chunk_id = rreq->local_blkid().chunk_num();
+            const auto last_append_lsn = m_data_journal->next_slot() - 1;
+            const auto last_commit_lsn = get_last_commit_lsn();
+
+            RD_LOGD("no_space_left occurs in raft channel for chunk_id={}, last_append_lsn={}, last_commit_lsn={}",
+                    chunk_id, last_append_lsn, last_commit_lsn);
+
+            if (s_cast< repl_lsn_t >(last_append_lsn) == last_commit_lsn) {
```
xiaoxichen, line 579: i  think it should be `lsn == last_commit_lsn` ?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -538,23 +555,54 @@ repl_req_ptr_t RaftReplDev::applier_create_req(repl_key const& rkey, journal_typ
...
+                // handle error immediately if all the appended logs are committed
+                RD_LOGD("All logs are committed, handle no space left immediately in raft channel");
+                // we have to hanlde it here, since the last_commit_lsn has been committed and as a result,
+                // handle_no_space_left will not be triggered in committing thread.
+                handle_no_space_left();
+            } else if (m_no_space_left_error_info.lsn == -1 /*set a new error info*/ ||
```
xiaoxichen, line 585: initialize _no_space_left_error_info.lsn to max_int64 so we can unifiy the if?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -538,23 +555,54 @@ repl_req_ptr_t RaftReplDev::applier_create_req(repl_key const& rkey, journal_typ
...
+
     if (status != ReplServiceError::OK) {
-        RD_LOGD("For Repl_key=[{}] alloc hints returned error={}, failing this req", rkey.to_string(), status);
+        if (status == ReplServiceError::NO_SPACE_LEFT && !is_data_channel && !rreq->is_proposer()) {
+            RD_LOGD("For Repl_key=[{}] alloc hints returned error={} , lsn={}", rkey.to_string(), status, lsn);
+
+            // only when the follower got no_space_left error in raft channel, we try to handle it. if it is a leader,
+            // we will not handle it and return fail to client.
+
+            pause_statemachine();
```
xiaoxichen, line 567: why we want to pause state machine here? to avoid  the case 
```
last_commit_lsn = 99;
m_no_space_left_error_info.lsn  = 100;
```
however the commit of lsn 100 happened on commit thread alread?

xiaoxichen: If this is the case, setting `m_no_space_left_error_info.lsn  = 100;` prior to check last_commit_lsn probabaly fix this .

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1688,4 +1760,162 @@ bool RaftReplDev::is_resync_mode() {
...
+void RaftReplDev::set_no_space_left_error_info(repl_lsn_t lsn, chunk_num_t chunk_id) {
+    // this will be called only in raft channel for follower, since `RaftReplDev::raft_event` is always called in a
+    // single thread, so it is safe to call this method without lock
+    // TODO:: add lock if it might be modified by multiple threads in the future
+    RD_LOGD("set no_space_left error info: chunk {}, lsn {}", chunk_id, lsn);
+    m_no_space_left_error_info.lsn = lsn;
+    m_no_space_left_error_info.chunk_id = chunk_id;
+}
+
+// Note that, before calling this method, the caller should make sure that all the appended logs have been committed,
```
xiaoxichen, line 1788: this comment is very hard to follower

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1688,4 +1760,162 @@ bool RaftReplDev::is_resync_mode() {
...
+
+    HS_LOG_ASSERT(-1 != chunk_id, "chunk_id should be valid when handling no_space_left error, chunk_id={}, lsn={}",
+                  chunk_id, lsn);
+
+    RD_LOGD("start handling no_space_left error for chunk {} , lsn {}", chunk_id, lsn);
+
+    enter_emergency();
+
+    // 1. wait for all the pending reqs to be initialized and refuse later coming new requests
+    RD_LOGD("enter emergency state, waiting for all the pending req to be initialized");
```
xiaoxichen, line 1804: This part is confusing.  IIUC it is trying to ensure all req passed `if (is_in_emergency()) {` had allocated its space from allocator.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1688,4 +1760,162 @@ bool RaftReplDev::is_resync_mode() {
...
+            "no_space_left error",
+            chunk_id);
+    std::vector< folly::Future< folly::Unit > > futs;
+    for (auto& [key, rreq] : m_repl_key_req_map) {
+        if (rreq->has_state(repl_req_state_t::BLK_ALLOCATED)) {
+            auto blkid = rreq->local_blkid();
+            if (chunk_id == blkid.chunk_num()) {
+                // only clean the rreqs which has allocated blks on the emergent chunk
+                futs.emplace_back(
+                    std::move(data_service().async_free_blk(blkid).thenValue([this, &blkid, &key](auto&& err) {
```
xiaoxichen, line 1841: it makes sence to remove those key from `m_repl_key_req_map` as they contains legacy local_blk_id.
however I dont get the necessity of async_free_blk

src/include/homestore/replication/repl_dev.h
```diff
@@ -380,7 +379,7 @@ class ReplDevListener {
     }
 
     /// @brief ask upper layer to handle no_space_left event
-    virtual folly::Future< std::error_code > on_no_space_left(uint32_t pdev_id, chunk_num_t chunk_id) {
+    virtual folly::Future< std::error_code > on_no_space_left(chunk_num_t chunk_id) {
```
xiaoxichen, line 382: most of the callback are sync.  
 Why we want to make this callback async(return a future) then waiting the future in sync manner?

src/lib/replication/repl_dev/raft_repl_dev.h
```diff
@@ -380,6 +383,25 @@ class RaftReplDev : public ReplDev,
...
+    void reset_no_space_left_error_info();
+    void handle_no_space_left();
+
+    void enter_emergency() { m_in_emergency.store(true, std::memory_order_release); }
+    void leave_emergency() { m_in_emergency.store(false, std::memory_order_release); }
+    bool is_in_emergency() { return m_in_emergency.load(std::memory_order_acquire); }
+
+    void inc_pending_init_req_num() { m_pending_init_req_num.fetch_add(1, std::memory_order_acq_rel); }
+    void dec_pending_init_req_num() { m_pending_init_req_num.fetch_sub(1, std::memory_order_acq_rel); }
+    uint64_t get_pending_init_req_num() { return m_pending_init_req_num.load(std::memory_order_acquire); }
```
xiaoxichen, line 404: suggestion RAII pattern for these kind of pending_io counting
```
class OperationCounter {
public:
    OperationCounter() {
       counter.fetch_add(1)
    }

    ~OperationCounter() {
        counter.fetch_add(-1)
    }

    static int getCount() {
        return counter;
    }

private:
    static std::atomic_uint64_t counter;
};
```

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk, line 567: I fail to follow the reason to pause state machine in this error case. If it is to block I/O, we should populate the no space error to HomeStore consumer, in this case, the HomeObject and pause I/O from there. Kindly be noted, some use cases it is just a general no space error, reflecting the fact of the storage engine (HomeStore), how consumer wants to act on this error, the logic can happen on the consumer. In NuBlox case, fail on no space left is Okay, because there is no shard layout and between replicas, they don't need to be on same layout. it is a back pressure to HomeBlocks layer and populate all the way to client to fail the I/O.

yamingk (changes requested):
Please share the reason to pause statemachine at homestore side.

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: if all the appended logs are committed, we can say the state is clean and we can start handling no_space_left.

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: ack , will make this change

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: when we check `last_commit_lsn`, the committing thread is probably still running. which means, if we get `last_commit_lsn = 99`, lsn=100 is probably being committed but not completed. if the commit  of lsn 100 is completed between the two line codes you posted above, handle_no_space_left will not be triggered.

after pause statemachine, we can make sure there is no on-going commit , everything is static:
1 last_commit_lsn will not change , since the commit thread has been paused.
1 last_append_log will not change, since we are now here blocking the log channel.

only at this moment, we can go ahead to check the last_commit_lsn and change m_no_space_left_error_info or handle_no_space_left directly.

cc @yamingk

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: I have a method
```
ReplServiceError RaftReplDev::init_req_ctx(repl_req_ptr_t rreq, repl_key rkey, journal_type_t op_code, bool is_proposer,
                                           sisl::blob const& user_header, sisl::blob const& key, uint32_t data_size,
                                           cshared< ReplDevListener >& listener) {
    inc_pending_init_req_num();
    if (is_in_emergency()) {
        // In emergency state, reject any new requests.
        RD_LOGD("Rejecting new request in emergency state, rkey=[{}]", rkey.to_string());
        dec_pending_init_req_num();
        return ReplServiceError::EMERGENCY_STATE;
    }
    auto status = rreq->init(rkey, op_code, is_proposer, user_header, key, data_size, m_listener);
    dec_pending_init_req_num();
    return status;
}
```

all the block allocation happens in  `rreq->init`. so after we wait for all the pending req has been initialized we can make sure
1 all the pending reqs has allocated their blocks
2 no new pending reqs will be initialized again.

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: for nuobject , we don`t need the necessity of async_free_blk since we will reset the chunk. but for other case(for example , nublox), which not use append_blk_allocator, not sure if they need this. so I add it here to make sure.

src/lib/replication/repl_dev/raft_repl_dev.cpp
xiaoxichen: I think pausing the state machine is still way too heavy and as @yamingk said this is not a generic logic shares across product. 

IMO,  we can handle the error in HS by providing the LSN+ chunk to HO immediately,  which record a pending_gc task,  then it is HO that checks the pre-condition (committed_lsn == wait_for_lsn)  that calls the HS set_emergent.

src/lib/replication/repl_dev/raft_repl_dev.cpp
xiaoxichen: yes I got you , The logging and comment is confusing.  Your reply here is providing more background and better to rephrase it into code.

```
all the block allocation happens in rreq->init. so after we wait for all the pending req has been initialized we can make sure
1 all the pending reqs has allocated their blocks
2 no new pending reqs will be initialized again.
```

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: >I think pausing the state machine is still way too heavy

1 pausing state machine is not that heavy as that in our mind. what it does is just waiting for the current commit to be completed and stop committing new log.  

2 handle_no_space_left (if happens) takes much more time(probably involves disk IO) than pausing statemachine. 

3 this only happens **on the follower which encounters handle_no_space_left**. so , this will almost not block the entire raft group, and thus will not block client request.  it will be blocked only if the two followers both hit this error , which rarely happens.

so , from my perspective,  the overhead of pausing statemachine is acceptable.

cc @yamingk 

>this is not a generic logic shares across product. 

 actually , when writing this code, I also did not want to pause statemachine here. however, I can not find a graceful way to guarantee the correctness  of last_commit_lsn. let me think it a little more bit.

src/include/homestore/replication/repl_dev.h
JacksonYao287: may be we can use thenValue to do something ? or we want to wait in another thread?  this left the possibility ^^

src/lib/replication/repl_dev/raft_repl_dev.cpp
xiaoxichen: Lets rethink the whole flow, assuming we have no-space error on LSN 100

-we want to wait LSN 99 to be committed because the way we do GC rely on Index. This is application specific, if other product has different handling logic , they probably dont need to wait LSN 99.  So it doesnt make sense to put the behavior of "delay handling no_space error till LSN 99 committed" into HS.

- when LSN 99 reached,  we want a callback to application (HO) be called once and only once,  the only purpose of pausing state-machine and etc is to achieve this. However, this is also implementation choice, the "wait till committed to LSN X then do something" can be in HS or application, in commit thread or a separate thread.

- The argument that pausing statemachine is not expensive etc is valid for HO use case but not necessary true for other use cases.

src/include/homestore/replication/repl_dev.h
xiaoxichen: If this is the argument we should move all CB into future:)

Probably this express your hidden intent that the `on_no_space_left` will take a long time.

src/lib/replication/repl_dev/raft_repl_dev.h
JacksonYao287: we can not use static here, since  OperationCounter will be share by diffferent raft_repl_dev, OperationCounter should be valid for a particular raft_dev_dev

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: @yamingk @xiaoxichen  I do get your concern, and actually that is also mine.

well, the most important intent of pausing statemachine here is that I want handle_no_space_left to be definitely triggered if the precondition is achieved.

> which means, if we get last_commit_lsn = 99, lsn=100 is probably being committed but not completed. if the commit of lsn 100 is completed between the two line codes you posted above, handle_no_space_left will not be triggered in commit thread since lsn 100 has already been committed, and there is no new log should be committed.

above is the problem statement.  it is a little bit like the notification of condition variable. if we do not wait before notify, then we will lost that notify and we will never got notified. yes , I can use a separate thread to periodically check the lsn , if it matches, then start handling no_space_left. this definitely works, but looks really ugly for me. if you can accept this, I can create another pr to make this change.

>Lets rethink the whole flow

I think a more essential questions is what preconditon the upper layer want before handling no_space_left 

it is hard to determined, right? 

for nuobject, it wants all the appended logs are committed and clear all the req which has allocated blk on the problematic chunk

for nublox, it does not care if all the appended logs are committed and no need to clear all the req which has allocated on the problematic chunk

for nushare(maybe they want use homestore someday in the future),    it want clears all the req which has allocated blk on the problematic chunk and other chunks(which stores their snapshot), just my guess

generally speaking, precondition from upper layer is hard to predict. there are two choices:
1 make homestore very flexibility to support user-defined precondition, but this will make the code very complicated.

2 be a little bossy, if you want to use homestore, just follow my rules.


Let`s think through this a little bit more. waiting for your inputs

xiaoxichen: my proposal would be
- call upstream to handle nospace error immediately.
- upstream homeobj register a gc task, with dependency to lsn 99
- homeobj gc scheduling system check commit lsn>=99 before starting gc.
- when gc starts，call hs start_gc, which cleanup the rreq.

src/lib/replication/repl_dev/raft_repl_dev.h
xiaoxichen: I think the key point is not the static,  but more about encapsulate the inc_pending and dec_pending into RAII, similar to cp_guard().    There are too many places that need to do dec_pending() ,  not only here,  the graceful-shutdown has more....  the dec spread across every `else` is a concerning.

Sorry this is a bit off-topic.

JacksonYao287: done, change has been made @xiaoxichen , ptal

src/lib/replication/repl_dev/raft_repl_dev.h
```diff
@@ -192,6 +205,17 @@ class RaftReplDev : public ReplDev,
...
 
+    // this is used to track the latest no_space_left error. It means after we commit to lsn, we have to start handling
+    // no_space_left for the chunk(chunk_id)
+    struct no_space_left_error_info {
+        repl_lsn_t lsn;
+        chunk_num_t chunk_id;
+    } m_no_space_left_error_info{std::numeric_limits< repl_lsn_t >::max(), 0};
+
+    // pending create requests, including both raft and data channel
+    std::atomic_uint64_t m_pending_init_req_num;
```
yamingk, line 216: Can we explore to use sisl::object_coutner<repl_req_t>>?  Is there a reason we can't leverage that. Having a atomic for it seems a done job that sisl.

src/lib/replication/repl_dev/raft_repl_dev.h
```diff
@@ -149,6 +149,19 @@ class nuraft_snapshot_context : public snapshot_context {
...
+        init_req_counter(std::atomic_uint64_t& counter) : my_counter(counter) {
+            my_counter.fetch_add(1, std::memory_order_acq_rel);
+        }
+
+        ~init_req_counter() { my_counter.fetch_sub(1, std::memory_order_acq_rel); }
+
+    private:
+        std::atomic_uint64_t& my_counter;
+    };
+
```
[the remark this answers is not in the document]
JacksonYao287: if we use `std::atomic_uint64_t my_counter`,  not a reference, then each `counter` instance will have a separate `std::atomic_uint64_t my_counter`.  so a `counter` instance will have nothing to do with each other, which is not expected

here, we must make sure different `counter` instance will share a same `std::atomic_uint64_t`(here is m_pending_init_req_num), like shared_ptr. so when we get_pending_num, we can get the real pending number of all the `counter` instance.

src/lib/replication/repl_dev/raft_repl_dev.h
JacksonYao287: can you share where it is? I can not find

yamingk: https://github.com/eBay/sisl/blob/master/include/sisl/utility/obj_life_counter.hpp

Wait, checked the code, repl_context_t already inherits from sisl::ObjLifeCounter, is that not serve the purpose already: https://github.com/eBay/HomeStore/blob/master/src/include/homestore/replication/repl_dev.h#L112

Check the ObjLifeCounter member function, the ref_counter is RAII.

yamingk: Screenshot 2025-04-07 at 4 36 34 PM

src/lib/replication/repl_dev/raft_repl_dev.cpp
yamingk: This approach LGTM

src/lib/replication/repl_dev/raft_repl_dev.h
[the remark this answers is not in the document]
xiaoxichen: thx you are right

src/lib/replication/repl_dev/raft_repl_dev.h
JacksonYao287:
```
template < typename T >
struct ObjLifeCounter {
    ObjLifeCounter() {
        s_created.fetch_add(1, std::memory_order_relaxed);
        s_alive.fetch_add(1, std::memory_order_relaxed);
        s_type.m_dummy = 0; // To keep s_type initialized during compile time
    }

    virtual ~ObjLifeCounter() {
        assert(s_alive.load() > 0);
        s_alive.fetch_sub(1, std::memory_order_relaxed);
    }

    ObjLifeCounter(const ObjLifeCounter&) noexcept { s_alive.fetch_add(1, std::memory_order_relaxed); }
    static std::atomic< int64_t > s_created;
    static std::atomic< int64_t > s_alive;
    static ObjTypeWrapper< T > s_type;
};
```

we can see the code here, it use static member to remember created and alive instance. static means all the instance will share the same counter, like I said here ,  all the repl_dev will share this counter. 
https://github.com/eBay/HomeStore/pull/682#discussion_r2028003619

so I think it does not work for the case here, where we want a per repl_dev counter

JacksonYao287:
@yamingk @xiaoxichen I have updated this PR according to our agreement. the use pattern is as follows:

1 if no_space_left error happens in raft channel(log channel), we will call `listener->on_no_space_left(lsn, chunk_id)`, to notify upstream.

2 for nuobject, in on_no_space_left it will first call `repl_dev->pause_statemachine()` and check whether all the appended logs have been committed.

2.1  if yes, it will call `enter_emergency` to make all the pending req has been created and no new req to be created again.(no pending blk allocation and no new blk allocation.) then call `clear_chun_req` to clear all the necessary reqs. and then trigger an emergent gc.

2.2 if no, it will store a error_info(lsn, chunk). and resume_statemachine.

3 in `listener->on_commit()` it will check the error_info. if the commtted lsn is equal to the lsn stored in error_info, it will start handling this no_space_left error as above.

here, pausing/resuming statemachine and clear_chunk_req will all happen in homeobject level, so the changes here will not impact other homestore usecase (like nublox).

PATL!

xiaoxichen:
lgtm for HS side.   HO part still has several unknowns but we can discuss that later when we have an implementation. mainly
assuming LSN 100 is the one triggered NO-SPACE, last append log = 97.  
98, 99 ,100 are in the same batch of this attempt. 
Committed lsn = 95
1.  It is possible new log appened  after `2.2` in your description, so I believe we should wait for `LSN -1 = 99` instead of 97
2. how rollback get handled , say if a 97-100 get rolled back and re-written with irrelevant logs , is that safe to start GC when lsn 99 get committed? probably not.

JacksonYao287:
> It is possible new log appened  after `2.2` in your description, so I believe we should wait for `LSN -1 = 99` instead of 97

waiting for 97 or 99 are both ok , no correctness issues, as long as we guarantee **all the appended log are comitted** before we start gc the emergent chunk. 

gc 97 here will be more simple and efficient. it can be considered that we do emermget gc in advance.


> how rollback get handled , say if a 97-100 get rolled back and re-written with irrelevant logs , is that safe to start GC when lsn 99 get committed? probably not.

I guess here you mean the last appended log is 100, the last commit log is 97. 
let`s say we encounter the no_space_left when receiving lsn 101 in log channel and in chunk 1. after we pause state machine 
, we will find last_commit_lsn != last_append_lsn, then we will set an error info [100, 1], which means after we commit lsn 1, we need gc chunk 1. and then resume statemachine.

when rollback happens, it will call on_rollback for lsn 100. in `on_rollback`, we can check if the lsn in error info is equal or bigger than the rollbacked lsn, if yes, 
1 we  can just reset the error info to cancel the emergent gc
2 or , we also can just go on, the the new lsn 100 is committed, we do emregent gc(even if it is unnecessary now). no correctness issue.

after this PR is merged, I will create another PR at HO side for above logic. we can review and discuss in that PR.  
cc @yamingk

src/lib/replication/repl_dev/raft_repl_dev.h
```diff
@@ -199,6 +212,10 @@ class RaftReplDev : public ReplDev,
     static std::atomic< uint64_t > s_next_group_ordinal;
     bool m_log_store_replay_done{false};
 
+    // pending create requests, including both raft and data channel
+    std::atomic_uint64_t m_pending_init_req_num;
+    std::atomic< bool > m_in_emergency;
```
yamingk, line 217: I think we should replace emergency with something else. HomeStore shouldn't have idea about upper layer logic.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1744,4 +1760,113 @@ void RaftReplDev::report_blk_metrics_if_needed(repl_req_ptr_t rreq) {
...
+}
+
+void RaftReplDev::resume_statemachine() {
+    if (raft_server()->is_state_machine_execution_paused()) {
+        raft_server()->resume_state_machine_execution();
+        RD_LOGD(NO_TRACE_ID, "statemachine is resumed!");
+    }
+}
+
+void RaftReplDev::enter_emergency() {
```
yamingk, line 1779: we can rename it to something like `drain_outstanding_rreqs`, since it is some fact API homestore can provide to consumer to do stuff.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1744,4 +1760,113 @@ void RaftReplDev::report_blk_metrics_if_needed(repl_req_ptr_t rreq) {
...
+        uint64_t pending_req_num = get_pending_init_req_num();
+        if (pending_req_num) {
+            RD_LOGD(NO_TRACE_ID, "wait for {} pending create_req requests to be completed", pending_req_num);
+            std::this_thread::sleep_for(std::chrono::microseconds(1));
+        } else
+            break;
+    }
+}
+
+void RaftReplDev::leave_emergency() { m_in_emergency.store(false, std::memory_order_release); }
```
yamingk, line 1796: Then we don't need this api, and reset m_is_sm_paused to false in resume_state_machine.

yamingk:
LGTM overall, with a few comments. Please take a look. Thx

src/lib/replication/repl_dev/common.cpp
```diff
@@ -26,17 +27,34 @@ ReplServiceError repl_req_ctx::init(repl_key rkey, journal_type_t op_code, bool
...
-    // We need to allocate the block if the req has data linked, since entry doesn't exist or if it exist, two threads(data channel and raft channel) are trying to do the same
-    // thing. So take state mutex and allocate the blk
+    // We need to allocate the block if the req has data linked, since entry doesn't exist or if it exist, two
+    // threads(data channel and raft channel) are trying to do the same thing. So take state mutex and allocate the blk
     std::unique_lock< std::mutex > lg(m_state_mtx);
     if (has_linked_data() && !has_state(repl_req_state_t::BLK_ALLOCATED)) {
-        auto alloc_status = alloc_local_blks(listener, data_size);
+        ReplServiceError alloc_status;
+#ifdef _PRERELEASE
+        if (iomgr_flip::instance()->test_flip("fake_no_space_left") && !is_proposer) {
```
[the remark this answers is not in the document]
JacksonYao287: done

JacksonYao287:
@yamingk thank for the comments.

1 enter_emgerency means we waiting for all the pending initing req to be completed and refuse new req to be initialized again until we leave_emergency.

2 pasue_statemachine provide upstream a chance to pause the committing thread of raft.

so basically they are two different thing. upper layer can decide to use one or both of them, since it depends.

> When enter_emgerency is being called, it can be determined state machine is paused

this is not true.

we have two place to do emergent gc in homeobject:

1 in raft channel(the raft log handling thread),  we will call pasue_statemachine before we do the check between the last_append_lsn and last_commit_lsn. pasue_statemachine here is necessary and workable.

2 in commit channel(the raft committing thread), we do not pause_statemachine here and do the check,  since we already in the committing thread. and if we call pause_statemachine in committing thread, it will never be paused.

no sure if I fully get your point and answer you concern. if not, pls let me know

xiaoxichen:
enter_emergency is a needed feature though I am also thinking a general name like `enter_emergency` or a more clear action like `set_stop_accepting_rreq` (for example) is better ?   This is more a taste stuff we can discuss later.

The pause_statemachine is a clear action,   whether it is needed is fully based on the implementation of HO.  Although I think  there is a way to avoid using this, but again it is also a taste issue which can be discussed later.

JacksonYao287:
@yamingk @xiaoxichen 
I think we can not 100 percents determine how we want the interface looks like in the future, since it depends on the implementation details.

well , since these several function is not that complicated and easy to change, so we can postpone changing the interface here until we are very clear about this.

xiaoxichen:
agree. I already gave my LGTM.  Feel free to merge after yaming approved.

xiaoxichen (approved):

PR #685, opened 2025-04-09
add metric for blk usage diff

Besroy:


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1729,4 +1731,16 @@ bool RaftReplDev::is_resync_mode() {
     return resync_mode;
 }
 
+void RaftReplDev::report_blk_metrics_if_needed(repl_req_ptr_t rreq) {
+    auto chunk_id = rreq->local_blkid().chunk_num();
+    auto chunk = hs()->device_mgr()->get_chunk(chunk_id);
+    if (chunk->get_blk_usage() >= chunk->get_blk_usage_report_threshold()) {
+        auto local_blk_num = rreq->local_blkid().blk_num();
+        auto remote_blk_num = rreq->remote_blkid().blkid.blk_num();
+        auto blk_diff_with_remote = local_blk_num > remote_blk_num ? local_blk_num - remote_blk_num
```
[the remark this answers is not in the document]
xiaoxichen: we dont have a Histogram bucket that has negtive,  so maybe go with zero?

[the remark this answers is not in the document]
Besroy: Do you mean we should disregard the scenario where the leader has more allocated blocks than the follower, since this won't cause the follower to encounter a NO_SPACE_LEFT issue, and therefore set it to zero?

src/lib/replication/repl_dev/raft_repl_dev.h
```diff
@@ -90,6 +90,13 @@ class RaftReplDevMetrics : public sisl::MetricsGroup {
         REGISTER_HISTOGRAM(rreq_pieces_per_write, "Number of individual pieces per write",
                            HistogramBucketsType(LinearUpto64Buckets));
 
+        // In the identical layout chunk, the blk num of the follower and leader is expected to be the same.
+        // However, due to the concurrency between the data channel and the raft channel, there might be some
+        // allocation differences on the same lsn. When a leader switch occurs, these differences could become garbage.
+        // This metric can partially reflect the potential amount of garbage.
+        REGISTER_HISTOGRAM(blk_diff_with_proposer,
+                           "allocated blk num diff on the same lsn with proposer when chunk usage >= 0.9");
```
yuwmao, line 98: Considering the upper limit=chunk_size/blk_size*(1-usage_threshold). Maybe we can define a customized HistogramBuckets because many of the DefaultBuckets will be empty.
`
    X(DefaultBuckets, 10, 40, 70, 100, 160, 220, 280, 340, 400, 475, 600, 750, 900, 1100, 1400, 1700, 2000, 3500, 5000, 6500, 8000, 10000, 13000, 16000, 20000, 50000, 80000, 100000, 150000, 180000, 200000, 500000, 2000000,  3000000, 4000000)   
`

yuwmao: Sorry, I was wrong, ignore it.

yuwmao (approved):
LGTM

src/lib/replication/repl_dev/raft_repl_dev.h
xiaoxichen: I think you are right about defining bucket, it would even better if we take negative into account.

But let's start with basic to see how it works

PR #690, opened 2025-04-15
Add btree Node to the dirty buf list to keep the counter alive in the…

raakella1:
… lru cache

### DIFF
--- a/src/lib/common/homestore_config.fbs
+++ b/src/lib/common/homestore_config.fbs
@@ -130,7 +130,11 @@ table Generic {
 
     cache_max_throttle_cnt : uint32 = 4; // writeback cache max q depth
 
-    cache_min_throttle_cnt : uint32 = 4; // writeback cache min q deoth
+    cache_min_throttle_cnt : uint32 = 4; // writeback cache min q depth
+
+    cache_hashmap_nbuckets : uint32 = 1000000; // num buckets for sisl::SimpleHashmap used in wbcache
+
+    cache_evictor_npartitions: uint32 = 1000; // num partitions for lru evictor in the cache
 
     // if this value is set to 0, no sanity check will be run;
     sanity_check_level: uint32 = 1 (hotswap);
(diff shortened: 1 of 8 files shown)


src/lib/index/index_cp.cpp
```diff
@@ -63,8 +63,8 @@ void IndexCPContext::add_to_txn_journal(uint32_t index_ordinal, const IndexBuffe
     }
 }
 
-void IndexCPContext::add_to_dirty_list(const IndexBufferPtr& buf) {
-    m_dirty_buf_list.push_back(buf);
+void IndexCPContext::add_to_dirty_list(const IndexBufferPtr& buf, const BtreeNodePtr& node) {
+    m_dirty_buf_list.push_back(std::make_pair(buf, node));
```
yamingk, line 67: The node ref_cnt is automatically increased in its constructor, right?

yamingk (dismissed):

src/lib/index/index_cp.cpp
raakella1: Yes, the unit test actually fails without this.
Sure, I will add the comment.

shosseinimotlagh (approved):
how did test validate if the eviction functionality is as expected, (so only evict the clean ones)? number of eviction and/or number of objects in the cache.

Also is it needed to validate it in case of concurrency?

src/lib/common/homestore_config.fbs
```diff
@@ -130,7 +130,11 @@ table Generic {
 
     cache_max_throttle_cnt : uint32 = 4; // writeback cache max q depth
 
-    cache_min_throttle_cnt : uint32 = 4; // writeback cache min q deoth
+    cache_min_throttle_cnt : uint32 = 4; // writeback cache min q depth
+
+    cache_hashmap_nbuckets : uint32 = 1000000; // num buckets for sisl::SimpleHashmap used in wbcache
```
shosseinimotlagh, line 135: We have similar parameters in  Table Cache. Can't we use it and remove it from here?

xiaoxichen:
> how did test validate if the eviction functionality is as expected, (so only evict the clean ones)? number of eviction and/or number of objects in the cache.
> 
> Also is it needed to validate it in case of concurrency?

I asked same question in the NuObject standup,  can we tune down the cache size to force all our Index related testing(at least long running) go through the eviction path?

PR #692, opened 2025-04-18
Add additional on_commit repldev listener api's.

sanebay:
Add additional on_commit to support vector of blkids.

### DIFF
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -270,6 +270,20 @@ public:
     virtual void on_commit(int64_t lsn, sisl::blob const& header, sisl::blob const& key, MultiBlkId const& blkids,
                            cintrusive< repl_req_ctx >& ctx) = 0;
 
+    /// @brief Called when the log entry has been committed in the replica set.
+    ///
+    /// This function is called from a dedicated commit thread which is different from the original thread calling
+    /// replica_set::write(). There is only one commit thread, and lsn is guaranteed to be monotonically increasing.
+    ///
+    /// @param lsn - The log sequence number
+    /// @param header - Header originally passed with replica_set::write() api
+    /// @param key - Key originally passed with replica_set::write() api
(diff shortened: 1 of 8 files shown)


src/include/homestore/replication/repl_dev.h
```diff
@@ -270,6 +270,20 @@ class ReplDevListener {
...
+    /// This function is called from a dedicated commit thread which is different from the original thread calling
+    /// replica_set::write(). There is only one commit thread, and lsn is guaranteed to be monotonically increasing.
+    ///
+    /// @param lsn - The log sequence number
+    /// @param header - Header originally passed with replica_set::write() api
+    /// @param key - Key originally passed with replica_set::write() api
+    /// @param blkids - List of independent blkids where data is written to the storage engine.
+    /// @param ctx - Context passed as part of the replica_set::write() api
+    ///
+    virtual void on_commit(int64_t lsn, sisl::blob const& header, sisl::blob const& key,
```
xiaoxichen, line 284: unless we make it non pure virtual, merging this will still require upper layer having a change.   If upper layer change is unavoidable,  should we unified to the `vector<MultiBlkId>` approach ?

sanebay: After merging this, will update upper layers in HO and HB. Unifying is lot of changes across HO. For HB, we will have more fragmentation, so vector of blkid's is needed for large allocations to support upto 1 MB.

yamingk:
LGTM, do you want to syncup with Xiaoxi's's?

yamingk (approved):

sanebay:
> LGTM, do you want to wait with Xiaoxi's approval?
Thought about a single on_commit, but it looks like lot of changes in HO. We can discuss in HS meeting if needed. Index value is a single MultiBlkId in HO today, that also will change, so lot of changes. As you mentioned, if append blk allocator, chance of fragmentation is less compared to nublocks

PR #693, opened 2025-04-18
Add multiple blkid's support for solorepldev.

sanebay:
Solorepldev can write single multiblkid or vector of blkid's for nublocks case. Test case modified to support vector of blkids.

### DIFF
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -49,6 +49,12 @@ VENUM(journal_type_t, uint16_t,
       HS_CTRL_REPLACE = 3, // Control message to replace a member
 )
 
+// Whether block allocation type is a single multiblkid or vector of blkids.
+VENUM(blkid_alloc_type_t, uint8_t,
+      SINGLE = 0,   // Single multi blkid
+      MULTIPLE = 1, // Vector of independent blkids
+)
+
 // magic num comes from the first 8 bytes of 'echo homestore_resync_data | md5sum'
 static constexpr uint64_t HOMESTORE_RESYNC_DATA_MAGIC = 0xa65dbd27c213f327;
 static constexpr uint32_t HOMESTORE_RESYNC_DATA_PROTOCOL_VERSION_V1 = 0x01;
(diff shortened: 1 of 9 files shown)


src/lib/replication/repl_dev/solo_repl_dev.cpp
```diff
@@ -60,8 +96,15 @@ void SoloReplDev::write_journal(repl_req_ptr_t rreq) {
             auto cur_lsn = m_commit_upto.load();
             if (cur_lsn < lsn) { m_commit_upto.compare_exchange_strong(cur_lsn, lsn); }
 
-            data_service().commit_blk(rreq->local_blkid());
-            m_listener->on_commit(rreq->lsn(), rreq->header(), rreq->key(), rreq->local_blkid(), rreq);
+            if (rreq->blkid_alloc_type() == blkid_alloc_type_t::SINGLE) {
```
yamingk, line 99: nit: replace to `rreq->is_single_blkalloc_req()` so that caller even doesn't need to be aware the specific alloc type?

src/lib/replication/repl_dev/solo_repl_dev.cpp
```diff
@@ -83,16 +126,39 @@ void SoloReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
     raw_ptr += entry->key_size;
     remain_size -= entry->key_size;
 
-    sisl::blob value_blob{raw_ptr, remain_size};
-    MultiBlkId blkid;
-    if (remain_size) { blkid.deserialize(value_blob, true /* copy */); }
+    if (entry->blkid_alloc_type == blkid_alloc_type_t::SINGLE) {
```
yamingk, line 129: similar

yamingk:
Update conan version. Currently HB 2 does support minor version update only. E.g. if this PR updates minior version, it will NOT be picked by HB repo.

src/include/homestore/replication/repl_dev.h
```diff
@@ -49,6 +49,12 @@ VENUM(journal_type_t, uint16_t,
       HS_CTRL_REPLACE = 3, // Control message to replace a member
 )
 
+// Whether block allocation type is a single multiblkid or vector of blkids.
```
yamingk, line 52: Nit: we probably can move this into `repl_req_cxt` to make it only visiable to itself, not to callers.

src/include/homestore/replication/repl_dev.h
```diff
@@ -49,6 +49,12 @@ VENUM(journal_type_t, uint16_t,
...
+VENUM(blkid_alloc_type_t, uint8_t,
```
yamingk, line 53: Nit: probably to move it inside `repl_req_ctx` to make it self-contained. And caller to get the type by `repl_req_ctx::blkid_alloc_type_t::xxx`?

src/lib/replication/repl_dev/solo_repl_dev.cpp
sanebay: Done.

src/include/homestore/replication/repl_dev.h
sanebay: repl_req_ctx init is taking blkid_alloc_type as parameter. Because other enums like journal_type_t etc were outside, so kept it outside.

raakella1:
The changes looks good to me

PR #695, opened 2025-04-22
Improve raftReplDev creation

yuwmao:
Use user-defined config(e.g. priority) to create raftReplDev.

### DIFF
--- a/src/lib/replication/service/raft_repl_service.cpp
+++ b/src/lib/replication/service/raft_repl_service.cpp
@@ -59,6 +59,17 @@ ReplServiceError RaftReplService::to_repl_error(nuraft::cmd_result_code code) {
     return ret;
 }
 
+// NuRaft priority decay coefficient is set to 0.8(currently not configurable). For more details, please refer to
+// https://github.com/eBay/NuRaft/blob/master/docs/leader_election_priority.md
+int32_t RaftReplService::compute_raft_follower_priority() {
+    auto max_wait_round = std::min(raft_priority_election_round_upper_limit,
+                                   HS_DYNAMIC_CONFIG(consensus.max_wait_rounds_of_priority_election));
+    if (max_wait_round == 0) { return raft_leader_priority; }
+    auto priority = 1 + static_cast< int32_t >(
+                        std::ceil(raft_leader_priority * std::pow(raft_priority_decay_coefficient, max_wait_round)));
+    return priority;
+}
+
 RaftReplService::RaftReplService(cshared< ReplApplication >& repl_app) : GenericReplService{repl_app} {
     m_config_sb_bufs.reserve(100);
     meta_service().register_handler(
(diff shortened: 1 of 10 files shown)


xiaoxichen (commented):
Reading this doc https://github.com/eBay/NuRaft/blob/1c2587f159775a4924573db64ffac5c973230761/docs/leader_election_priority.md#L16

I am concerning of open the priority of int32 to CM,  as without understanding the implementation detail it probabaly set  < int32_max,  1, 0>  which will be a nightmare for  leader election.

Probabaly HO can still provide a list of `replica_id_t` and HS decide the priority based on the internal knowledge,  we can make the K  configurable where K = round of election timeout needed. 

Decay = (0.8^k) , for example K =2 make it <10000, 6400, 4096 >. (each with 0.64 decay meaning 2 election timeout).

yuwmao:
> Reading this doc https://github.com/eBay/NuRaft/blob/1c2587f159775a4924573db64ffac5c973230761/docs/leader_election_priority.md#L16
> 
> I am concerning of open the priority of int32 to CM, as without understanding the implementation detail it probabaly set < int32_max, 1, 0> which will be a nightmare for leader election.
> 
> Probabaly HO can still provide a list of `replica_id_t` and HS decide the priority based on the internal knowledge, we can make the K configurable where K = round of election timeout needed.
> 
> Decay = (0.8^k) , for example K =2 make it <10000, 6400, 4096 >. (each with 0.64 decay meaning 2 election timeout).

Good point, and the gap should be bigger than 10. Regarding to `decide the priority based on the internal knowledge`, I think we can only infer who is the leader although it's enough in our NuObject scenario.
There are two thoughts:
1. Make strict control of the priority setting in HS, make sure they are valid and we can define K as max round the election timeout, leaving some flexibility for CM.
2. Simplify the contract with CM, change priority to a bool which means who is assigned to the leader(maybe not necessary in proto, but at least CM should keep the intent in its DB). And in HS's replace_member, the in_member inherit out_member's priority directly based on srv_config.

I think option2 is simpler and more controllable.

xiaoxichen:
Yes we discussed this a bit in CM meeting,  currently there is no use-case we want to control the secondary-leader. 

- CM will nominate a leader, and CM will persist the leader in its DB as well as internal PG structure. So that CM is able to tell for any given PG if the leader has drifted. 

- When CM send create pg request,  it will carry the leader intent, either through priority or we change the grpc protocol. But at least for now it is clear the one who receive the PG_CREATE is the leader.
- HO/HS should decide the priority setting of NuRaft, based on the K configuration or other internal knowledge.
- When replace member,  the in_member inherit the priority ---though for the sake of completeness I believe the in_member will have a flag `should_be_leader` indicating if it supposed to be leader.

So basically it is your approach 2

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -59,6 +59,17 @@ ReplServiceError RaftReplService::to_repl_error(nuraft::cmd_result_code code) {
     return ret;
 }
 
+// NuRaft priority decay coefficient is set to 0.8(currently not configurable). For more details, please refer to
+// https://github.com/eBay/NuRaft/blob/master/docs/leader_election_priority.md
+int32_t RaftReplService::compute_raft_follower_priority() {
+    auto max_wait_round = std::min(raft_priority_election_round_upper_limit,
+                                   HS_DYNAMIC_CONFIG(consensus.max_wait_rounds_of_priority_election));
+    if (max_wait_round == 0) { return raft_leader_priority; }
+    auto priority = 1 + static_cast< int32_t >(
```
xiaoxichen, line 68: this is a nice formula

xiaoxichen (dismissed):

xiaoxichen (approved):

xiaoxichen (approved):

PR #703, opened 2025-04-28
support handling config rollback and add periodical notfity the lastest committed lsn to upper layer

JacksonYao287:


src/include/homestore/replication/repl_dev.h
```diff
@@ -270,6 +270,20 @@ class ReplDevListener {
...
+    ///
+    /// This function is called from a dedicated commit thread which is defferent from the
+    /// replace_member/add_member/remove_member thread. There is only one commit thread, and lsn is guaranteed to be
+    /// monotonically increasing.
+    ///
+    /// @param lsn - The log sequence number
+    /// @param new_config - the new config that is being committed.
+    ///
+    /// config change will not involve any data write, so the blkids are not passed.
+    virtual void on_commit_config(int64_t lsn, raft_cluster_config_ptr_t& new_config){
```
xiaoxichen, line 283: on_config_commit

JacksonYao287: ack

Besroy (dismissed):
LGTM

src/include/homestore/replication/repl_dev.h
```diff
@@ -270,6 +270,19 @@ class ReplDevListener {
...
+    /// @brief Called when the config log entry has been committed in the replica set.
+    ///
+    /// This function is called from a dedicated commit thread which is defferent from the
+    /// replace_member/add_member/remove_member thread. There is only one commit thread, and lsn is guaranteed to be
+    /// monotonically increasing.
+    ///
+    /// @param lsn - The log sequence number
+    ///
+    /// config change will not involve any data write, so the blkids are not passed.
+    virtual void on_config_commit(int64_t lsn){
```
Besroy, line 282: Just curious why not set '=0' to ensure that HO implements those functions.

JacksonYao287: 1 if we set it to 0, HO will have to implement this. this means before HO implements this, all the PR in homeobject will hit building failure. 

2 some other upper layers(for example ,nublox) probably not care about this signal and this callback. if we make this pure virtual, all the upper layers will have to implement this callback, even if they don`t care about this.

but, yes, we can make it pure virtual, I can change it after homeobject side change is completed so that it will not block any other pr for homeobject

xiaoxichen:
lets wait @yamingk @sanebay to review before merging this.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1035,14 +1035,20 @@ void RaftReplDev::handle_config_commit(const repl_lsn_t lsn, raft_cluster_config
     // when reaching here, the new config has already been applied to the cluster.
     // since we didn't create repl req for config change, we just need to update m_commit_upto_lsn here.
     RD_LOGD(NO_TRACE_ID, "config commit on lsn {}", lsn);
-    // keep this variable in case it is needed later
-    (void)new_conf;
+    (void)new_conf; // use this if we need it in the future
+    m_listener->on_config_commit(lsn);
```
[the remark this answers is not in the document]
JacksonYao287: if we pass this as void*, when we want to use it, we have to convert it to a nuraft type, which will also break the boundary. well, IMO, we can pass it to upstream until we know we definitely need it and how to use it , changes about this will not be too expensive.

yuwmao:
Just a reminder, `HS_CTRL_REPLACE` won't call on_commit as well. But it always happens with config change, so seems it's ok in GC scenario, right?

JacksonYao287:
> Just a reminder, `HS_CTRL_REPLACE` won't call on_commit as well. But it always happens with config change, 

wow, this is a very nice catch. thanks for pointing out this @yuwmao.

basically, from the perspective of nuraft, HS_CTRL_REPLACE/HS_CTRL_DESTROY are both normal(not config) log, so they should be taken into account for this case.  

>it's ok in GC scenario, right?
do we mean chunk gc or pg gc?  I think it has nothing to do with chunk gc.

JacksonYao287:
handle op code in https://github.com/eBay/HomeObject/pull/293

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1012,15 +1012,18 @@ void RaftReplDev::handle_commit(repl_req_ptr_t rreq, bool recovery) {
...
         leave();
     } else if (rreq->op_code() == journal_type_t::HS_CTRL_REPLACE) {
         replace_member(rreq);
-    } else {
-        m_listener->on_commit(rreq->lsn(), rreq->header(), rreq->key(), {rreq->local_blkid()}, rreq);
     }
 
+    // for all the logs(excluding config), we will call on_commit callback to the listener.
+    // upstream should filter out the log type that they are not interested in.
+    m_listener->on_commit(rreq->lsn(), rreq->header(), rreq->key(), {rreq->local_blkid()}, rreq);
```
xiaoxichen, line 1025: I think this implementation is strange and lead to bugs in future.
the commit of replace_member and the callback up to consumer is in ` m_listener->on_replace_member`.
Now it need to go through `m_listener->on_commit` again which is wired.

if we want similar guarantee like NuRaft, for every LSN either on_commit or on_rollback will be called.  then literary we need to merge on_replace_member/on_destroy to on_commit....

xiaoxichen (commented):
Overall, I question whether the effort to ensure every commit calls HO for handling GC in the context of on_commit is more beneficial than simply monitoring `get_last_commit_lsn` in another thread..

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: 1 I have tried to find if there is any event (like raft_repl_dev#raft_event)after [commit/commit_config ](https://github.com/eBay/NuRaft/blob/68e36210e0dc224852890c90517fbb33c9365ac5/src/handle_commit.cxx#L251-L256), so that we can add the logic there. **if we can add an event here, i think this will be the most graceful way**, since what we want is only be notified a lsn is committed, and not care about the logic of  commit.
we can discuss this with @greensky00 to see if we can add a specific cb_func type
I think we probably can implement this [event callback ](https://github.com/eBay/NuRaft/blob/68e36210e0dc224852890c90517fbb33c9365ac5/src/handle_commit.cxx#L266-L267), then it will be all good

2  we have done in this way in on_pre_commit of HO, [which filter out the uninterested op code](https://github.com/eBay/HomeObject/blob/be7114069beb8694bc658f94e40c1e46fe61849f/src/lib/homestore_backend/replication_state_machine.cpp#L79-L87),

>literary we need to merge on_replace_member/on_destroy to on_commit

3 I get your point, and honestly I wanted to do this when writing code. but actually. well, HS_CTRL_DESTROY and HS_CTRL_REPLACE are internal control log, which should not been seen by upper layer. if we have another way to implement `destroy` and `replace` without sending an internal log, then this will not be a problem. 


as a summary, what we want here is only be notified when a lsn has been committed but not care about the commit logic, but in the current implementation of raft_repl_dev,  we can not achieve this without involving on_commit.

JacksonYao287: monitoring in a separate will be more complicated , I prefer not to do this

JacksonYao287:
@xiaoxichen  thanks for the great idea! I have made this change, pls take a look.

src/include/homestore/replication/repl_dev.h
```diff
@@ -270,6 +270,12 @@ class ReplDevListener {
     virtual void on_commit(int64_t lsn, sisl::blob const& header, sisl::blob const& key,
                            std::vector< MultiBlkId > const& blkids, cintrusive< repl_req_ctx >& ctx) = 0;
 
+    /// @brief periodically called to notify the lastest committed lsn to the listener.
+    ///
+    /// @param lsn - The log sequence number
```
xiaoxichen, line 275: Concerning any heavy/blocking operations in the CB as it will block the DC_LSN flushing.

Lets start with putting clear comment here

xiaoxichen (approved):

src/include/homestore/replication/repl_dev.h
JacksonYao287: comments updated, ptal

xiaoxichen (approved):
lgtm.

Though I wish in GC thread model if this thread can be taken care by a global thread (across all ReplDev) in HO.

xiaoxichen:
please write a decent commit message when squashing , either now or when merging.

PR #706, opened 2025-04-28
Add async_write, alloc blks for solo repl dev.

sanebay:
Add support for async write data, journal, alloc blks for solo repl dev. Raft repl dev doesnt support these operations. This is needed for nublocks where it need to write free blkids also to the journal. Free blocks are obtained after writing the new blkids to index. Add apis for allocation and write for vector of blkids . Raft repldev currently uses only a single blkid. Test solo repl dev changes to support vector of blkids.


conanfile.py
```diff
@@ -54,7 +54,7 @@ def build_requirements(self):
     def requirements(self):
         self.requires("iomgr/[^11.3]@oss/master", transitive_headers=True)
         self.requires("sisl/[^12.2]@oss/master", transitive_headers=True)
-        self.requires("nuraft_mesg/[>=3.7.5]@oss/main", transitive_headers=True)
+        self.requires("nuraft_mesg/3.8.0@oss/main", transitive_headers=True)
```
yamingk, line 57: should we want change to [~3.8.0]?

conanfile.py
```diff
@@ -9,7 +9,7 @@
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.12.3"
+    version = "6.12.4"
```
yamingk, line 12: I believe version is already 6.12.4, need to bump one more

src/include/homestore/replication/repl_dev.h
```diff
@@ -195,7 +206,8 @@ struct repl_req_ctx : public boost::intrusive_ref_counter< repl_req_ctx, boost::
     bool save_fetched_data(sisl::GenericClientResponse const& fetched_data, uint8_t const* data, uint32_t data_size);
 
     void set_remote_blkid(RemoteBlkId const& rbid) { m_remote_blkid = rbid; }
-    void set_local_blkid(MultiBlkId const& lbid) { m_local_blkid = lbid; } // Only used during recovery
+    void set_local_blkid(MultiBlkId const& lbid) { m_local_blkids.push_back(lbid); } // Only used during recovery
```
yamingk, line 209: Looks like there is one place calling for `set_local_blkid`, can we merge into your new API: `set_local_blkids(std::vector< MultiBlkId > const& lbids)` 

by ```rreq->set_local_blkid({entry_blkid});``` ?

src/include/homestore/replication/repl_dev.h
```diff
@@ -226,9 +238,10 @@ struct repl_req_ctx : public boost::intrusive_ref_counter< repl_req_ctx, boost::
     std::atomic< bool > m_is_volatile{true};                   // Is the log still in memory and not flushed to disk yet
 
     /////////////// Data related section /////////////////
-    MultiBlkId m_local_blkid;   // Local BlkId for the data
-    RemoteBlkId m_remote_blkid; // Corresponding remote blkid for the data
-    uint8_t const* m_data;      // Raw data pointer containing the actual data
+    MultiBlkId dummy_blkid;
```
yamingk, line 241: Do you think it worth a member inside reqpl_req_ctx, given there might be a lot of outstanding rreqs which might add to memory budget? Can we `return {}` when we want a dummy_blkid?

src/include/homestore/replication/repl_dev.h
```diff
@@ -130,7 +130,16 @@ struct repl_req_ctx : public boost::intrusive_ref_counter< repl_req_ctx, boost::
...
     sisl::blob const& header() const { return m_header; }
     sisl::blob const& key() const { return m_key; }
-    MultiBlkId const& local_blkid() const { return m_local_blkid; }
+    MultiBlkId const& local_blkid() const {
+        // Currently used by raft repl dev only where a single blob is expected.
+        // Code checks if its a valid blkid so return a dummy blkid.
+        if (!m_local_blkids.empty())
+            return m_local_blkids[0];
+        else
+            return dummy_blkid;
```
yamingk, line 139: How about `return {}` ?

src/include/homestore/blkdata_service.hpp
```diff
@@ -156,6 +168,16 @@ class BlkDataService {
...
 
+    /**
+     * @brief Allocates blocks of disk space of the given size.
+     *
+     * @param size The size of the block to allocate, in bytes.
+     * @param hints Hints for how to allocate the block.
+     * @param out_blkids Output parameter that will be filled with the IDs of the allocated blocks.
+     * @return The status of the block allocation attempt.
+     */
+    BlkAllocStatus alloc_blks(uint32_t size, blk_alloc_hints const& hints, std::vector< BlkId >& out_blkids);
```
yamingk, line 179: I am not sure it is possible or making things messy, does it make sense to take `std::vector<MultiBlkID>& out_blkids` here as well? The reason is if we are able to do this, we can even merge the two alloc_blks APIs, right?

src/include/homestore/replication/repl_dev.h
```diff
@@ -399,6 +412,39 @@ class ReplDev {
...
 
+    /// @brief Allocates blkids from the storage engine to write the value into. Storage
+    /// engine returns a blkid_list in cases where single contiguous blocks are not
+    /// available.
+    ///
+    /// @param data_size - Size of the data.
+    /// @param hints - Specify block allocation hints.
+    /// @param out_blkids - List of bilkid's which may not be contiguous.
+    virtual std::error_code alloc_blks(uint32_t data_size, const blk_alloc_hints& hints,
+                                       std::vector< MultiBlkId >& out_blkids) = 0;
```
yamingk, line 423: looks like we are already exposing `std::vector< MultiBlkId >& out_blkids`, not `std::vector< BlkId >& out_blkids`, right?

src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -219,6 +232,18 @@ BlkAllocStatus BlkDataService::alloc_blks(uint32_t size, const blk_alloc_hints&
     return ret;
 }
 
+BlkAllocStatus BlkDataService::alloc_blks(uint32_t size, const blk_alloc_hints& hints,
+                                          std::vector< BlkId >& out_blkids) {
+    if (is_stopping()) return BlkAllocStatus::FAILED;
+    incr_pending_request_num();
+    HS_DBG_ASSERT_EQ(size % m_blk_size, 0, "Non aligned size requested");
```
yamingk, line 239: can we also print the size and m_blk_size in the print message for ease of debug?

src/lib/replication/repl_dev/solo_repl_dev.cpp
```diff
@@ -60,12 +60,90 @@ void SoloReplDev::write_journal(repl_req_ptr_t rreq) {
...
+    sisl::sg_iterator sg_it{value.iovs};
+
+    for (const auto& blkid : blkids) {
+        auto sgs_size = blkid.blk_count() * data_service().get_blk_size();
+        const auto iovs = sg_it.next_iovs(sgs_size);
+        uint32_t total_size = 0;
+        for (auto& iov : iovs) {
+            total_size += iov.iov_len;
+        }
+        HS_REL_ASSERT(total_size == sgs_size, "Block size mismatch");
```
yamingk, line 111: Can we just fail the IO, but not crash?

conanfile.py
sanebay: Ack

conanfile.py
sanebay: Ack

src/include/homestore/replication/repl_dev.h
sanebay: Ack

src/include/homestore/replication/repl_dev.h
sanebay: Couldnt do it because it gives error returning reference to temporary

src/include/homestore/blkdata_service.hpp
```diff
@@ -156,6 +168,16 @@ class BlkDataService {
      */
     BlkAllocStatus alloc_blks(uint32_t size, blk_alloc_hints const& hints, MultiBlkId& out_blkids);
 
+    /**
+     * @brief Allocates blocks of disk space of the given size.
```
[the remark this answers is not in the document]
sanebay: Yeah thats a good suggestion. Ack

src/include/homestore/blkdata_service.hpp
sanebay: alloc_blks internally call respective virtual_vdev->alloc_blks which behave differently. For vector , it sets    h.partial_alloc_ok = true and h.is_contiguous = true and and allocation could span across multiple chunk (https://github.com/eBay/HomeStore/blob/master/src/lib/device/virtual_dev.cpp#L246), whereas in single multiblkid case it expects to be contingious in a single chunk.

src/include/homestore/replication/repl_dev.h
```diff
@@ -130,7 +130,16 @@ struct repl_req_ctx : public boost::intrusive_ref_counter< repl_req_ctx, boost::
...
+    MultiBlkId const& local_blkid() const {
+        // Currently used by raft repl dev only where a single blob is expected.
+        // Code checks if its a valid blkid so return a dummy blkid.
+        if (!m_local_blkids.empty())
+            return m_local_blkids[0];
+        else
+            return dummy_blkid;
+    }
+
+    std::vector< MultiBlkId >& local_blkids() { return m_local_blkids; }
```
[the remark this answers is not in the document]
sanebay: Yeah its a minor change but in lot of places in raft repl dev.

src/lib/replication/repl_dev/solo_repl_dev.cpp
```diff
@@ -60,12 +60,90 @@ void SoloReplDev::write_journal(repl_req_ptr_t rreq) {
...
+        sisl::sg_list sgs{sgs_size, iovs};
+        futs.emplace_back(data_service().async_write(sgs, blkid, part_of_batch));
+    }
+
+    return folly::collectAllUnsafe(futs).thenTry([this](auto&& v_res) {
```
yamingk, line 116: is there a reason to use thenTry not thenValue? Or do we expect caller to do thenError in this case?

src/lib/blkdata_svc/blkdata_service.cpp
sanebay: Ack

src/lib/replication/repl_dev/solo_repl_dev.cpp
```diff
@@ -60,12 +60,90 @@ void SoloReplDev::write_journal(repl_req_ptr_t rreq) {
...
+    std::vector< BlkId > blkids;
+    auto status =
+        data_service().alloc_blks(sisl::round_up(uint32_cast(data_size), data_service().get_blk_size()), hints, blkids);
+    if (status != BlkAllocStatus::SUCCESS) {
+        DEBUG_ASSERT_EQ(status, BlkAllocStatus::SUCCESS, "Unable to allocate blks");
+        decr_pending_request_num();
+        return std::make_error_code(std::errc::no_space_on_device);
+    }
+    for (auto& blkid : blkids) {
+        out_blkids.emplace_back(blkid);
```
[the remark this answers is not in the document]
sanebay: Yes

src/lib/replication/repl_dev/solo_repl_dev.cpp
sanebay: This shouldnt happen due to correctness. But we should return error instead of assert when write fails.

src/include/homestore/replication/repl_dev.h
raakella1: can this be a static instead?

src/lib/replication/repl_dev/solo_repl_dev.cpp
```diff
@@ -83,22 +161,31 @@ void SoloReplDev::on_log_found(logstore_seq_num_t lsn, log_buffer buf, void* ctx
     raw_ptr += entry->key_size;
     remain_size -= entry->key_size;
 
-    sisl::blob value_blob{raw_ptr, remain_size};
-    MultiBlkId blkid;
-    if (remain_size) { blkid.deserialize(value_blob, true /* copy */); }
+    std::vector< MultiBlkId > blkids;
+    if (remain_size == 0) {
+        blkids.push_back(MultiBlkId{});
```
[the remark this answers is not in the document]
sanebay: Ack. We dont need it anymore as we have single on_commit.

src/lib/replication/repl_dev/solo_repl_dev.cpp
sanebay: Both should be fine. Changed to thenValue.

conanfile.py
yamingk: that build error is already fixed with my yesterday's PR

src/include/homestore/replication/repl_dev.h
yamingk: Then can we move this to the function itself, have it be a local var.

src/lib/replication/repl_dev/solo_repl_dev.cpp
yamingk: as we discussed, let's return io error since this is input parameter error

sanebay: Ack

src/include/homestore/replication/repl_dev.h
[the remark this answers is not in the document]
sanebay: Created a separate issue to track the local_blkid change, as there are lot of places it expect blkids check and its not valid. UT fails too.

src/include/homestore/replication/repl_dev.h
sanebay: Its returning reference, so cant return the ref. Made it static

src/include/homestore/blkdata_service.hpp
sanebay: As discussed, added comment in data service header about the difference in the api's.

xiaoxichen (commented):
Do we get a feeling that solo dev is no longer "like" the repl  dev?  

Previously the solo_repl_dev was written as a phototype for integration IIUC.  Now it becomes first class citizen, then the question comes to as it diverse from the entry (async_write vs async_alloc_write,  write_journal etc),  implementation,  logic,   also the consumer (user of repl dev).   


What is the benefit of putting them under same base class?  @yamingk

sanebay:
> Previously the solo_repl_dev was written as a phototype for integration IIUC. Now it becomes first class citizen, then the question comes to as it diverse from the entry (async_write vs async_alloc_write, write_journal etc), implementation, logic, also the consumer (user of repl dev).
> 
solo repl dev was not prototype. Hari planned to use it for nublocks. We didnt go to details at that time and assumed async_alloc_write would be enough. But with detail design, we found that we will need this additional api's

xiaoxichen:
> > Previously the solo_repl_dev was written as a phototype for integration IIUC. Now it becomes first class citizen, then the question comes to as it diverse from the entry (async_write vs async_alloc_write, write_journal etc), implementation, logic, also the consumer (user of repl dev).
> 
> solo repl dev was not prototype. Hari planned to use it for nublocks. We didnt go to details at that time and assumed async_alloc_write would be enough. But with detail design, we found that we will need this additional api's

OK,  as far as you are still seeing its beneficial to stay under the repl_dev umbrella instead of feeling burden to refactor/keep consistency with raft_repl_dev,  I am good with either.

yamingk:
Can we document the one todo here in your PR description that we plan to even simplify the interface further by merging local_blkid and local_blkIds into one, have have the [issue](https://github.com/eBay/HomeStore/issues/707) number as well. 

Also can we also list the discussion we had as well for the possibility to even mrege the two alloc_blks in repl_dev.h, e.g. 
1. alloc_blks with MultiBlkId& out_bid
2. alloc_blks with std::vector<MultiBlkId>& out_bids

We didn't do that in this PR as we are not sure wether NuObject use case would prefer to use its own interface or have a combined one for better readability (e.g. why append blk allocator would need a vector of bids), so we didn't make this merge at the moment. cc @xiaoxichen

conanfile.py
sanebay: Yeah it works 7.x

sanebay:
> Can we document the one todo here in PR description that we plan to even simplify the interface further by merging local_blkid and local_blkIds into one, have the [issue](https://github.com/eBay/HomeStore/issues/707) number as well.
> 
Agree to the comment above. I have added a todo in commit description about the issue created. We should simplify in later PR if needed.

yamingk:
LGTM, after the build failure is fixed in CI

yamingk (approved):

PR #708, opened 2025-04-30
Add async_read api which takes a vector of MultiBlkIds

raakella1:
Add the API 
```
virtual folly::Future< std::error_code > async_read(std::vector<MultiBlkId> const& blkids, sisl::sg_list& sgs, uint32_t size,
                                                        bool part_of_batch = false, trace_id_t tid = 0) = 0;                              
```
This is to remove the complexity of collecting futures during the read of a list of BlkIds

### DIFF
--- a/src/include/homestore/blkdata_service.hpp
+++ b/src/include/homestore/blkdata_service.hpp
@@ -140,6 +140,18 @@ public:
                                                 bool part_of_batch = false);
 
     /**
+     * @brief Asynchronously reads data from the specified block ID.
+     *
+     * @param bids List of block IDs to read from.
+     * @param sgs The scatter-gather list to store the read data.
+     * @param part_of_batch Whether this read is part of a batch.
+     *
+     * @return A `folly::Future` that will contain the error code of the read operation.
+     */
+    folly::Future< std::error_code > async_read(std::vector<MultiBlkId> const& bids, sisl::sg_list& sgs, uint32_t size,
+                                                bool part_of_batch = false);
+
+    /**
      * @brief Commits the block with the given MultiBlkId.
      *
      * @param bid The MultiBlkId of the block to commit.
--- a/src/lib/blkdata_svc/blkdata_service.cpp
+++ b/src/lib/blkdata_svc/blkdata_service.cpp
@@ -110,6 +110,11 @@ folly::Future< std::error_code > BlkDataService::async_read(MultiBlkId const& bl
 
 folly::Future< std::error_code > BlkDataService::async_read(MultiBlkId const& blkid, sisl::sg_list& sgs, uint32_t size,
                                                             bool part_of_batch) {
+    return async_read(std::vector<MultiBlkId>{blkid}, sgs, size, part_of_batch);
+}
+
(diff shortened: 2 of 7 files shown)


src/lib/blkdata_svc/blkdata_service.cpp
```diff
@@ -110,6 +110,11 @@ folly::Future< std::error_code > BlkDataService::async_read(MultiBlkId const& bl
 
 folly::Future< std::error_code > BlkDataService::async_read(MultiBlkId const& blkid, sisl::sg_list& sgs, uint32_t size,
                                                             bool part_of_batch) {
+    return async_read(std::vector<MultiBlkId>{blkid}, sgs, size, part_of_batch);
```
szmyd, line 113: This will add latency to the operation purely to reduce the redundancy in the code no? Can we do decompose these two into three functions even if there is some repetition so we don't pay a cost?

src/include/homestore/blkdata_service.hpp
```diff
@@ -139,6 +139,18 @@ class BlkDataService {
...
+     * @brief Asynchronously reads data from the specified block ID.
+     *
+     * @param bids List of block IDs to read from.
+     * @param sgs The scatter-gather list to store the read data.
+     * @param part_of_batch Whether this read is part of a batch.
+     *
+     * @return A `folly::Future` that will contain the error code of the read operation.
+     */
+    folly::Future< std::error_code > async_read(std::vector<MultiBlkId> const& bids, sisl::sg_list& sgs, uint32_t size,
+                                                bool part_of_batch = false);
```
szmyd, line 152: I'm a little confused what this will be used for for a couple reasons:

1. There is only a single size; all all reads then assumed to be for the same length of data?
2. There is only a single Future provided, so all submitted READs have to pass/fail; this won't be useful to the target (SCSTConnector) since all READs are independent and get their own status (can succeed/fail independently). Is this purely for internal use?

src/lib/blkdata_svc/blkdata_service.cpp
raakella1: Yes, it is to purely reduce duplicate code. I can separate the code to avoid the latency

src/include/homestore/blkdata_service.hpp
raakella1: 1. That is a good point. The size should not be a single variable
2. This API is for HomeBlocks internal use and scst connector will not use it. To be more specific, Each read() call that scst connector makes to Homeblocks will have lba and nlbas specified which is a contiguous lba range but the pbas are not necessarily contiguous. So, we need a vector of block ids to be read from solo repl dev

src/include/homestore/blkdata_service.hpp
```diff
@@ -139,6 +139,18 @@ class BlkDataService {
     folly::Future< std::error_code > async_read(MultiBlkId const& bid, sisl::sg_list& sgs, uint32_t size,
                                                 bool part_of_batch = false);
 
+    /**
+     * @brief Asynchronously reads data from the specified block ID.
```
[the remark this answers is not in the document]
raakella1: ack

raakella1:
Moving this change to HomeBlocks

PR #712, opened 2025-05-02
Use application context in req struct for multi put and multi remove.

sanebay:
Shift is passed user context which gives users additinal control on what to serialize.


shosseinimotlagh (approved):
LG!

PR #720, opened 2025-05-13
add long running test with put and remove

raakella1:


src/tests/test_index_crash_recovery.cpp
```diff
@@ -494,8 +520,214 @@ struct IndexCrashTest : public test_common::HSTestHelper, BtreeTestHelper< TestT
...
+                    }
+                    file.close();
+                }
+            }
+            
+            LOGDEBUG("Lets before crash print operations\n{}", SequenceGenerator::printOperations(operations));
+            
+            for (auto [k, op] : operations) {
+                if (op == OperationType::Remove) {
+                    if(this->tree_key_count() < 1) {
```
shosseinimotlagh, line 673: this is very heavy operation. I suggest book keeping the number entities in a variable

shosseinimotlagh (approved):
LG

PR #723, opened 2025-05-16
[Solo repl dev] Fix log dev flush timer cancel race 

yamingk:
Only impact Solo Repl Dev which enables timer based flush thread:
* Fixed a race (happens ocasionally) that cancel_timer can finish later than the main fiber which will cause asan issue.

Testing:
======
Run 100 times for the test case:
```
for i in {1..100};  do ./Debug/src/tests/test_meta_blk --gtest_filter=VMetaBlkMgrTest.write_to_full_test --use_file=true ; done
```

Cross tested wit HomeBlocks as well to run 100 times.

### DIFF
--- a/src/lib/logstore/log_dev.cpp
+++ b/src/lib/logstore/log_dev.cpp
@@ -149,7 +149,10 @@ void LogDev::stop() {
     // after we call stop, we need to do any pending device truncations
     truncate();
     m_id_logstore_map.clear();
-    if (allow_timer_flush()) stop_timer();
+    if (allow_timer_flush()) {
+        auto f = stop_timer();
+        std::move(f).get();
+    }
 }
 
 void LogDev::destroy() {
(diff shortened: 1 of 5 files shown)


src/lib/logstore/log_dev.cpp
```diff
@@ -167,13 +171,19 @@ void LogDev::start_timer() {
         });
 }
 
-void LogDev::stop_timer() {
-    if (m_flush_timer_hdl != iomgr::null_timer_handle) {
-        iomanager.run_on_forget(logstore_service().flush_thread(), [this]() {
+folly::Future< int > LogDev::stop_timer() {
```
sanebay, line 174: nit: You can use folly::Unit too.

sanebay (dismissed):

sanebay (approved):

shosseinimotlagh (approved):

PR #724, opened 2025-05-16
Fix prefix - reload CompactBitSet

shosseinimotlagh:
Reload CompactBitSet after updating node phys buffer.

When a multiple happens and previous cp X is in flushing state and a node buffer is getting freed, the node buffer is copied for the working cp X+1. This leads to updating the physical node buffer. If there is any class object in a node type, the object still has the previous address which can result in either corruption of flushing buffer or head-use-after-free right after cp X flushing is complete and freeing the dirty index buffer.

### DIFF
--- a/src/tests/test_mem_btree.cpp
+++ b/src/tests/test_mem_btree.cpp
@@ -50,8 +50,8 @@ SISL_OPTION_GROUP(
      ::cxxopts::value< std::vector< std::string > >(), "operations [...]"),
     (preload_size, "", "preload_size", "number of entries to preload tree with",
      ::cxxopts::value< uint32_t >()->default_value("1000"), "number"),
-    (max_keys_in_node, "", "max_keys_in_node", "max_keys_in_node",
-     ::cxxopts::value< uint32_t >()->default_value("0"), ""),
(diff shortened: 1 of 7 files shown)


xiaoxichen:
great work! if you can add more details in the commit message that will be helpful for others.

src/tests/test_mem_btree.cpp
```diff
@@ -317,9 +315,7 @@ struct BtreeConcurrentTest : public BtreeTestHelper< TestType >, public ::testin
 #endif
         this->m_cfg.m_max_merge_level = SISL_OPTIONS["max_merge_level"].as< uint8_t >();
         this->m_cfg.m_merge_turned_on = !SISL_OPTIONS["disable_merge"].as< bool >();
-        if constexpr (std::is_same_v<TestType, PrefixIntervalBtreeTest>) {
-            this->m_cfg.m_merge_turned_on = false;
-        }
+        if constexpr (std::is_same_v< TestType, PrefixIntervalBtreeTest >) { this->m_cfg.m_merge_turned_on = false; }
```
sanebay, line 318: Why are disabling merge ?

sanebay (approved):

PR #728, opened 2025-05-28
Add truncation and recovery changes for solo repl dev.

sanebay:
Add commit blk for recovery and  truncation for solo repl dev.


yamingk:
Can we do some brief documentation either in the document of the IO flow/trunction section about the policy or some wording here in the PR description, it would be easy to reviewers & others to catch up. Thx

PR #737, opened 2025-06-04
UTs for simulating tombstone and GC

shosseinimotlagh:
1. UTs for simulating tombstone and GC
2. UPDATE: LEAVE IT FOR FUTURE DECISION:  **Contract change**: when index service returns cp_mistmatch, it is callers' responsibility to retry.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.20.13"
+    version = "6.20.14"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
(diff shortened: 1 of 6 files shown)


JacksonYao287 (commented):
for the implementation, LGTM, thanks for this patching.
>Contract change: when index service returns cp_mistmatch, it is callers' responsibility to retry.

I concern more about this part. this will make all the upper layer caller to change their behavior. we`d better to make the whole team (maybe in the homestore meeting) to be aware of this contract change, and see if anyone has any concerns?

shosseinimotlagh:
> we`d better to make the whole team (maybe in the homestore meeting) to be aware of this contract change, and see if anyone has any concerns?
Agreed! I am creating a doc/git issue and we can discuss in the next homestore meeting

JacksonYao287 (approved):
LGTM ， @shosseinimotlagh thanks for this UT。
@raakella1  pls take a look

PR #739, opened 2025-06-06
fix append allocator free

JacksonYao287:
we need to check if the blk to be freed is an alloced blk. if yes , free it. if no, do not increase the counter.
in homeobject gc, some times a blk free will hit the chunk which is already reset and reused since we always async_free

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.16.1"
+    version = "6.16.2"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/lib/blkalloc/append_blk_allocator.cpp
+++ b/src/lib/blkalloc/append_blk_allocator.cpp
@@ -138,8 +138,16 @@ void AppendBlkAllocator::cp_flush(CP* cp) {
 
 // free operation books keeping "total freeable" space
 void AppendBlkAllocator::free(const BlkId& bid) {
-    m_freeable_nblks.fetch_add(bid.blk_count());
-    m_is_dirty.store(true);
+    if (is_blk_alloced(bid)) {
+        m_freeable_nblks.fetch_add(bid.blk_count());
+        m_is_dirty.store(true);
+        return;
+    }
+    // we have no lock in append_blk_allocator, so there is a corner case that
+    //  is_blk_alloced(bid) is true, so we lost the update of m_freeable_nblks.
+    //  should we involve a lock here?
+    LOGWARN("Trying to free an unallocated block: {} , m_last_append_offset is {}", bid,
+            m_last_append_offset.load(std::memory_order_relaxed));
 }
 
 bool AppendBlkAllocator::is_blk_alloced(const BlkId& in_bid, bool) const {


xiaoxichen (changes requested):
is_blk_alloced is not reliable for append_allocator
```
is_blk_alloced(blk_1)
free(blk_1)
assert(is_blk_allocated(blk_1)
```
can pass.

It might fix your very specific problem that you want to gate the "free" after resetting the allocator. 

I am neutral to the "fix" as it do improve specific case although in general it is bugly. If we want to take this path, write clear comments here with reasons.

JacksonYao287:
my point is we should never free a blk that never allocated ( blk num exceed the last_append_offset).  this will gate a lot of unexpected free to a chunk, this will be helpful for gc to some extent. yes, this can be considered as an improvement.

since the defrag_number of the append_allocator is no reliable, and homeobject gc will not fully depend on this. 
this PR is a nice-to-have , but not must-to-have.

```
is_blk_alloced(blk_1)
free(blk_1)
assert(is_blk_allocated(blk_1)
```
also seems it is better to change `is_blk_alloced` to `is_blk_ever_alloced`, which means it is smaller than last_append_offset.   this will make the meaning more accurate

JacksonYao287:
@yamingk @xiaoxichen  can you help review this PR?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1318,9 +1308,24 @@ void RaftReplDev::on_fetch_data_received(intrusive< sisl::GenericRpcData >& rpc_
     folly::collectAllUnsafe(futs).thenValue(
         [this, rpc_data = std::move(rpc_data), sgs_vec = std::move(sgs_vec)](auto&& vf) {
             for (auto const& err_c : vf) {
-                if (sisl_unlikely(err_c.value())) {
+                RD_REL_ASSERT(err_c.hasValue(), "should have an error code in fetching data");
+                const auto& err = err_c.value();
+                if (sisl_unlikely(err)) {
                     COUNTER_INCREMENT(m_metrics, read_err_cnt, 1);
-                    RD_REL_ASSERT(false, "Error in reading data");
```
xiaoxichen, line 1323: have we ever hit here? Yawen and I did the test before and as the comment indicate it fails in IOMgr.

And in current implementation if we return an empty image, wont the client (folllower) panic?

```
void RaftReplDev::handle_fetch_data_response(sisl::GenericClientResponse response,
                                             std::vector< repl_req_ptr_t > rreqs) {
    auto resp_blob = response.response_blob();
    auto raw_data = resp_blob.cbytes();
    auto total_size = resp_blob.size();

    COUNTER_INCREMENT(m_metrics, fetch_total_blk_size, total_size);

    RD_DBG_ASSERT_GT(total_size, 0, "Empty response from remote");
    RD_DBG_ASSERT(raw_data, "Empty response from remote");

    RD_LOGD(NO_TRACE_ID, "Data Channel: FetchData completed for {} requests", rreqs.size());
```

JacksonYao287: yes, I hit here once, but only once.

```
RD_DBG_ASSERT_GT(total_size, 0, "Empty response from remote"); RD_DBG_ASSERT(raw_data, "Empty response from remote");
```
yes , this will make follower panic if empty image received.  there two are very old assert, which was made one year ago. we can remove it add return directly if got empty blob?


theoretically，we might encounter read error [here](https://github.com/eBay/HomeObject/blob/1923a585edaa12a3147dbf127db61a5e337ebf11/src/lib/homestore_backend/replication_state_machine.cpp#L764).

xiaoxichen (approved):

Issue #738, opened 2025-06-04
Replace member pre-check does not support non-3-replica env


Besroy:
In a 5-replica replace member process, starting the replace member is allowed even when there are only 2 active peers. However, once an active member out and flip to a learner, it does not participate in elections, making it impossible to meet the commit quorum requirement. Consequently, subsequent logs (e.g., set_priority) cannot be committed, causing the replace member operation to be canceled due to timeout, returning an INVALID_ARG error. 
Consider the following example:

1. Five replicas, M1 through M5, exist. Replicas M1, M2, and M3 have a committed LSN of 1000. Replicas M4 and M5 have a committed  LSN of 100. M1 is the leader.
2. M2 is selected for removal and set as a learner with LSN 1001. After the flip is complete, M1, M2, and M3 have LSN 1001, while M4 and M5 have LSN 100.
3. M2's priority is set to 0 with LSN 1002, but this change cannot be committed due to an insufficient quorum of only two active peers. (Only M1 and M3 are active. M4 and M5 are behind, and M3 is a learner.)
4. Learner status checks fail as the learner's priority remains non-zero during the timeout, causing the member replacement operation to be cancelled.
More details in [issue51](https://docs.google.com/document/d/16DMqv9J-JuNs5c25IBuX01QXe5WbxPm4B4hAfz4JNZM/edit?tab=t.0#heading=h.6afohvvjhmpv)

PR #741, opened 2025-06-09
Enhance quorum check in replace_member

yuwmao:
Related issue: https://github.com/eBay/HomeStore/issues/738


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1541,6 +1542,15 @@ std::set< replica_id_t > RaftReplDev::get_active_peers() const {
     return res;
 }
 
+uint32_t RaftReplDev::get_quorum_for_commit() const {
```
xiaoxichen, line 1545: can I take this as a short cut instead of exposing through nuraft_mesg `int32 raft_server::get_quorum_for_commit()` ?

:)

xiaoxichen (approved):

Besroy (approved):
Thanks for the enhancement!

PR #745, opened 2025-06-12
add on_reject to repl_req_ctx

JacksonYao287:
if leader try to propose a new log to raft and before this log is appended to log store, this leader changes to follower, then raft_server#drop_all_pending_commit_elems will be called, in which this new log(request) will be dropped. this means this req will be rejected, and will not be committed , nor rolled back. in this case, we need notify upperlayer through this virtual function.

in homeobject, we can use this callback to notify the upper layer. otherwise, it will stuck there. for example, for put_blob, if this req is rejected, then the ThenValue of put blob will not be called and the request will never be completed, since we set value for the promise in the put_blob_rreq only in on_commit or on_rollback

### DIFF
--- a/src/lib/replication/repl_dev/raft_state_machine.cpp
+++ b/src/lib/replication/repl_dev/raft_state_machine.cpp
@@ -45,6 +45,7 @@ ReplServiceError RaftStateMachine::propose_to_raft(repl_req_ptr_t rreq) {
     if (append_status && !append_status->get_accepted()) {
         RD_LOGE(rreq->traceID(), "Raft Channel: Failed to propose rreq=[{}] result_code={}", rreq->to_compact_string(),
                 append_status->get_result_code());
+        rreq->on_reject();
         return RaftReplService::to_repl_error(append_status->get_result_code());
     }
     return ReplServiceError::OK;
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.16.4"
+    version = "6.17.0"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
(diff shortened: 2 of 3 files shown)


src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -45,6 +45,7 @@ ReplServiceError RaftStateMachine::propose_to_raft(repl_req_ptr_t rreq) {
     if (append_status && !append_status->get_accepted()) {
         RD_LOGE(rreq->traceID(), "Raft Channel: Failed to propose rreq=[{}] result_code={}", rreq->to_compact_string(),
                 append_status->get_result_code());
+        rreq->on_reject();
```
Besroy, line 48: Thanks for the fix! But I'm not sure if it addresses issue #55. I haven't seen any error log about propose failures in the leader's log, and the pending put_blob requests seem have completed the propose->append->end_of_append_batch phase.

xiaoxichen (commented):
Could you please provide more information about the issue ?

```
```

IIUC any failure in propose will go to handle_error where we fulfills the future.

Besroy:
> Could you please provide more information about the issue ?
> 
> ```
>                     auto raft_status = m_state_machine->propose_to_raft(rreq);
>                     if (raft_status != ReplServiceError::OK) { handle_error(rreq, raft_status); }
> ```
> 
> IIUC any failure in propose will go to handle_error where we fulfills the future.

What SH sees is: After receiving the shutdown signal, the leader becomes a follower, and NuRaft drops all pending commit requests during the transition. This results in HS's pending requests never being committed. However, HS does not perceive that all pending requests are canceled on the NuRaft side and continues waiting for these requests to be completed, causing the shutdown process to be perpetually stuck in the waiting phase. More details in [issue55](https://docs.google.com/document/d/16DMqv9J-JuNs5c25IBuX01QXe5WbxPm4B4hAfz4JNZM/edit?tab=t.0#heading=h.jdg1yew49tya)

PR #746, opened 2025-06-13
Add a get_replace_member_status API for upper layer to query status

yuwmao:
- get_replace_member_status should always be called to leader.
- Add a task_id in replace member implementation to track the status.

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -1530,10 +1530,11 @@ void RaftReplDev::handle_error(repl_req_ptr_t const& rreq, ReplServiceError err)
 void RaftReplDev::start_replace_member(repl_req_ptr_t rreq) {
     auto ctx = r_cast< const replace_member_ctx* >(rreq->header().cbytes());
 
-    RD_LOGI(rreq->traceID(), "Raft repl start_replace_member commit member_out={} member_in={}",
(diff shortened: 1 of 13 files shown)


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -349,20 +371,111 @@ AsyncReplResult<> RaftReplDev::complete_replace_member(const replica_member_info
...
+    bool found_in = false;
+    for (auto p : peers) {
+        if (p.id_ == member_out.id) {
+            out_peer_info = p;
+            found_out = true;
+        } else if (p.id_ == member_in.id) {
+            found_in = true;
+        }
+    }
+    // Task cleaned, out_member removed, in_member added.
```
xiaoxichen, line 417: suggest re-group the if statements as

```
bool intent_completed = !found_out && found_in;
if (task_id.is_null() ) {
    if(intent_completed) { ...}
    else {
       return ReplaceMemberStatus::UNKNOWN;
    } 
}

if (m_rd_sb->replace_member_task.task_id != task_id) { 
      return ReplaceMemberStatus::TASK_ID_MISMATCH; 
}

if (intent_completed) {
    LOGGING?
}
return ReplaceMemberStatus::IN_PROGRESS;
```

xiaoxichen (commented):
lgtm aside from a small suggestion.

xiaoxichen (dismissed):
lgtm

An un-related feeling is that we may need adding RAII for desc_pending_request_num/incr_pending_request_num

PR #748, opened 2025-06-15
return error blk if no_space_left happens

JacksonYao287:
when we got no_space_left error when allocating blk in a specifc chunk, we need let the upper layer know this through the out_blk_id, this is crucial for handling no space left case. so that:

1 upper layer can check the returned status of alloc_blks to be aware of whether the blk is allocated successfully.
2 if the return status is no success, from the out_blk_id, the upper layer can know which chunk fails to allocate blk.

this bug is found in emergent gc test.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.16.4"
+    version = "6.16.5"
 
     homepage = "https://github.com/eBay/Homestore"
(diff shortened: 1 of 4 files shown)


xiaoxichen (approved):

conanfile.py
```diff
@@ -9,7 +9,7 @@
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.17.0"
+    version = "6.17.1"
```
xiaoxichen, line 12: this should be 6.18 however as we cannot proper do semantic versioning with GH... I dont really care

PR #754, opened 2025-06-18
Issue 753 Make CP number Fibers dynamic config

### LINKED ISSUE #753: Make CP Number Fibers to dynamic config
Change harded coded number CP fibers to dynamic config so that we can do perf tuning for HomeStore consumers.

yamingk:
Changes:
1. Make CP number Fibers dynamic config
2. Fix asan issue which is a race during shutdown that resoruce mgr trigger solo repl dev truncation.

### DIFF
--- a/src/lib/checkpoint/cp_mgr.cpp
+++ b/src/lib/checkpoint/cp_mgr.cpp
@@ -302,7 +302,9 @@ void CPManager::start_cp_thread() {
     // Multiple sync_io fibers may acquire a thread-level mutex and perform synchronous I/O using io_uring.
     // This can block the fiber and allow other fibers to be scheduled.
     // If another fiber tries to acquire the same mutex, a deadlock can occur.
-    iomanager.create_reactor("cp_io", iomgr::INTERRUPT_LOOP, 2u, [this, ctx](bool is_started) {
+    auto const num_fibers = HS_DYNAMIC_CONFIG(generic.cp_io_fibers); // default: 2
+    LOGINFO("Starting CP IO fibers with count: {}", num_fibers);
+    iomanager.create_reactor("cp_io", iomgr::INTERRUPT_LOOP, num_fibers, [this, ctx](bool is_started) {
         if (is_started) {
             {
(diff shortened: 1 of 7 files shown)


sanebay (dismissed):
lgtm

src/lib/common/resource_mgr.cpp
```diff
@@ -46,7 +48,21 @@ void ResourceMgr::stop() {
 //    writes on this descriptor;
 //
 void ResourceMgr::trigger_truncate() {
+    if (m_is_stopped_.load()) {
```
JacksonYao287, line 51: NIT: use memory_order_release for line 35 and memory_order_acquire for this line.

using there two memory order , we can 100% guarantee if line 51 can be aware of the change of 
`m_is_stopped_`

JacksonYao287 (commented):
LG, thanks for the fix, only two minor comments, FYI

src/lib/checkpoint/cp_mgr.cpp
```diff
@@ -302,7 +302,9 @@ void CPManager::start_cp_thread() {
     // Multiple sync_io fibers may acquire a thread-level mutex and perform synchronous I/O using io_uring.
     // This can block the fiber and allow other fibers to be scheduled.
     // If another fiber tries to acquire the same mutex, a deadlock can occur.
-    iomanager.create_reactor("cp_io", iomgr::INTERRUPT_LOOP, 2u, [this, ctx](bool is_started) {
+    auto const num_fibers = HS_DYNAMIC_CONFIG(generic.cp_io_fibers); // default: 2
```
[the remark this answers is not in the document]
yamingk: as mentioned in the comment it is defaulted to 2 and has no impact to exiting project. Thx for the heads up.

src/lib/common/resource_mgr.cpp
yamingk: the default behavior of load() and operator= is equivalent of std::memory_order_seq_cst garunteens.

JacksonYao287 (approved):
LGTM

PR #757, opened 2025-06-25
 Avoid dup-append logs for committing logs

Besroy:


Besroy:
This pr aims to resole a corner case, but using last_lsn/last_term rather than last_commit_lsn might be more dangerous and requires more consideration, so covert it to raft ATM.

PR #758, opened 2025-06-26
Change task_id to string

yuwmao:
Changing task_id to string allows users to use more meaningful information as id.
In NuObject2.0, CM will use "timestamp:pg_id" as task_id for easy time based range query in its kvstore.


Besroy (dismissed):
LGTM

xiaoxichen (dismissed):

PR #761, opened 2025-07-01
Run cp mgr timer in its own reactor.

sanebay:
Add reactor for cp manager timer. In high intensive IO tests, cp timer's are not executed.


raakella1 (approved):
LG!

PR #770, opened 2025-07-14
fetch data from leader

JacksonYao287:
1 fetch data from leader , not leader, to avoid the case that originator is the out_member when replace member happens

2 if error happens when handling fetch_data request, return empty mesage and let follower retry. don`t crash

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -1206,7 +1206,28 @@ void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
     shared< flatbuffers::FlatBufferBuilder > builder = std::make_shared< flatbuffers::FlatBufferBuilder >();
     RD_LOGD(NO_TRACE_ID, "Data Channel : FetchData from remote: rreq.size={}, my server_id={}", rreqs.size(),
             server_id());
-    auto const& originator = rreqs.front()->remote_blkid().server_id;
+
+    int32_t fetch_from_peer_id{0};
+
+    if (RaftReplDev::is_fetch_data_only_from_originator()) {
+        // in raft_repl_dev UT, we do not implement the header parser, so it has to fetch data from originator.
+        fetch_from_peer_id = rreqs.front()->remote_blkid().server_id;
+    } else {
+        auto const& srv_configs = raft_server()->get_config()->get_servers();
+        std::vector< int32_t > peer_ids;
+        for (const auto& srv_config : srv_configs) {
+            auto peer_id = srv_config->get_id();
+            if (peer_id != m_raft_server_id) peer_ids.emplace_back(peer_id);
+        }
+        const auto available_peer_size = peer_ids.size();
+        if (!available_peer_size) {
+            RD_LOGE(NO_TRACE_ID, "Data Channel: No available peer to fetch data from, ignoring this fetch call");
+            return;
+        }
+
+        // here we use time as the random generator to select a random peer
+        fetch_from_peer_id = peer_ids[static_cast< long int >(std::time(nullptr)) % available_peer_size];
+    }
 
     for (auto const& rreq : rreqs) {
         entries.push_back(CreateRequestEntry(*builder, rreq->lsn(), rreq->term(), rreq->dsn(),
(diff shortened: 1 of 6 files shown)


xiaoxichen (commented):
LGTM 

It will be interesting if we allow to fetch from non-leader which will offload the IO pressure of fetch_data on leader. cc @Besroy 

But that can be done in other track.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1354,10 +1352,12 @@ void RaftReplDev::handle_fetch_data_response(sisl::GenericClientResponse respons
     auto raw_data = resp_blob.cbytes();
     auto total_size = resp_blob.size();
 
-    COUNTER_INCREMENT(m_metrics, fetch_total_blk_size, total_size);
+    if (!raw_data || total_size == 0) {
+        RD_LOGW(NO_TRACE_ID, "Data Channel: FetchData returned empty payload, ignoring");
+        return;
```
Besroy, line 1357: JFYI: If a fetch is called during the end_of_append_batch and the leader returns empty data, the follower currently does not retry. After expiration, it will trigger the assertion at `notify_after_data_written:HS_REL_ASSERT(false, "Data fetch timeout, should not happen");`. However, this seems to be an existing issue; perhaps we could add a TODO for a retry mechanism.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1195,24 +1183,18 @@ void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
...
+    const auto& leader_server_id = nuraft_mesg::to_server_id(get_leader_id());
 
     for (auto const& rreq : rreqs) {
         entries.push_back(CreateRequestEntry(*builder, rreq->lsn(), rreq->term(), rreq->dsn(),
                                              builder->CreateVector(rreq->header().cbytes(), rreq->header().size()),
                                              builder->CreateVector(rreq->key().cbytes(), rreq->key().size()),
-                                             rreq->remote_blkid().server_id /* blkid_originator */,
+                                             leader_server_id,
                                              builder->CreateVector(rreq->remote_blkid().blkid.serialize().cbytes(),
                                                                    rreq->remote_blkid().blkid.serialized_size())));
```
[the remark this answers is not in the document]
xiaoxichen: chunk reclaim is fine as validate_blob will fail.  Bad drive (or degrade mode) is a concern..... lets create an issue and wait zhiteng's POC on single-disk-mode....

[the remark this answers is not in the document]
Besroy: Created a issue https://github.com/eBay/HomeObject/issues/330 in HO

JacksonYao287:
@Besroy @xiaoxichen 

I have changed the policy of fetching data to fetch data from a random peer. as a result , it need the host to be capable of parsing user_header and get the correct blob. However, In the current raft_repl_dev UT, it use the default implementation of fetch_data, which will directly fetch data by blk_id and not do any user_header parsing, so this will lead to some data verification error. to support this case, I also add a choice to also support the case of only fetching data from originator.

xiaoxichen (commented):
lgtm aside from a nit

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1195,7 +1183,28 @@ void RaftReplDev::fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs) {
...
+
+    if (RaftReplDev::is_fetch_data_only_from_originator()) {
+        // in raft_repl_dev UT, we do not implement the header parser, so it has to fetch data from originator.
+        fetch_from_peer_id = rreqs.front()->remote_blkid().server_id;
+    } else {
+        auto const& srv_configs = raft_server()->get_config()->get_servers();
+        std::vector< int32_t > peer_ids;
+        for (const auto& srv_config : srv_configs) {
+            auto peer_id = srv_config->get_id();
+            if (peer_id != m_raft_server_id) peer_ids.emplace_back(peer_id);
```
Besroy, line 1197: fetch data might fail and then generate some garbage here if the selected peer doesn't have data in the following case, but anyway, raft will retry:
T1: leader push blob=100 to F1 and F2
T2: F1 and F2 cannot alloc blk because the related shard not committed
T3: F1 append log, fetch data from F2
T4: F2 append log, fetch data from F1
then F1 and F2 will failed to get data, waiting raft retry and select the leader

xiaoxichen: Thanks this is a good point.

Maybe we should only fetch data when we are in resync mode ? in that case (only consider 3 copies) leader and another follower both committed.

Lets track this and verify in SH testing,  and decide based on metrics, 

- how much improvement we see in fetch latency ?
- how many more garbages this approach create vs previous?

JacksonYao287:
I find a read verification issue in storage hammer testing. I need to confirm whether it is caused by recent PRs. so I will hold off on merging the PR until I  figure out what happens. cc @Besroy @xiaoxichen

PR #781, opened 2025-07-26
flush a buffer only if it is dirtied in the current cp

raakella1:


src/lib/index/wb_cache.cpp
```diff
@@ -1023,7 +1023,8 @@ void IndexWBCache::get_next_bufs_internal(IndexCPContext* cp_ctx, uint32_t max_c
         std::optional< IndexBufferPtr > buf = cp_ctx->next_dirty();
         if (!buf) { break; } // End of list
```
shosseinimotlagh, line 1025: Add comments to know what was the problem (buffer reuse in overlapping cp)

shosseinimotlagh (changes requested):

shosseinimotlagh (dismissed):

sanebay (approved):

PR #788, opened 2025-08-07
Support rolling back replace member operation

yuwmao:
Add four functions to implement rollback logic:
1. remove_member: It will call `on_remove_member` cb because HO also maintains the membership.
2. flip_learner_flag
3. clean_replace_member_task
4. list_replace_member_task: This is used for verifying the result of 3.

Also add a destroy_repl_dev api to force destroy local repl dev. This helps handle some pgmove corner cases and https://github.com/eBay/HomeStore/issues/823

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -497,6 +497,9 @@ AsyncReplResult<> RaftReplDev::remove_member(const replica_id_t& member, uint32_
     }
     RD_LOGI(trace_id, "Remove member step1. Member has been removed, member={}", boost::uuids::to_string(member));
     // 2. propose to raft to remove member
+    // Note: The removed member will not receive this HS_CTRL_REMOVE_MEMBER log entry since it has already been
+    // removed from the raft group in step 1. Therefore, on_remove_member callback will only be executed on the
+    // remaining members.
     RD_LOGI(trace_id, "Remove member step2. Propose to raft for HS_CTRL_REMOVE_MEMBER req, group_id={}",
             group_id_str());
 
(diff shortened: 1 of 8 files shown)


src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -559,6 +559,70 @@ AsyncReplResult<> RaftReplService::flip_learner_flag(group_id_t group_id, const
         });
 }
 
+AsyncReplResult<> RaftReplService::remove_member(group_id_t group_id, const replica_id_t& member,
+                                                 uint32_t commit_quorum, bool wait_and_verify,
+                                                 uint64_t trace_id) const {
+    if (is_stopping()) return make_async_error<>(ReplServiceError::STOPPING);
+    incr_pending_request_num();
```
JacksonYao287: nit, pls use init_req_ctx instead of this manual increase and decrease

src/include/homestore/replication/repl_dev.h
```diff
@@ -387,6 +389,9 @@ class ReplDevListener {
     virtual void on_complete_replace_member(const std::string& task_id, const replica_member_info& member_out,
                                             const replica_member_info& member_in, trace_id_t tid) = 0;
 
+    /// @brief Called when remove a member.
+    virtual void on_remove_member(const replica_id_t& member, trace_id_t tid) = 0;
```
JacksonYao287: we will implement this in homeobject side , right?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1614,6 +1721,28 @@ void RaftReplDev::complete_replace_member(repl_req_ptr_t rreq) {
...
 }
 
+void RaftReplDev::remove_member(repl_req_ptr_t rreq) {
+    auto member = *r_cast< const replica_id_t* >(rreq->header().cbytes());
+    RD_LOGI(rreq->traceID(), "Raft repl remove_member commit, member={}", boost::uuids::to_string(member));
+    m_listener->on_remove_member(member, rreq->traceID());
+}
+
+void RaftReplDev::clean_replace_member_task(repl_req_ptr_t rreq) {
+    auto task_id = std::string(r_cast< const char* >(rreq->header().cbytes()));
```
JacksonYao287: one stupid question , why not use uint64 instead of string for task_id. that is more simple to persist.

to keep align with cm proto?

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -503,9 +481,55 @@ ReplServiceError RaftReplDev::do_add_member(const replica_member_info& member, u
...
+    RD_LOGI(trace_id, "Remove member, member={}", boost::uuids::to_string(member));
+    if (is_stopping()) {
+        RD_LOGI(trace_id, "repl dev is being shutdown!");
+        return make_async_error<>(ReplServiceError::STOPPING);
+    }
+    init_req_counter counter(pending_request_num);
+
+    if (commit_quorum >= 1) { reset_quorum_size(commit_quorum, trace_id); }
+    // 1.remove member
+    auto ret = do_remove_member(member, wait_and_verify, trace_id);
```
JacksonYao287: should we check if there is a on-going replace member task about this member before we take this action?

for example , the member should not be a move_in or move_out member of a on-going replace member task.

src/include/homestore/replication/repl_dev.h
yuwmao: Yes, I bump to 7.1.0 and HO can still using 7.0.x for build until HO adaptive changes PR comes out.

src/lib/replication/repl_dev/raft_repl_dev.cpp
yuwmao: This is mainly used in PGMove rollback process.
If the in_member is stuck in BR, we need to rollback this replace_member action as following:
1. remove the in_member
2. clear the task sb

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -571,6 +599,17 @@ ReplaceMemberStatus RaftReplService::get_replace_member_status(group_id_t group_
         ->get_replace_member_status(task_id, member_out, member_in, others, trace_id);
 }
 
+ReplServiceError RaftReplService::destroy_repl_dev(group_id_t group_id, uint64_t trace_id) {
```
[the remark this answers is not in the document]
yuwmao: Yes

src/lib/replication/service/raft_repl_service.cpp
```diff
@@ -516,22 +509,15 @@ AsyncReplResult<> RaftReplService::replace_member(group_id_t group_id, std::stri
...
     return std::dynamic_pointer_cast< RaftReplDev >(rdev_result.value())
         ->start_replace_member(task_id, member_out, member_in, commit_quorum, trace_id)
         .via(&folly::InlineExecutor::instance())
         .thenValue([this](auto&& e) mutable {
-            if (e.hasError()) {
-                decr_pending_request_num();
-                return make_async_error<>(e.error());
-            }
-            decr_pending_request_num();
+            if (e.hasError()) { return make_async_error<>(e.error()); }
```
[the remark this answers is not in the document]
yuwmao: In my first version, I moved counter into thenValue, but I found it would be deconstructed twice, so `pending_request_num` will increase once but decrease twice unless we revise the init_req_counter. 
Then I realized `thenValue` in `raft_repl_service` only convert the error, so maybe we don't need to protect it?Since it's simpler, that's why I didn't capture.

JacksonYao287 (dismissed):

fishfc:
The code coverage is 70% only. We need to get to 80%.

PR #789, opened 2025-08-07
SDSTOR-18915: Check if the devices are unique during boot

raakella1:
If the device list that we pass on to homestore has duplicates, we do not detect it and will have data corruption. Verify and add assert to cover this case.

### DIFF
--- a/src/lib/device/device_manager.cpp
+++ b/src/lib/device/device_manager.cpp
@@ -115,7 +115,10 @@ void DeviceManager::format_devices() {
     // Get common iomgr_attributes
     for (auto& dinfo : m_dev_infos) {
         format_single_device(dinfo);
-    }
+    } 
+
+    // Verify the first blocks to see if the devs are unique
+    HS_REL_ASSERT(verify_unique_devs(), "Found duplicate physical devices in the system");
 }
 
 uint32_t DeviceManager::format_single_device(dev_info& dinfo) {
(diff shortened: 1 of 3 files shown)


src/lib/device/device_manager.cpp
```diff
@@ -154,6 +157,34 @@ uint32_t DeviceManager::format_single_device(dev_info& dinfo) {
     return pdev_id;
 }
 
+bool DeviceManager::verify_unique_devs() const {
+    bool ret = true;
+    std::unordered_map< uint32_t, std::pair< pdev_info_header, uint32_t > > pdev_info_map;
+    for (auto& pdev : m_all_pdevs) {
+        if (!pdev) { continue; }
+        auto buf = hs_utils::iobuf_alloc(hs_super_blk::first_block_size(), sisl::buftag::superblk, 512);
```
sanebay, line 165: Need iobuf_free

src/lib/device/device_manager.cpp
```diff
@@ -154,6 +157,34 @@ uint32_t DeviceManager::format_single_device(dev_info& dinfo) {
...
+        if (auto err = pdev->read_super_block(buf, hs_super_blk::first_block_size(), hs_super_blk::first_block_offset()); err) {
+            LOGERROR("Failed to read first block from device={}, error={}", pdev->get_devname(), err.message());
+            ret = false;
+            continue;
+        }
+        auto pdev_hdr = (r_cast< first_block* >(buf))->this_pdev_hdr;
+        if (pdev_info_map.find(pdev_hdr.pdev_id) != pdev_info_map.end()) {
```
sanebay, line 172: why do we need a pair of pdev_info_header. Wont a count be enough ?. then you can simply do pdev_info_map[pdev_hdr.pdev_id]++

src/lib/device/device_manager.cpp
raakella1: ack

src/lib/device/device_manager.cpp
raakella1: I added the header for logging the header before assert

sanebay (approved):

PR #790, opened 2025-08-08
support disable push_data by rreq

JacksonYao287:
extend rreq to support disable push_data if rreq has linked data. this is useful in some scenario of homeobject, and maybe also useful for other cases in the future


Besroy (approved):
lgtm

PR #796, opened 2025-08-28
SDSTOR-18909: Fix bugs in btree merge code

raakella1:
We are creating empty new nodes during btree merge due to a bug which is causing an assert. Added UT to reproduce the issue and fixed the bug.


JacksonYao287:
@shosseinimotlagh  can you review this PR?

PR #797, opened 2025-09-02
Add reconcile_leader function in case of imbalanced leader distribution

Besroy:


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1697,6 +1697,46 @@ std::vector< peer_info > RaftReplDev::get_replication_status() const {
...
+
+    if (expected_leader_id == m_my_repl_id) {
+        RD_LOGD(NO_TRACE_ID, "Current leader {} has highest priority {}, no need to reconcile", get_leader_id(),
+                my_priority);
+        return;
+    }
+
+    RD_LOGI(NO_TRACE_ID, "Current leader {} is different from expected leader {}, reconciling it, my id={}",
+            get_leader_id(), expected_leader_id, m_my_repl_id);
+    raft_server()->yield_leadership(false /* immediate_yield */, -1 /* successor_id */);
```
xiaoxichen, line 1736: why -1 instead of expected_leader_id?

xiaoxichen (approved):

Issue #810, opened 2025-09-24
Save incorrect data during leader re-election


Besroy:
Found a data corruption issue in storage hammer test. This issue occurs under specific conditions involving leader re-election and log term changes, leading to incorrect data being saved and committed.

### **Root Cause:**

- The leader re-creates the rreq with new blocks after a term change but does not copy the data into the new blocks because it is the proposer.
- The fetch-data logic on the leader reads and returns incorrect data from the uncommitted state.
- The follower uses the wrong remote blkid during the fetch-data process, leading to the storage of incorrect data.

### **On leader side:**
T1: leader receives a put_blob request for blob_id=3054. It allocates blocks: [[{blk#=24583, count=4097, chunk=215}]] with dsn=4199 and term=10.
```
[09/22/25 02:54:33.385] [storage_mgr] [debug] [77] [hs_blob_manager.cpp:115:_put_blob] [traceID=6866359719276110938,shardID=0x4e,pg=0,shard=0x4e,blob=3054] Blob Put request: pg=0, group=ff9c6a28-11c7-4d32-8c36-1430a5c78b2e, shard=0x4e, length=16777216
[09/22/25 02:54:33.386] [storage_mgr] [trace] [77] [hs_blob_manager.cpp:188:_put_blob] [traceID=6866359719276110938,shardID=0x4e,pg=0,shard=0x4e,blob=3054] Put blob: header=magic=0x21fdffdba8d68fc6 version=1 shard=0x4e blob_size=16777216 user_size=0 algo=1 hash=ea 47 1a 79 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
[09/22/25 02:54:33.386] [storage_mgr] [debug] [77] [hs_blob_manager.cpp:431:blob_put_get_blk_alloc_hints] [traceID=6866359719276110938,shardID=0x4e,pg=0,shard=0x4e,blob=3054] Picked p_chunk_id=215, reserved_blks=16384
[09/22/25 02:54:33.386] [storage_mgr] [debug] [77] [raft_repl_dev.cpp:784:async_alloc_write] [traceID=6866359719276110938] [rdev0:ff9c6a28-11c7-4d32-8c36-1430a5c78b2e] repl_key [server=1055378172, term=10, dsn=4199, hash=1055382161], header size [48] bytes, user_key size [8] bytes, data size [16781312] bytes
```
T2: leader push data to all followers
```
[...]

PR #812, opened 2025-09-25
add rreq check before appending logs

JacksonYao287:


JacksonYao287:
this PR tries to solve this issue #810

PR #813, opened 2025-09-25
add rreq check before appending logs

JacksonYao287:
this PR tried to fix issue #810 .
before append log, we check if the corresponding rreq exist . if not , we drop this request. so , when term changes between init_req_ctx and propose_to_raft, the request will be rejected

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.20.25"
+    version = "6.20.26"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/lib/replication/log_store/repl_log_store.cpp
+++ b/src/lib/replication/log_store/repl_log_store.cpp
@@ -15,8 +15,27 @@ uint64_t ReplLogStore::append(nuraft::ptr< nuraft::log_entry >& entry) {
         return lsn;
     }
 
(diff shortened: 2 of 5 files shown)


xiaoxichen (commented):
1. I think we should use the `expected_term` on leader which solves the conflict  ( by changing the append_entries to append_entries_ext)
https://github.com/eBay/NuRaft/blob/68e36210e0dc224852890c90517fbb33c9365ac5/include/libnuraft/raft_server.hxx#L306

dont need touching `nuraft::cb_func::Type::ProcessReq`

2. on follower side I dont get the case that rreq doesnt exists --- the term was given by leader in the log,  the rreq created by `localize_journal_entry_prepare` looking at the term from log entries.  

The other changes to init_req_ctx looks good

Besroy:
Curious why the choice was made to crash when terms are inconsistent, rather than handling the case where the proposer has different blocks by reading data from `remote_blk` and writing it to newly allocated blocks. In fact, I’m somewhat concerned this could lead to unexpected frequent crashes. During storage hammer testing, there were cases where leader re-election caused `rreq` to be missing, as well as cases where a follower’s `rreq` was gc, resulting in a missing `rreq`. There might be additional corner cases that could cause members to crash.

Besroy:
Thanks @xiaoxichen  This is awesome!

xiaoxichen:
This is a very nice catch , kudo's to @JacksonYao287 and @Besroy .

Lets get this merge asap

xiaoxichen (approved):
lgtm

PR #816, opened 2025-09-28
fix remote_blk_id

JacksonYao287:


xiaoxichen (commented):
```
    // If we are not able to locate that req in the map for this entry, it means that no entry from raft leader is
    // appended and data channel has not received a data. This is typical scenario if we do unpack().
```

It looks to me like we can do 
```
RELEASE_ASSERT(rreq, "Failed to find in-memory rreq for repl_key={} before appending log", rkey.to_string());
```
as we dont use pack/unpack at all

src/lib/replication/repl_dev/raft_state_machine.cpp
```diff
@@ -164,6 +164,12 @@ repl_req_ptr_t RaftStateMachine::localize_journal_entry_finish(nuraft::log_entry
         .server_id = jentry->server_id, .term = lentry.get_term(), .dsn = jentry->dsn, .traceID = jentry->traceID};
 
     auto rreq = m_rd.repl_key_to_req(rkey);
+
+    if (is_append_log) {
+        // make sure the rreq exists before appending the log, both leader and follower
+        RELEASE_ASSERT(rreq, "Failed to find in-memory rreq for repl_key={} before appending log", rkey.to_string());
```
Besroy, line 170: This `assert` appears to conflict with the logic in lines L173-L184. If this `assert` is added, shouldn't the conditions below it be impossible? Alternatively, if the conditions in the `if` statement below are allowed, should this `assert` be removed?

JacksonYao287: I get your point. I don`t want to touch the existing code, which might make the logic hard to understand and bring some unknown bug,  so add a separate and quick assert outside the logic you mentioned!

we can even remove this assert one day in the future.

JacksonYao287: but, if we create a new rreq, then that will be very dangerous. since the data has been written to the blk of the older rreq, which have been gc. for the newly created rreq, a new blk is allocated, and we lose the data?

xiaoxichen: we  only GC those rreq whose `DSN <  m_next_dsn.load()`, suppose should not

JacksonYao287: yes, let`s add this assert here to see if there is any error case happen

xiaoxichen (approved):

PR #817, opened 2025-09-29
fix init rreq

JacksonYao287:
we should do a deep copy for the user header , otherwise if the input user_header is freed after the call is returned, then m_header will become a dangling blob(the pointer in it is a dangling pointer).

> why homestore start_replace_member/complete_replace_member(header involved) works well with this error

previously, we do not create and add rreq in to m_repl_key_req_map before propose_to_raft. the rreq is created and inserted into rreq when applier_create_req, where we created the rreq from the raft log entry, and the the in-memory log entry is available when the rreq#header is used, so the shallow copy works.

but now we create and add rreq in to m_repl_key_req_map before propose_to_raft.

so for follower , it works well since it is still created in applier_create_req

but for leader, the rreq , where the [header blob](https://github.com/eBay/HomeStore/blob/2f0e74d8147f1dbacbb981d4c71cff837167f573/src/lib/replication/repl_dev/raft_repl_dev.cpp#L250) is destroyed after we propose_to_raft. so the header of the rreq in m_repl_key_req_map will become a dangling blob.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "6.20.27"
+    version = "6.20.28"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -72,7 +72,8 @@ struct repl_key {
 
     bool operator==(repl_key const& other) const = default;
     std::string to_string() const {
-        return fmt::format("server={}, term={}, dsn={}, hash={}", server_id, term, dsn, Hasher()(*this));
+        return fmt::format("server={}, term={}, dsn={}, hash={}, trace_id={}", server_id, term, dsn, Hasher()(*this),
+                           traceID);
     }
 };
 
--- a/src/lib/replication/log_store/repl_log_store.cpp
+++ b/src/lib/replication/log_store/repl_log_store.cpp
@@ -15,7 +15,7 @@ uint64_t ReplLogStore::append(nuraft::ptr< nuraft::log_entry >& entry) {
         return lsn;
     }
 
-    repl_req_ptr_t rreq = m_sm.localize_journal_entry_finish(*entry, true /* is_append_log */);
+    repl_req_ptr_t rreq = m_sm.localize_journal_entry_finish(*entry);
     RELEASE_ASSERT_NE(nullptr != rreq, "Failed to localize journal entry before appending log");
 
     ulong lsn = HomeRaftLogStore::append(entry);
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -260,10 +260,13 @@ AsyncReplResult<> RaftReplDev::start_replace_member(std::string& task_id, const 
 
     if (status != ReplServiceError::OK) {
         RD_LOGE(trace_id, "Initializing rreq failed, rreq=[{}], error={}", rreq->to_string(), status);
+        reset_quorum_size(0, trace_id);
+        decr_pending_request_num();
+        return make_async_error<>(ReplServiceError::RETRY_REQUEST);
     }
 
-    auto const [_, happened] = m_repl_key_req_map.emplace(rreq->rkey(), rreq);
-    RD_DBG_ASSERT(happened, "Duplicate repl_key={} found in the map", rreq->rkey().to_string());
+    // the rreq will be recreated when appending log, which is a follower-like flow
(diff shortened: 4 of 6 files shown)


Besroy:
Thank you very much for your fix! I am considering whether we might encounter a similar issue when the follower saves push data before raft channel here: https://github.com/eBay/HomeStore/blob/master/src/lib/replication/repl_dev/raft_repl_dev.cpp#L942

xiaoxichen (commented):
The intent of the fix is good. However it is very tricky to follow as the whole repl_ctx_t is all ptr based, for data, journal_buf, and every other field.  Why the header need special cares....

The logic flow strarts from 
1. we want to use `init_req_ctx` so that the quience state get horned for EGC.
2. The `init_req_ctx` creates rreq, we want to save it  into the map
3.  The buffer referenced is a local buffer, which can be released.
4.  We need a deep copy of header,  technically other fields need deep-copy as well, just we probabaly dont access them.

I would suggest break the flow from #2 or #3,   
option 1:   can we not save the rreq returned by `init_req_ctx` into map?  Then later it will be recreated by using the journal buf.
option 2:   similar to async_alloc_write etc, can we ensure the header_buf is live till the proposal  committed ?  i.e wait till the rreq is committed.  

need @yuwmao 's input  regarding wheter we should wait for step 3 completed before starting step 4.  Also IIUC the step 4 (add member) wait till committed,  that implies step 3 is commiited,  then till the point of setp 3 committed the header is valid,  why we are hitting issues here?

xiaoxichen:
The async_alloc_wrie is already working fine as the caller (HO ) will not relase the buffer. Now the problem is only for replcae member.  Can we focuse on fixing these APIs as a fix, instead of a major API refactoring that assuming caller can release everything after `propose_to_raft`... I dont know if the later is even a goal for us, 

Previously the replace member works fine as it doesnt insert the rreq into map at all and it magically works as it go through the applier_create_req on leader. This behavior is wired I agree.

xiaoxichen:
Based on the business logic of replace member, it is likely the option 2 is naturally reached.

JacksonYao287:
this error will affect all the apis that involves header, including start_replace_member/complete_replace_member and propose_truncate_boundary. (and potentially affect the apis that involves key in the future)

let me put more details here.

previously( before commit id c5d067fa881625bd0e425e23085704f835f8cd93), for the above three log-only(no data written) and header-involved and apis,  we create rreq for each api call. but this created rreq is only used as an input parameter of  `propose_to_raft`, and it is not inserted in to req_map.  instead, they will be recreated and inserted into the req_map when [append log and localize_journal_entry_finish](https://github.com/eBay/HomeStore/blob/c5d067fa881625bd0e425e23085704f835f8cd93/src/lib/replication/log_store/repl_log_store.cpp#L18). [here](https://github.com/eBay/HomeStore/blob/c5d067fa881625bd0e425e23085704f835f8cd93/src/lib/replication/repl_dev/raft_state_machine.cpp#L88-L90), the blobs of header and key of the newly created rreq points to the header and key in the in-memory log entry. this means if the in-memory log entry is destroyed , then the blobs of header and key will become dangling, **this is a potential issue.**

but, fortunately，the log entry is a shard_ptr, and [we put it into the log cache when appending log](https://github.com/eBay/HomeStore/blob/c5d067fa881625bd0e425e23085704f835f8cd93/src/lib/replication/log_store/home_raft_log_store.cpp#L176), and can ensure the key and value blob is not dangling until the corresponding log entry is evicted. so , theoretically, if handle_commit of a rreq is executed before the log entry is evicted, then everything will go well.
this is  how "it magically works" as you said.


but we made some changes recently and  the behavior changes for leader. now, the rreq is created and inserted into req_map before propose_to_raft. so , for the above three apis, the header and key blob in rreq will definitely become dangling at leader(unless the caller can ensure they are valid).

> Now the problem is only for replcae member. Can we focuse on fixing these APIs as a fix,

I got your point, but I think we should thoroughly address this issue, it is only about a specific api, it is more about the rreq mechanism

xiaoxichen:
Lets put solution aside and get a solid  problem statement,

Any buffer given to rreq, including header/key/data,  is volatile, meaning,  rreq will NOT do any deep copy but relies on the owner of the buffer to sustain till the EOL of the rreq.  

On follower,  
- `repl_req_ctx::change_raft_journal_buf`  moves owner of log buffer into rreq, that ensures header/key buffer in rreq is valid.
On leader, 
- For `async_alloc_write`,  the caller(HO) holds the buffer till the request finished,  no issue.
- For other operations mainly those log only `start_replace_member/complete_replace_member/propose_truncate_boundary`,  the caller didnt hold the buffer properly,  the behavior is subject to detail of HS,  before commit id https://github.com/eBay/HomeStore/commit/c5d067fa881625bd0e425e23085704f835f8cd93,  the rreq was created in exactly same flow as folloer which works fine.  After this commit , we add the rreq into rreq_map  which bypassed the `localize_journal_entry_prepare`.

PR #839, opened 2025-12-08
add latch lsn

JacksonYao287:
there is a very corner case like following:

        T1: L1 is leader, push data to follower F1, F1 generate rreq1 and allocate blk for this data. let`s say the
        lsn of this data is lsn-100

        T2: leader switchs to L2, L2 send log to F1, F1 generate rreq2 and try to allocate blk but got
        no_space_left. then it will set the quiesce flag and waiting for the commit of lsn-99 (100 - 1).

        T3: leader switchs back to L1, L1 send log (lsn-100) to F1, when F1 tries to create rreq for lsn-100,  rreq1
        (which is created at T1) will be found since the rkey{server,term,dsn} is same as rreq1. so F1 will skip
        creating a new rreq. since the data for lsn-100 is already in written at T1, F1 start appending log entries
        to its log store, which will call logstore::append_log_entries and thus call localize_journal_entry_finish.

        T4: lsn-99 is committed at F1 and clear_chunk_req is called, so all the rreqs in F1 including rreq1 are
        cleared.

        T5: F1 call localize_journal_entry_finish, rreq1 is not found since it is cleared at T4, so F1 will try to
        create a new rreq for lsn-100. but now, F1 is in quiesce state, all the rreq creation will be rejected, so a
        nullptr will be returned. and cause RELEASE_ASSERT(rreq != nullptr) failure.

 when we got no_space_left in log channel at follower, we set this latch_lsn, so if any lsn in this batch that is >= latch_lsn, the whole batch will be rejected. after we successfully handle no_space_left and call resume_accepting_reqs, latch_lsn will be reset to max value.

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -177,9 +177,7 @@ AsyncReplResult<> RaftReplDev::start_replace_member(std::string& task_id, const 
         RD_LOGE(trace_id, "Step1. Replace member invalid parameter, out member is not found, task_id={}", task_id);
         return make_async_error<>(ReplServiceError::SERVER_NOT_FOUND);
     }
-    if (m_my_repl_id != get_leader_id()) {
-        return make_async_error<>(ReplServiceError::NOT_LEADER);
-    }
+    if (m_my_repl_id != get_leader_id()) { return make_async_error<>(ReplServiceError::NOT_LEADER); }
     // Check if leader itself is requested to move out.
     if (m_my_repl_id == member_out.id) {
         // immediate=false successor=-1, nuraft will choose an alive peer with highest priority as successor, and wait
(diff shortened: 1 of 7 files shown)


xiaoxichen (commented):
in general lgtm.  The corner case explanation can goes into commit message.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -2007,6 +2011,39 @@ nuraft::cb_func::ReturnCode RaftReplDev::raft_event(nuraft::cb_func::Type type,
...
+        // creating a new rreq. since the data for lsn-100 is already in written at T1, F1 start appending log entries
+        // to its log store, which will call logstore::append_log_entries and thus call localize_journal_entry_finish.
+
+        // T4: lsn-99 is committed at F1 and clear_chunk_req is called, so all the rreqs in F1 including rreq1 are
+        // cleared.
+
+        // T5: F1 call localize_journal_entry_finish, rreq1 is not found since it is cleared at T4, so F1 will try to
+        // create a new rreq for lsn-100. but now, F1 is in quiesce state, all the rreq creation will be rejected, so a
+        // nullptr will be returned. and cause RELEASE_ASSERT(rreq != nullptr) failure.
+
```
xiaoxichen: Suggest move these lines into commit message.

xiaoxichen (commented):
also considering adding more logs before latch_lsn.store() for the sake of debugability  (better to wrap it into a function ).

xiaoxichen (approved):
lgtm

Issue #835, opened 2025-12-04
Handle leader switch during Baseline Resync when new leader has all logs


Besroy:
### Problem Description

When a follower is catching up via Baseline Resync and a leader switch occurs where the new leader has all logs, the follower attempts to switch to incremental resync. However, this creates a stuck situation:
  1. During BR, the follower set last_snapshot_lsn as the snapshot lsn in [save_snp_resync_data](https://github.com/eBay/HomeStore/blob/cf805b91b7f090dbb636a115f7c760ab1f920aba/src/lib/replication/repl_dev/raft_repl_dev.cpp#L2429-L2436)
  2. The followers skip log commits when last_snapshot_lsn != 0 at [commit_ext](https://github.com/eBay/HomeStore/blob/cf805b91b7f090dbb636a115f7c760ab1f920aba/src/lib/replication/repl_dev/raft_state_machine.cpp#L207-L210)
This results in commit_lsn stuck at 0 while append_lsn is non-zero  (e.g., put_blob requests fail because shard create messages cannot commit)

The incremental resync being stuck seems to be a valid option . If it were allowed to proceed and another leader switch occurred during incremental resync:
  - The original BR process would resume (if the snapshot context still exists with same snapshot index)
  - This leads to unexpected corner cases such as blob leakage
  - Cleaning up the snapshot context to enable incremental resync requires significant efforts and introduces additional corner cases

### Current Workaround

Since the scenario where a member has all logs during BR is rare, the current workaround is: Manually delete the current leader pod to trigger re-election, then the follower can catch up by continuing the BR process

**Additional Issue**
Even if we have a workaround to let BR continue, there are potential problems:
  - Incremental resync allocates space during append log operations, which means BR may encounter no_space_left errors and cannot proceed
  - If incremental resync has appended logs up to LSN_A, while the new leader starts at LSN_B where LSN_B <= LSN_A, then the BR process will not continue

### Proposed Solutions
  Option 1: Clear snapshot context and last_snapshot_lsn (Complete Fix)
  - Pros: Resolves the issue completely
  - Cons: Requires significant effort
  - Alternative: Move the PG out and add it back again

  Option 2: Support a function to manually trigger truncate (Complete Fix)
  - Pros: Resolves the issue completely
  - Cons: Requires careful analysis of corner cases between manual compaction and background compaction

  Option 3: Disallow Append for Logs with LSN < snapshot_lsn (Partial Fix)
  - Pros: Simple implementation, prevents problematic append log operations
[...]

PR #841, opened 2025-12-15
Add nuraft_state field for complete Raft state persistence

Besroy:
This change use NuRaft's native binary serialization to persist srv_state instead of manual JSON field mapping. This ensures all state fields including receiving_snapshot_ are saved, preventing state inconsistency and unexpected raft behavior.

related issue: https://github.com/eBay/HomeStore/issues/835


xiaoxichen (approved):

yuwmao (approved):
LGTM

PR #844, opened 2026-01-06
Improve ReplaceMemberRollback UT

yuwmao:
- Add flip to control complete_replace_member, it is enabled after we check replace member intermediate status.
- In ReplaceMemberRollback UT, wait for in member being caught up before rollback.


JacksonYao287 (commented):
LGTM , except one minor comment.

on question, is this a temporary solution to unblock UT now? or this will be a final solution, which means we will never involve `monitor_replace_member_replication_status` in raft_repl_dev UT?

PR #846, opened 2026-01-07
Add interface to add client data rpc

JacksonYao287:
expose nuraft_message to upper layer, so that they can register their own rpc call by calling `bind_data_service_request`

### DIFF
--- a/src/include/homestore/replication/repl_dev.h
+++ b/src/include/homestore/replication/repl_dev.h
@@ -449,7 +449,7 @@ public:
     virtual void on_no_space_left(repl_lsn_t lsn, sisl::blob const& header) = 0;
 
     /// @brief when restart, after all the logs are replayed and before joining raft group, notify the upper layer
-    virtual void on_log_replay_done(const group_id_t& group_id) = 0;
+    virtual void on_log_replay_done(const group_id_t& group_id) {};
 
     virtual void on_become_leader(const group_id_t& group_id) {};
 
--- a/src/tests/test_common/raft_repl_test_base.hpp
+++ b/src/tests/test_common/raft_repl_test_base.hpp
@@ -206,8 +206,6 @@ public:
         repl_dev()->reset_latch_lsn();
     }
 
-    void on_log_replay_done(const group_id_t& group_id) override {};
-
     AsyncReplResult<> create_snapshot(shared< snapshot_context > context) override {
         std::lock_guard< std::mutex > lock(m_snapshot_lock);
         auto s = std::dynamic_pointer_cast< nuraft_snapshot_context >(context)->nuraft_snapshot();
--- a/src/tests/test_solo_repl_dev.cpp
+++ b/src/tests/test_solo_repl_dev.cpp
@@ -142,7 +142,6 @@ public:
         void notify_committed_lsn(int64_t lsn) override {}
         void on_config_rollback(int64_t lsn) override {}
         void on_no_space_left(repl_lsn_t lsn, sisl::blob const& header) override {}
-        void on_log_replay_done(const group_id_t& group_id) override {};
     };
 
     class Application : public ReplApplication {


src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -2820,6 +2820,7 @@ void RaftReplDev::become_leader_cb() {
     // becoming a leader.
 
     RD_LOGD(NO_TRACE_ID, "become_leader_cb: setting traffic_ready_lsn from {} to {}", current_gate, new_gate);
+    m_listener->on_become_leader(m_group_id);
```
Besroy: Just curious about the purpose of on_become_xxx. Should we move it to the beginning of become_xxx_cb to call the upper callback first, similar to how handle_commit does?

src/include/homestore/replication/repl_dev.h
```diff
@@ -624,6 +630,20 @@ class ReplDev {
     // create a snapshot manually and try to compact logs upto compact_lsn
     virtual void trigger_snapshot_creation(repl_lsn_t compact_lsn, bool wait_for_commit) = 0;
 
+    // add a data rpc service named request_name with request_handler
+    virtual bool add_data_rpc_service(std::string const& request_name,
+                                      data_service_request_handler_t const& request_handler) = 0;
+
+    // send a unidirectional data rpc to dest with request_name and cli_buf
+    virtual nuraft_mesg::NullAsyncResult data_request_unidirectional(nuraft_mesg::destination_t const& dest,
```
xiaoxichen: as a generic interface we probably cannot use nuraft_mesg::XXXX, 
we can use either
```
replica_member_info
```
or 
```
replica_id_t
```

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -138,6 +138,23 @@ bool RaftReplDev::join_group() {
     return true;
 }
 
+bool RaftReplDev::add_data_rpc_service(std::string const& request_name,
+                                       data_service_request_handler_t const& request_handler) {
+    return m_msg_mgr.bind_data_service_request(request_name, m_group_id, request_handler);
```
xiaoxichen: not sure, kind of concerning it will be a problem during shutdown,  the HO will be deconstructed first then the handler is invalid.  

It is not major atm.

src/include/homestore/replication/repl_dev.h
JacksonYao287: `using destination_t = std::variant< peer_id_t, role_regex, svr_id_t >`

destination_t support several type of input params,  this is more adaptive, 

maybe other upper layer(not homeobject) wants to use svr_id_t or role_regex .

JacksonYao287:
@Besroy @xiaoxichen  I add some encapsulation so that no nuraft_mesg type will be exposed to upper layer

src/lib/replication/repl_dev/raft_repl_dev.h
```diff
@@ -474,6 +483,8 @@ class RaftReplDev : public ReplDev,
     void fetch_data_from_remote(std::vector< repl_req_ptr_t > rreqs);
     void handle_fetch_data_response(sisl::GenericClientResponse response, std::vector< repl_req_ptr_t > rreqs);
     bool is_resync_mode();
+    data_rpc_error_code nuraft_to_data_rpc_error_code(nuraft::cmd_result_code const& nuraft_err);
+    nuraft_mesg::destination_t change_to_nuraft_mesg_destination(destination_t dest);
```
xiaoxichen: lets have a consistent naming for the converter.

src/include/homestore/replication/repl_dev.h
```diff
@@ -35,6 +35,19 @@ using raft_cluster_config_ptr_t = nuraft::ptr< nuraft::cluster_config >;
 using repl_req_ptr_t = boost::intrusive_ptr< repl_req_ctx >;
 using trace_id_t = u_int64_t;
 
+using data_service_request_handler_t = std::function< void(boost::intrusive_ptr< sisl::GenericRpcData >& rpc_data) >;
+ENUM(role_regex, uint8_t, LEADER, FOLLOWER, ALL, ANY);
+using peer_id_t = boost::uuids::uuid;
```
xiaoxichen: I think we can use replica_id_t directly.

src/include/homestore/replication/repl_dev.h
```diff
@@ -35,6 +35,19 @@ using raft_cluster_config_ptr_t = nuraft::ptr< nuraft::cluster_config >;
...
+using svr_id_t = int32_t;
+using destination_t = std::variant< peer_id_t, role_regex, svr_id_t >;
```
xiaoxichen: using repl_dest_t = std::variant< replica_id_t, repl_dest_role, svr_id_t>;

src/include/homestore/replication/repl_dev.h
```diff
@@ -35,6 +35,19 @@ using raft_cluster_config_ptr_t = nuraft::ptr< nuraft::cluster_config >;
...
+
+ENUM(data_rpc_error_code, uint8_t, SUCCESS, TIMEOUT, SERVER_NOT_FOUND, CANCELLED, SERVER_ALREADY_EXISTS, TERM_MISMATCH,
```
xiaoxichen: repl_deta_rpc_error_code

src/include/homestore/replication/repl_dev.h
```diff
@@ -436,7 +449,11 @@ class ReplDevListener {
     virtual void on_no_space_left(repl_lsn_t lsn, sisl::blob const& header) = 0;
 
     /// @brief when restart, after all the logs are replayed and before joining raft group, notify the upper layer
-    virtual void on_log_replay_done(const group_id_t& group_id) {};
+    virtual void on_log_replay_done(const group_id_t& group_id) = 0;
```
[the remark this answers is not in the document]
JacksonYao287: this was designed to be pure virtual originally。when we added this interface, in order to not block the homeobject build, we make it not pure virtual and have a default implementation.  now that we have already implement this virtual function in homeobject side, we can change it to pure virtual now.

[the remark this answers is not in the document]
xiaoxichen: I am not sure, at least in the SoloReplDevUT there is a ReplDevListener, so I am expecting same pattern, the ReplDevListener is not solely for RaftReplDev.

Personally I wont touch it , but you can ask in the channel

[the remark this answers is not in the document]
JacksonYao287: I remove the changes on this part, let`s keep it for now

PR #851, opened 2026-01-16
save HS knowledge 

Besroy:
This PR aims to preserve knowledge generated during regular debugging and work, making it easier for AI to understand our architecture without having to read the code every time it needs to solve a problem. Let's save them here for supplementation and correction. No need to merge it now, as it is still in a very early and rough stage


JacksonYao287:
this is also good for the later new comers to understand how homestore works ASAP.

can you also put what prompts are you using to generate this detailed doc? I think this is also important. 

IMO, in the future we (human being) need to improve the capability of generating precious prompts and the way how we ask AI will matter more and more.

PR #861, opened 2026-02-09
Make chunk_size to uint64_t to avoid overflow and support chunk size>=4GB

shosseinimotlagh:
Currently in HomeBlks, overflow can happen if chunk size is set for more than 4GB.  To avoid it, change chunk size type from uint32_t to uint64_t


JacksonYao287:
if compatibility can be well handled and can do inplace upgrade, I am good with this change.

PR #866, opened 2026-02-24
Fix superblock IO error handling and add footer validation

xiaoxichen:
This change improves superblock error handling and integrity checking:

- Fail fast on IO errors during superblock read instead of treating them as fresh/unformatted disks
- Add footer superblock validation for HDD devices to detect corruption
- Check header and footer write errors independently to prevent silent failures
- Add comprehensive unit tests covering all error scenarios

Changes:
- read_first_block(): Now throws std::system_error on IO errors instead of returning garbage data
- write_super_block(): Separately validates header and footer writes with independent error checking
- sanity_check(): New method that validates footer consistency on HDD devices by comparing header and footer superblocks using full memcmp
- Added 6 unit tests covering IO errors, corruption detection, and footer validation scenarios


JacksonYao287 (commented):
I am curious where the pdev superblk is found corrupted. production?

xiaoxichen:
> I am curious where the pdev superblk is found corrupted. production?

We hit one issue in QLC that the read superblk returns EIO,   the original code fails silently and pretend it is a new drive, sending registration to CM.

PR #867, opened 2026-03-03
complete the ctx future if no dirty to unblock cp flush

JacksonYao287:
Bug:
when there is no dirty buf in index table (example : all read ops or no ops), if we trigger cp flush, then future in cp_ctx of index will not be completed and then the whole cp_flush is stuck.

Fix: 
complete the future in cp_ctx if dirty buf is empty


src/lib/index/wb_cache.cpp
shosseinimotlagh: when there is no dirty buffers  flushing stopped here. the flushing doesn't reach to your change. Please look at my change in https://github.com/eBay/HomeStore/pull/861

shosseinimotlagh (commented):
I am not convinced that this bug can lead to cp stuck. It will be great if you may write a simple UT to verify the bug. 
1- write in index
2- restart 
3- read only 
(CP should stuck )

JacksonYao287:
@shosseinimotlagh  I will try to add a UT for this

PR #871, opened 2026-03-16
Fix root collapse CP hang and add regression test

yuwmao:
This commit addresses two critical CP hang scenarios during root collapse  and updates the test to serve as a regression test.
 Root Cause Fix (btree_remove_impl.ipp):
     Add write_node(child) before on_root_changed() in check_collapse_root() to prevent two CP hang scenarios:
     a) CLEAN Buffer Case:
        If promoted child is CLEAN (from previous CP), on_root_changed() would add it as dependency to Meta buffer. Since CLEAN buffers never enter CP dirty list, Meta's wait_count never decrements, causing CP hang.
     b) Overlapping CP Case:
        Without write_node(), there's a gap between dependency creation (link_buf) and dirty list addition (write_buf). If CP switches during this gap, Meta goes to CP X with wait_count=1, but child goes to CP X+1, creating cross-CP dependency and CP hang.
     By calling write_node() before on_root_changed(), we ensure child is already in current CP's dirty list before dependency link is created.
Additional changes:
  - Add cp_ctx->complete() call for early return path in async_cp_flush()

### DIFF
--- a/src/lib/index/wb_cache.cpp
+++ b/src/lib/index/wb_cache.cpp
@@ -430,6 +430,16 @@ void IndexWBCache::link_buf(IndexBufferPtr const& up_buf, IndexBufferPtr const& 
         down_buf->m_up_buffer->remove_down_buffer(down_buf);
     }
     down_buf->m_up_buffer = real_up_buf;
+
+    // Log when adding CLEAN buffer as dependency, should not happen
+    if (down_buf->state() == index_buf_state_t::CLEAN) {
+        LOGWARNMOD(wbcache,
+                   "CLEAN_BUF_DEBUG: Adding CLEAN down_buf {} to up_buf {}, up_buf wait_count before={}, may lead to "
+                   "up_buf being stuck in cp flush",
+                   down_buf->blkid().to_integer(), real_up_buf->blkid().to_integer(),
+                   real_up_buf->m_wait_for_down_buffers.get());
+    }
+
     real_up_buf->add_down_buffer(down_buf);
 }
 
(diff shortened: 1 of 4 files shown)


xiaoxichen (dismissed):
lgtm

src/lib/index/wb_cache.cpp
```diff
@@ -818,7 +828,11 @@ bool IndexWBCache::was_node_committed(IndexBufferPtr const& buf) {
 
 //////////////////// CP Related API section /////////////////////////////////
 folly::Future< bool > IndexWBCache::async_cp_flush(IndexCPContext* cp_ctx) {
+#ifdef _PRERELEASE
     LOGTRACEMOD(wbcache, "Starting Index CP Flush with cp \ndag={}", cp_ctx->to_string_with_dags());
+#else
+    LOGINFOMOD(wbcache, "Starting Index CP Flush with cp {}", cp_ctx->id());
```
shosseinimotlagh: nit: can you add more info here (number of dirty buffer list, number of free buffers, etc)

src/tests/test_index_btree.cpp
```diff
@@ -455,6 +455,111 @@ TYPED_TEST(BtreeTest, ThreadedCpFlush) {
...
+ *
+ * Test Objective:
+ * Ensure that when root collapse occurs and the promoted child buffer is added to the same CP with the meta buf,
+ * the CP flush completes successfully without hanging.
+ *
+ * Test Scenario:
+ * This test constructs a two-CP scenario:
+ *
+ * CP2 (Setup):
+ * - Insert 600 entries → creates 3-child tree (depth=1, interior=1, leaf=3)
```
shosseinimotlagh, line 469: nit you can use for prerelease
```
m_max_keys_in_node
m_min_keys_in_node
```

xiaoxichen: this is amazing for future UT, thanks!

src/lib/index/wb_cache.cpp
yuwmao: Sure

src/tests/test_index_btree.cpp
yuwmao: Thanks, it would be quite useful.

xiaoxichen (approved):

PR #872, opened 2026-03-18
Fix race condition between cp flush and chunk reset

Besroy:
Upper layer (GC) calls reset before chunk usage, which resets the old blk allocator. If the chunk is being cp flushed, resetting the allocator destroys the space, causing crashes when accessing invalid memory.

### DIFF
--- a/src/lib/blkalloc/append_blk_allocator.cpp
+++ b/src/lib/blkalloc/append_blk_allocator.cpp
@@ -16,6 +16,7 @@
 #include <homestore/checkpoint/cp_mgr.hpp>
 #include <homestore/checkpoint/cp.hpp>
 #include <homestore/meta_service.hpp>
+#include <iomgr/iomgr_flip.hpp>
 
 #include "append_blk_allocator.h"
 
@@ -125,6 +126,14 @@ BlkAllocStatus AppendBlkAllocator::reserve_on_cache(BlkId const& blkid) {
 
 void AppendBlkAllocator::cp_flush(CP* cp) {
     // check if current cp's context has dirty buffer already
+    std::lock_guard lg(m_sb_mtx);
+
+#ifdef _PRERELEASE
+    // Flip point: inside cp_flush with lock held
+    if (iomgr_flip::instance()->callback_flip("inside_append_cp_flush")) {
+        LOGINFO("Flip triggered: inside_append_cp_flush");
+    }
+#endif
     if (m_is_dirty.exchange(false)) {
         m_sb->commit_offset = m_commit_offset.load();
         m_sb->freeable_nblks = m_freeable_nblks.load();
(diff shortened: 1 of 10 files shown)


src/lib/blkalloc/append_blk_allocator.cpp
```diff
@@ -146,6 +147,7 @@ bool AppendBlkAllocator::is_blk_alloced(const BlkId& in_bid, bool) const {
 }
 
 void AppendBlkAllocator::reset() {
+    std::lock_guard lg(m_sb_mtx);
     m_is_dirty.store(false);
     m_sb.destroy();
```
JacksonYao287, line 160: suggest add the protection of m_sb_mtx  also into 

```
bool AppendBlkAllocator::is_blk_alloced_on_disk(BlkId const& bid, bool) const {
    return bid.blk_num() < m_sb->commit_offset;
}
```

where  m_sb is referenced

src/lib/device/virtual_dev.cpp
```diff
@@ -743,7 +743,9 @@ void VirtualDev::cp_flush(VDevCPContext* v_cp_ctx) {
     // pass down cp so that underlying components can get their customized CP context if needed;
     m_chunk_selector->foreach_chunks([this, cp](cshared< Chunk >& chunk) {
         HS_LOG(TRACE, device, "Flushing chunk: {}, vdev: {}", chunk->chunk_id(), m_vdev_info.name);
-        chunk->blk_allocator_mutable()->cp_flush(cp);
+        // Hold shared_ptr to prevent allocator from being reset during cp_flush.
+        // This fixes race with GC's purge_reserved_chunk() which calls reset_block_allocator().
+        chunk->blk_allocator_shared()->cp_flush(cp);
```
JacksonYao287: do we need a lock here? for example:

thread1 (here) : A = m_blk_allocator ( chunk->blk_allocator_shared())
thread2 (reset_block_allocator):  m_blk_allocator = new_blkallocator。

if thread 1 and thread 2 run concurrently,  finally, A is not determined, the worst case, maybe A is nullptr.

src/lib/blkalloc/append_blk_allocator.cpp
Besroy: Good suggestion, let me add it

src/lib/device/virtual_dev.cpp
Besroy: Could you explain more on it? I cannot get why A is nullptr instead of old blk allocator

JacksonYao287:
```
std::shared_ptr<int> A = std::make_shared<int>(42);
std::shared_ptr<int> C = std::make_shared<int>(100);
std::shared_ptr<int> B;

// thread 1:   B = A;
// thread 2:  A = C;
```

according to the above code, what is the final value of B? since shared_ptr is not thread-safe, there will be some unexpected behavior.

JacksonYao287 (dismissed):
LGTM

src/lib/device/virtual_dev.cpp
Besroy: Thanks! This is a nice catch.

src/lib/device/chunk.h
```diff
@@ -67,12 +67,20 @@ class Chunk {
     nlohmann::json get_status([[maybe_unused]] int log_level) const;
     const BlkAllocator* blk_allocator() const { return m_blk_allocator.get(); }
     BlkAllocator* blk_allocator_mutable() { return m_blk_allocator.get(); }
+    std::shared_ptr< BlkAllocator > blk_allocator_shared() {
+        std::shared_lock lg{m_mgmt_mutex};
+        return m_blk_allocator;
+    }
+    std::shared_ptr< BlkAllocator > blk_allocator_shared_unlocked() { return m_blk_allocator; }
```
JacksonYao287: when this function is used?  if it is never used, let`s remove this unsafe call

JacksonYao287 (dismissed):

src/lib/device/chunk.h
Besroy: Currently no used, just copy like HO get_hs_pg_unlocked() for potential future usage. However, its fine to remove it for now. If we really need an unsafe function in the future, we can add it at that time

JacksonYao287 (approved):

PR #874, opened 2026-03-27
 SDSTOR-21909: Reduce redundant logs for flush_durable_commit_lsn and gc_repl_reqs

Besroy:


JacksonYao287:
if we have different timer with different intervals， we have to calculate the frequency for each timer and add the into the config file. if the interval change, in order to keep the same log frequency, we need to recalculate and change the config file. this means:

1 for each timer, we need to add a single item in the config file. if we have a lot of timer, we will have a lot of config item in the config file.

2 if we change the interval of a timer, to keep the log frequency, we have to recalculate the frequency and change it in the config file.

I am thinking can we have some kind of log policies (for example, log_per_hour, log_per_minute or log_per_second), we only need to set the log policy for each timer, and the timer will calculate the frequency itself, so that it will be more adaptive。


the code here looks good to me, just bring this thought here for discussion

xiaoxichen:
log_per_interval is nice but expensive to maintain.  If such capability can be bring in it would be nice.

xiaoxichen (approved):

Besroy:
Yes log_per_interval can also be helpful for cleanup (to avoid log increase -> mem increase -> oom), let me create a github issue for it.

PR #875, opened 2026-04-02
SDSTOR-21438: Adjust commit_quorum reduction ordering

yuwmao:
When two members are down, the raft leader yields leadership after leadership_expiry (20x heartbeat). reset_quorum_size must be called before the NOT_LEADER check so the node can maintain/reclaim leadership with election_quorum=1. 

Previously, the TwoMemberDown UT called replace_member immediately, so the leadership doesn't expired. Sleep 10s to simulate leader expiry.

One thing to mention is that if a request fails after NOT_LEADER check, need retry twice, because it will reset commit_quorum to 0 before return.

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -219,12 +219,40 @@ AsyncReplResult<> RaftReplDev::start_replace_member(std::string& task_id, const 
             boost::uuids::to_string(member_out.id), boost::uuids::to_string(member_in.id));
 
     // Step1, validate request
+    bool is_leader = m_my_repl_id == get_leader_id();
+    // Check if leader itself is requested to move out.
+    if (m_my_repl_id == member_out.id) {
+        // immediate=false successor=-1, nuraft will choose an alive peer with highest priority as successor, and wait
+        // until the successor finishes the catch-up of the latest log, and then resign. Return NOT_LEADER and let
+        // client retry.
+        if (is_leader) {
+            RD_LOGI(trace_id, "Step1. Replace member, leader is the member_out so yield leadership, task_id={}", task_id);
+            raft_server()->yield_leadership(false /* immediate */, -1 /* successor */);
+        }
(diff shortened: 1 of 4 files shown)


yuwmao:
> When two members are down, the raft leader yields leadership after leadership_expiry (20x heartbeat). reset_quorum_size must be called before the NOT_LEADER check so the node can maintain/reclaim leadership with election_quorum=1. Add comment explaining this ordering constraint.
> 
> Previously, the TwoMemberDown UT called replace_member immediately, so the leadership doesn't expired. Sleep 10s to simulate leader expiry.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -241,6 +241,18 @@ AsyncReplResult<> RaftReplDev::start_replace_member(std::string& task_id, const
...
+        // reduced election quorum. By calling reset_quorum_size here first (which sets both custom_commit_quorum
+        // and custom_election_quorum to 1), the current leader is able to maintain leadership, and if leadership
+        // was already lost, the node will self-elect on the next election timeout. The caller should retry on
+        // NOT_LEADER to allow time for self-election to complete.
+        reset_quorum_size(commit_quorum, trace_id);
+    }
+
     if (m_my_repl_id != get_leader_id()) { return make_async_error<>(ReplServiceError::NOT_LEADER); }
     // Check if leader itself is requested to move out.
     if (m_my_repl_id == member_out.id) {
```
xiaoxichen: Is that better to move `if (m_my_repl_id == member_out.id) ` check at the very beginning?
```
bool is_leader = m_my_repl_id == get_leader_id());
// I am not the one to be removed
if (m_my_repl_id == member_out.id)  {
      if (is_leader). yield_leadership();
     return make_async_error<>(ReplServiceError::NOT_LEADER);
}
// not leader handling
if (!is_leader) {
    if (commit_quorum >=1) {
        // reduce quorum size ....
        reset_quorum_size(commit_quorum, trace_id);
    }
    return make_async_error<>(ReplServiceError::NOT_LEADER); 
}
// Now I am the leader

```

xiaoxichen (approved):

PR #879, opened 2026-04-20
SDSTOR-21880: MetaIndexBuffer blkid collision during recovery

yuwmao:
Previously all MetaIndexBuffer instances, regardless of which BTree index they belong to, use BlkId{} (all zeros) as their blkid.  buf_map is keyed by BlkId, so all MetaIndexBuffer entries collide. As a result, Sanity check fails due to mismatched index_ordinals.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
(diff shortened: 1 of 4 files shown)


Besroy (dismissed):
LGTM

xiaoxichen (dismissed):
lgtm

xiaoxichen (approved):

PR #886, opened 2026-05-15
SDSTOR-21863: Revert AppendBlkAllocator Reset logic

Besroy:
to prevent race condition between upper layer reset and self operation on it (cp_flush and free)

Fix https://github.com/eBay/HomeObject/issues/401


JacksonYao287 (approved):
LGTM

PR #887, opened 2026-05-27
Fix #581: Add CRC32 checksum verification for pushdata/fetchdata data channels

### LINKED ISSUE #581: Data Mismatch in read verify
From SM-long-running-test side, we encountered failures during the data read_verify process on two separate occasions:
**Hash Mismatch**: The issue occurred once where one follower experienced a hash mismatch.
**Invalid Header**: Another instance where a different follower encountered an invalid header.
**Additional Details**: For more comprehensive information and context regarding these failures, please refer to the documentation provided in the following link: [Google Docs](https://docs.google.com/document/d/16DMqv9J-JuNs5c25IBuX01QXe5WbxPm4B4hAfz4JNZM/edit?tab=t.0#heading=h.1l9cfab6npvs)

OmDoshi13:
## Summary

Fixes #581 — silent data corruption caused by network-layer corruption between leader and followers going undetected until user read/write time.

- **Root cause:** Data left the leader intact but could be corrupted in the network layer. Followers accepted and stored whatever arrived; users only discovered corruption at read/write time.
- **Fix:** Attach a CRC32 checksum to every data packet sent over the PushData and FetchData channels. The receiver recomputes the checksum and drops the packet on mismatch, letting Raft retry rather than writing corrupted bytes to disk.

### Key implementation details

- CRC32 computed over the full data payload at send time; verified before any write on the follower
- FetchData response uses a self-describing magic-byte framing header (`FETCH_DATA_RESPONSE_MAGIC`) so receivers detect the header without consulting per-node config — safe under hotswap config asymmetry during rolling upgrades
- `data_checksum_enabled` defaults to `false`; operators enable via hotswap **after** all nodes are running the new binary to avoid old nodes misinterpreting the framing header as block data
- Added `data_checksum_mismatch_cnt` metric for observability
- FlatBuffer verifier added on the receive path to guard against malformed headers
- Per-entry `lsn`/`dsn`/`raft_term` attached to `ResponseEntry` for richer diagnostics

### Correctness fixes (commit 3)

Three issues identified during review and fixed before merge:

- **F1 — Immediate re-fetch on FetchData checksum mismatch:** Previously a mismatch silently dropped the entry and waited for the full `data_receive_timeout_ms` (10s) before Raft retried. Now mismatched rreqs are collected during the response loop and passed to `check_and_fetch_remote_data()` immediately after, eliminating the stall.
- **F2 — Unsigned underflow guard in `handle_fetch_data_response`:** A truncated or malformed response could cause `total_size -= data_size` to wrap to ~4GB and advance `raw_data` past the buffer boundary in release builds. Added an explicit `data_size > total_size` guard with early return before each subtraction.
- **F3 — FlatBuffer verifier on PushData receive path:** `push_req` fields were accessed before the buffer was verified in `on_push_data_received`. A corrupted or crafted push RPC could cause out-of-bounds reads via the unverified FlatBuffer accessor. Added `flatbuffers::Verifier` call before any field access, matching the pattern already present on the FetchData response path.

## Test plan

- [x] `Checksum_Enabled_PushData_Path` — happy path with checksums on; writes commit correctly across all replicas
- [x] `Checksum_Enabled_FetchData_Path` — drops all push-data on followers forcing fetch path; verifies framing header is parsed correctly and data arrives intact
- [x] Existing `Follower_Fetch_OnActive_ReplicaGroup` still passes (no regression)

All 3 tests passed across all 4 CI build variants (Debug/Sanitize, Debug/Coverage, Release/tcmalloc prerelease, Release/tcmalloc) — 100% tests passed, 0 failed out of 21 total.

### DIFF
--- a/src/lib/replication/repl_dev/raft_repl_dev.cpp
+++ b/src/lib/replication/repl_dev/raft_repl_dev.cpp
@@ -1194,6 +1194,13 @@ void RaftReplDev::on_push_data_received(intrusive< sisl::GenericRpcData >& rpc_d
         rpc_data->send_response();
         return;
     }
+    // user_header and user_key are optional FlatBuffer fields; guard against null even though
+    // a well-formed sender always sets them — the Verifier above only checks structural integrity.
+    if (!push_req->user_header() || !push_req->user_key()) {
+        RD_LOGW(NO_TRACE_ID, "Data Channel: PushData missing user_header or user_key, ignoring");
+        rpc_data->send_response();
+        return;
+    }
     sisl::blob header = sisl::blob{push_req->user_header()->Data(), push_req->user_header()->size()};
     sisl::blob key = sisl::blob{push_req->user_key()->Data(), push_req->user_key()->size()};
     repl_key rkey{.server_id = push_req->issuer_replica_id(),
@@ -1206,7 +1213,10 @@ void RaftReplDev::on_push_data_received(intrusive< sisl::GenericRpcData >& rpc_d
 
     if (push_req->checksum() != 0) {
         auto const data_ptr = r_cast< const unsigned char* >(incoming_buf.cbytes() + fb_size);
-        auto const computed = crc32_ieee(init_crc32, data_ptr, push_req->data_size());
+        auto computed = crc32_ieee(init_crc32, data_ptr, push_req->data_size());
+#ifdef _PRERELEASE
+        if (iomgr_flip::instance()->test_flip("corrupt_push_data_checksum")) { computed ^= 0xdeadbeef; }
+#endif
         if (computed != push_req->checksum()) {
             COUNTER_INCREMENT(m_metrics, push_data_checksum_mismatch_cnt, 1);
             RD_LOGE(rkey.traceID,
@@ -1668,7 +1678,7 @@ void RaftReplDev::on_fetch_data_received(intrusive< sisl::GenericRpcData >& rpc_
             // data_checksum_enabled is true.  When disabled, send raw block data (pre-checksum wire
             // format) so old receivers are never surprised by an unexpected header during rolling
             // upgrades.  Receivers use try-and-fallback to detect the FlatBuffer transparently.
-            uint8_t* hdr_buf = nullptr;
+            std::unique_ptr< uint8_t[] > hdr_buf;
             uint32_t hdr_size = 0;
 
             if (HS_DYNAMIC_CONFIG(consensus.data_checksum_enabled)) {
@@ -1693,21 +1703,23 @@ void RaftReplDev::on_fetch_data_received(intrusive< sisl::GenericRpcData >& rpc_
 
                 // Heap-copy the FlatBuffer so it outlives resp_builder until the send completion callback.
                 hdr_size = resp_builder.GetSize();
-                hdr_buf = new uint8_t[hdr_size];
-                std::memcpy(hdr_buf, resp_builder.GetBufferPointer(), hdr_size);
+                hdr_buf = std::make_unique< uint8_t[] >(hdr_size);
+                std::memcpy(hdr_buf.get(), resp_builder.GetBufferPointer(), hdr_size);
             }
 
             // now prepare the io_blob_list to response back to requester;
             nuraft_mesg::io_blob_list_t pkts = sisl::io_blob_list_t{};
-            if (hdr_buf) { pkts.emplace_back(sisl::io_blob{hdr_buf, hdr_size, false}); }
+            if (hdr_buf) { pkts.emplace_back(sisl::io_blob{hdr_buf.get(), hdr_size, false}); }
             for (auto const& sgs : sgs_vec) {
                 auto const ret = sisl::io_blob::sg_list_to_ioblob_list(sgs);
                 pkts.insert(pkts.end(), ret.begin(), ret.end());
             }
+            // All potential throws are past; release ownership of hdr_buf to the completion lambda.
+            auto* raw_hdr = hdr_buf.release();
 
             rpc_data->set_comp_cb(
-                [sgs_vec = std::move(sgs_vec), hdr_buf](boost::intrusive_ptr< sisl::GenericRpcData >&) {
-                    delete[] hdr_buf; // delete[] nullptr is a no-op when checksums are disabled
+                [sgs_vec = std::move(sgs_vec), raw_hdr](boost::intrusive_ptr< sisl::GenericRpcData >&) {
+                    delete[] raw_hdr; // delete[] nullptr is a no-op when checksums are disabled
                     for (auto const& sgs : sgs_vec) {
                         for (auto const& iov : sgs.iovs) {
                             iomanager.iobuf_free(reinterpret_cast< uint8_t* >(iov.iov_base));
@@ -1758,6 +1770,10 @@ void RaftReplDev::handle_fetch_data_response(sisl::GenericClientResponse respons
                                 resp_entries->size(), rreqs.size());
                     }
                 }
+            } else {
+                RD_LOGD(NO_TRACE_ID,
+                        "Data Channel: FetchData response FlatBuffer verifier failed despite matching identifier, "
+                        "treating as legacy raw-block format");
             }
         }
     }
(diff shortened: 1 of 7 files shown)


xiaoxichen (commented):
@OmDoshi13 thanks for taking this and helping with the tech debt,  It looks nice in general.

I think your changes can be categorised into 3 parts
A:  Corner cases fixes, mainly in your commit 3.
B:  Moving `FetchDataResponse` from raw to Flatbuffer 
C: CRC implementation.

I dont have any problem with A and hoping A can be done in a separate PR  to allow it be merged sooner.

For B,   I saw you added a `FETCH_DATA_RESPONSE_MAGIC` as well as  corresponding logic to support rolling upgrade and hybrid version,  that is nice.  And yes,   It is totally a mistake in the past that skipping flatbuffer instead using raw format...

 however,  using the magic header is another dead-end, as we have to carry this "compatibility" tax forever, it doesnt help us in next release that removing the magic header and merge back to the native flat-buffer approach. 

With this thinking,    do you think we can directly using FlatBuffer w/o MAGIC, as flatbuffer is size-prefixed,  we can read the first size_bytes, and  check if that number makes sense , then proceed to `flatbuffer::verify`,  if that is also passing it is very likely it is new format, then we check CRCs of raw data against what was in flatbuffer. 


Basically I am trying to suggest try-and-fallback pattern.Note the raw data should be a homeobject blobs/shards where has a DataHeader with magic `uint64_t magic{0x21fdffdba8d68fc6};,` , that makes the flatbuffer prefixed size to be 2.6GB, which will very probably fails the basic size check.


```
    struct DataHeader {
        static constexpr uint8_t data_header_version = 0x02;
        static constexpr uint64_t data_header_magic = 0x21fdffdba8d68fc6; // echo "BlobHeader" | md5sum

        enum class data_type_t : uint32_t { SHARD_INFO = 1, BLOB_INFO = 2 };

        bool valid() const { return ((magic == data_header_magic) && (version <= data_header_version)); }

        uint64_t magic{data_header_magic};
        uint8_t version{data_header_version};
        data_type_t type{data_type_t::BLOB_INFO};
    };
```

xiaoxichen:
For C:
Lets try to separate the enablement of CRC with the wire-format changes.  If in B we moved to Flatbuffer, then checksum would have a default value of zero if not enabled, we probably want to skip the CRC checking if CRC field is zero.

A bit more context on the CRC:  once we added length check, we never saw data corruption on wire anymore.  Basically it is hard to see bit rot in TCP protocol as TCP protocol  already provides its CRC in TCP-Header.  Usually the error was caused by early terminating --- sender disconnected  before sending full data,  receiver received partial data and blindly using it without checking the prefix-size.  

With this assumption and observation in production ,  I think we wont enable the CRC for generic use cases, only use in more durability sensitive environment if needed.

OmDoshi13:
Hey @xiaoxichen, the branch has been updated — `147fcc8a` is the latest commit and addresses your review feedback on parts B and C. Please ignore the earlier push events in the timeline.

xiaoxichen (dismissed):
lgtm.
cc  @Besroy note this "might" be a breaking change if the detect-and-fallback on receiver side doesn't works as expected.

src/lib/replication/repl_dev/raft_repl_dev.cpp
```diff
@@ -1612,13 +1690,71 @@ void RaftReplDev::handle_fetch_data_response(sisl::GenericClientResponse respons
         return;
     }
 
+    // Try-and-fallback: attempt to parse a size-prefixed FetchDataResponse FlatBuffer at the start
+    // of the blob.  New senders (data_checksum_enabled=true) prepend this header; old senders emit
+    // raw block data with no prefix.  In HomeObject (the layer above HomeStore), raw block data
```
JacksonYao287: we should not make any assumption of the layer above homestore. raft_repl_dev is not only used by homeobject , it`s probably used by other storage applicaion in the future.

if another application built on homestore has a very small first 4 bytes(let`s say 0x10, for example),  the following logic will confuse the old version with the new one

src/lib/replication/repl_dev/raft_repl_dev.h
```diff
@@ -71,6 +71,8 @@ class RaftReplDevMetrics : public sisl::MetricsGroup {
         REGISTER_COUNTER(read_err_cnt, "total read error count", "read_err_cnt", {"op", "read"});
         REGISTER_COUNTER(write_err_cnt, "total write error count", "write_err_cnt", {"op", "write"});
         REGISTER_COUNTER(fetch_err_cnt, "total fetch data error count", "fetch_err_cnt", {"op", "fetch"});
+        REGISTER_COUNTER(data_checksum_mismatch_cnt, "CRC32 mismatches on push/fetch data channels",
```
JacksonYao287: can we distinguish push_data_checksum_mismatch_cnt and fetch_data_checksum_mismatch_cnt

src/lib/replication/repl_dev/raft_repl_dev.cpp
xiaoxichen: who is "another" application as of now?  If there is no one, then it is safe.  Tomorrow after the PR merged if there is new application on top of homestore they already on new code.

src/lib/replication/repl_dev/raft_repl_dev.h
xiaoxichen: reason?

OmDoshi13:
@xiaoxichen — A, B, and C are all addressed in the current branch:

**A (corner cases):** The three fixes (F1 immediate re-fetch on mismatch, F2 underflow guard, F3 PushData Verifier null-guard) are all included.

**B (FlatBuffer with try-and-fallback, no magic):** The `FETCH_DATA_RESPONSE_MAGIC` approach has been replaced with the pattern you suggested. The receiver reads bytes 4–7 for the `"FDRS"` file identifier (set via `FinishSizePrefixedBuffer`), checks the parsed size against `data_fetch_max_size_kb`, then runs `flatbuffers::Verifier`. All three must pass before treating the response as the new format; otherwise it falls through to legacy raw-block processing. Old senders' DataHeader magic (`0x21fdffdba8d68fc6`) produces a `uoffset_t` of ~2.64 GB which fails the size check independently.

**C (CRC with checksum=0 = disabled):** The receiver skips CRC verification whenever the checksum field is zero (the FlatBuffer default). The config flag only controls the send side. The comment on `data_checksum_enabled` has been expanded to explicitly call out the UNSAFE direction (new sender with flag enabled → old receiver = silent corruption) and adds a DEPLOYMENT RULE: do not enable via hotswap until every node in the cluster has the try-and-fallback receiver.

src/lib/replication/repl_dev/raft_repl_dev.cpp
OmDoshi13: Thanks @JacksonYao287 — and that's a valid concern in principle, but the detection doesn't rely on the DataHeader magic at all. The code uses three independent gates before treating a response as the new format:

1. `BufferHasIdentifier(raw_data + sizeof(uoffset_t), "FDRS")` — bytes 4–7 must literally be the ASCII string `FDRS`. This is emitted only by `FinishSizePrefixedBuffer(..., "FDRS")` in our `on_fetch_data_received`, so it cannot accidentally appear in a future application's raw block data.
2. Parsed size must be ≤ `data_fetch_max_size_kb` (2 MB).
3. `flatbuffers::Verifier` — validates internal FlatBuffer structure.

Any future application's raw block data would need to accidentally have `0x46 0x44 0x52 0x53` at byte offset 4, then pass the size check, then satisfy the Verifier — all three independently. The DataHeader magic example @xiaoxichen gave was illustrating why the size check works for the *current* application, but the "FDRS" identifier gate is what makes this safe for any future application.

Also fully agree with @xiaoxichen: any new application built on top of HomeStore after this PR merges will already have the new receiver code, so there's no backward compatibility concern there either.

src/lib/replication/repl_dev/raft_repl_dev.h
OmDoshi13: Done — both `push_data_checksum_mismatch_cnt` and `fetch_data_checksum_mismatch_cnt` are now separate counters.

To answer @xiaoxichen's "reason?": push and fetch are distinct data paths with different failure modes. A mismatch on push (leader → follower direct RPC) narrows the problem to in-flight corruption or a sender-side CRC bug. A mismatch on fetch (follower pulling from leader) points to the same class of issue on a different code path. Keeping them separate lets operators immediately identify which channel is affected without having to correlate log timestamps or RPC traces.

src/lib/replication/repl_dev/raft_repl_dev.cpp
JacksonYao287: >the detection doesn't rely on the DataHeader magic at all
yes, this is what I want. it makes sense to me to involve independent gate  handling this case.

my thought is that , as you said, we should have independent gate and not rely on the header magic(defined by the application).

@OmDoshi13  thanks for the change

PR #892, opened 2026-06-08
SDSTOR-21939: index/wb_cache: Fix recovery corruption when a node is created and freed in the same crashed CP

JacksonYao287:
During index CP recovery, if a btree node was both created (e.g. via a split) and freed (e.g. via a subsequent merge) within the same CP that crashed before any data or bitmap was flushed, the recovery path would incorrectly call free_blk on the node's blkid. If the repair phase had already reused that blkid for a new live interior node, the free_blk would silently corrupt it.

The first recovery pass processes all journal-recorded nodes and decides which blkids to release via deleted_bufs. For freed nodes the code first checked was_node_committed, then inside the "not committed" branch checked m_created_cp_id:

    if (buf->m_node_freed) {
        if (was_node_committed(buf)) {
            deleted_bufs.push_back(buf);        // correct for prev-CP nodes
        } else {
            if (buf->m_created_cp_id != icp_ctx->id()) {
                m_vdev->commit_blk(buf->m_blkid);  // correct
            } else {
                deleted_bufs.push_back(buf);        // BUG
            }
            prune_from_up_buffer(buf);
        }
    }

A node that was created in the crashed CP (m_created_cp_id == icp_ctx->id()) and later freed in the same CP (m_node_freed == true) always landed in deleted_bufs, triggering free_blk during recovery cleanup.

The same bug also existed in the was_node_committed == true branch: a same-CP created+freed node whose parent had already been flushed to disk would also be added to deleted_bufs via line 624.

A node created within the crashed CP only had its blkid reserved via reserve_on_disk; the blkid was never committed to the allocator bitmap. After restart the allocator reloads the bitmap from the last fully committed CP and already treats that blkid as free. During the repair phase (repair_links -> alloc_interior_node), the allocator may reassign that exact blkid to a new live interior node. Recovery then calls free_blk(blkid) from deleted_bufs, releasing a node that is still in active use. Any subsequent allocation of that blkid overwrites the repair node's data, silently corrupting the index.

    Committed CP:  Parent P(100) -> Leaf L(200) with keys [10,20,30,40]

    Crashed CP (journal flushed, no node data or bitmap flush):
      Step 1: insert 25 -> L splits, new node N(300) created
              N: m_created_cp_id=current_cp, blkid 300 reserve_on_disk only
      Step 2: delete 25,30,40 -> N empties, merge frees N
              N: m_node_freed=true
      Crash

    Recovery:
      - N: m_node_freed=true AND m_created_cp_id==icp_ctx->id()
      - Old code puts N in deleted_bufs - repair_links calls alloc_interior_node() -> allocator picks blkid 300 (free in the committed bitmap) - Cleanup: free_blk(300) from deleted_bufs -> live repair node released - Next allocation at blkid 300 overwrites repair node -> index corrupted

Check m_created_cp_id == icp_ctx->id() before was_node_committed so that same-CP created+freed nodes are handled as a distinct case and always bypass deleted_bufs:

    if (buf->m_node_freed) {
        if (buf->m_created_cp_id == icp_ctx->id()) {
            // blkid never reached the bitmap; repair may already hold it
            if (was_node_committed(buf))
                pending_bufs.push_back(buf->m_up_buffer);
            else
                prune_from_up_buffer(buf);
            continue;                   // never touches deleted_bufs
        }
        if (was_node_committed(buf)) {
            // prev-CP node flushed to disk: schedule blkid release
            deleted_bufs.push_back(buf);
            pending_bufs.push_back(buf->m_up_buffer);
        } else {
            // prev-CP node freed but bitmap not flushed: recommit blkid
            m_vdev->commit_blk(buf->m_blkid);
            prune_from_up_buffer(buf);
        }
    }

For same-CP nodes whose parent was already flushed (was_node_committed true), the parent still needs repair to remove the stale child link, so its up_buffer is pushed to pending_bufs. When the parent was not flushed, the in-memory child link is simply pruned. In both cases the blkid is never freed.

Also extracted the repeated three-step prune pattern (remove_down_buffer + prune_up_buffers + clear m_up_buffer) into a prune_from_up_buffer lambda to eliminate duplication in the same block.

### DIFF
--- a/src/lib/index/wb_cache.cpp
+++ b/src/lib/index/wb_cache.cpp
@@ -432,11 +432,13 @@ void IndexWBCache::link_buf(IndexBufferPtr const& up_buf, IndexBufferPtr const& 
     down_buf->m_up_buffer = real_up_buf;
 
     if (down_buf->state() == index_buf_state_t::CLEAN) {
-        //In root collapse case, down_buf is written before up_buf, but in root split case, down_buf is written after up_buf.
+        // In root collapse case, down_buf is written before up_buf, but in root split case, down_buf is written after
+        // up_buf.
         LOGDEBUGMOD(wbcache,
-                   "CLEAN_BUF_DEBUG: Adding CLEAN down_buf {} to up_buf {}, up_buf wait_count before={}, might be dirtied later",
-                   down_buf->blkid().to_integer(), real_up_buf->blkid().to_integer(),
-                   real_up_buf->m_wait_for_down_buffers.get());
+                    "CLEAN_BUF_DEBUG: Adding CLEAN down_buf {} to up_buf {}, up_buf wait_count before={}, might be "
+                    "dirtied later",
+                    down_buf->blkid().to_integer(), real_up_buf->blkid().to_integer(),
+                    real_up_buf->m_wait_for_down_buffers.get());
     }
 
     real_up_buf->add_down_buffer(down_buf);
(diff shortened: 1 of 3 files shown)


src/tests/test_index_crash_recovery.cpp
```diff
@@ -1416,6 +1420,334 @@ TEST_F(IndexCrashTestTwoTables, MultiTableMetaBufOrdinalCollisionOnRecovery) {
...
+    this->wait_for_crash_recovery(true);
+
+    LOGINFO("Step 6: do sanity check");
+    // 1 all the recovered nodes should not exist in m_recovery_freed_node_ids either ( should not be freed during
+    // recovery)
+    auto const recovered_nodes = inspectable_bt()->collect_node_ids();
+    for (const auto id : m_recovery_freed_node_ids) {
+        ASSERT_FALSE(recovered_nodes.contains(id))
+            << "created_freed_node " << id << " was freed and also recovered as a live node";
+    }
```
JacksonYao287, line 1732: this is the critical case that this PR want to fix. 

a bnode has already be allocated to a btree(recovered_nodes holds all the nodes that belongs to this btree ATM), but this node is finally freed (m_recovery_freed_node_ids holds all the freed nodes , which exists in the deleted_bufs) at the end of recovery.

this case will lead to corruption if this node is allocated to other btree node

src/tests/test_index_crash_recovery.cpp
```diff
@@ -1416,6 +1420,334 @@ TEST_F(IndexCrashTestTwoTables, MultiTableMetaBufOrdinalCollisionOnRecovery) {
...
+
+    // 2 all the created+freed blkids should not be freed during recovery ( should not appear in
+    // m_recovery_freed_node_ids)
+    for (auto const id : created_freed_node_ids) {
+        ASSERT_FALSE(m_recovery_freed_node_ids.contains(id))
+            << "created_freed_node " << id << " was freed during recovery";
+    }
```
JacksonYao287, line 1739: we check if a created_freed_node will be freed after recovery.

src/tests/test_index_crash_recovery.cpp
```diff
@@ -1416,6 +1420,334 @@ TEST_F(IndexCrashTestTwoTables, MultiTableMetaBufOrdinalCollisionOnRecovery) {
...
+    insert_sparse_multiples(0, preload_last_key, sparse_step);
+    test_common::HSTestHelper::trigger_cp(true);
+
+    LOGINFO("Step 1b: persistently consume index free queue so crash-CP created nodes are near recovery queue head");
+
+    // simulate a case that a lot of blk is consumed by other btree node allocation. for example, we have 2 btrees. one
+    // is used for reproduce this issue(the current one) , the other is used for consuming blks(receives lots of put
+    // request and lead to a lot of  blk allocation). commit_index_free_queue_until is used to consume free blks , just
+    // like what the second btree does.
+    commit_index_free_queue_until(crash_cp_free_queue_blks);
```
JacksonYao287, line 1695: commit_index_free_queue_until is used to consume blks. 

yes , we can do it by insert a lot of keys , but that will be very slow. so , here I directly consume blks from vdev.

src/lib/index/wb_cache.cpp
```diff
@@ -609,48 +611,55 @@ void IndexWBCache::recover(sisl::byte_view sb) {
...
+        buf->m_up_buffer = nullptr;
+    };
+
     LOGTRACEMOD(wbcache, "\n\n\nRecovery processing begins\n\n\n");
     for (auto const& [_, buf] : bufs) {
         load_buf(buf);
 
         if (buf->m_node_freed) {
             LOGTRACEMOD(wbcache, "recovering free buf {}", buf->to_string());
+            if (buf->m_created_cp_id == icp_ctx->id()) {
```
JacksonYao287, line 630: if a node is freed and the node is created in the crash cp (buf->m_created_cp_id == icp_ctx->id()), do not put it into the deleted_bufs.

this is the main fix of this PR.

src/lib/index/wb_cache.cpp
```diff
@@ -609,48 +611,55 @@ void IndexWBCache::recover(sisl::byte_view sb) {
...
+                }
+                continue;
+            }
             if (was_node_committed(buf)) {
-                // Mark this buffer as deleted, so that we can avoid using it anymore when repairing its parent's link
+                // Node data is on disk from a previous CP; mark it deleted so repair skips its stale
+                // parent link, schedule the blkid for release, and queue the parent for repair.
                 r_cast< persistent_hdr_t* >(buf->m_bytes)->node_deleted = true;
-                write_buf(nullptr, buf, icp_ctx); // no need to write it here !!
+                write_buf(nullptr, buf, icp_ctx);
```
shosseinimotlagh: delete this line which is not needed.

JacksonYao287: **Context**
Line 647 is in the branch for a node that was:
marked freed (m_node_freed == true)
created in a previous CP (not the same-CP create+freed case)
already committed to disk (was_node_committed(buf) == true)

```
r_cast<persistent_hdr_t*>(buf->m_bytes)->node_deleted = true;
write_buf(nullptr, buf, icp_ctx);   // line 647
deleted_bufs.push_back(buf);
pending_bufs.push_back(buf->m_up_buffer);
```

Line 646 only updates the in-memory copy in the recovery buf_map. Line 647 writes that flag to the disk.

**What write_buf does in recovery**
In recovery mode, write_buf does not add the buffer to the dirty list. It immediately sync-writes to disk:

```
if (m_in_recovery) {
    m_vdev->sync_write(buf->raw_buffer(), m_node_size, buf->m_blkid);
}
```
So its job here is to persist node_deleted = 1 before repair runs.

**Why it cannot be removed**

**1. Repair re-reads nodes from disk, not from the recovery buf_map**
During recovery, read_buf skips the cache and always reads from the vdev:

```
if (!m_in_recovery && m_cache.get(blkid, node)) { return; }
m_vdev->sync_read(..., blkid);
```

That means Line 646 updates only the recovery buffer in memory.
Without line 647, the on-disk copy still has node_deleted = 0.

**2. Downstream repair logic depends on the on-disk flag**
Paths like delete_stale_links and repair_links call read_node_impl, then check is_node_deleted():

```
if (child_node->is_node_deleted()) {
    // remove stale child link from parent
}
```

If write_buf is omitted, the freed node is read from disk as still live. The parent keeps a stale pointer to it, and repair can take the wrong path.

shosseinimotlagh: so the bug is actually second read of this buffer in delete_stale_links.right? This needs to fix to avoid one extra write. since functionality is served here, we can keep and later on, pass the already-deleted bufs to delete_stale_links and repair_links

JacksonYao287: no, the bug this pr try to fix is not second read of this buffer in delete_stale_links, it is described here
https://github.com/eBay/HomeStore/pull/892#issue-4610838164

shosseinimotlagh:
./test_index_crash_recovery --gtest_filter=*long_running_remove_crash* --seed=1784695170680725942 --log_mods=wbcache:0,base:0

[07/21/26 21:47:02-07:00] [T] [245408] [wb_cache.cpp:456:free_buf] Freeing bkid = 281479271678062. Remove from cache?(aka not recovery mode) = true
[07/21/26 21:47:02-07:00] [T] [245408] [wb_cache.cpp:352:transact_bufs] tranasction till now: cp: 96 : 0 - parent=281479271678276 child=281479271678330 new=[281479271678700, ] freed=[281479271678634, 281479271678062, ]
[07/21/26 21:47:02-07:00] [D] [245408] [test_index_crash_recovery.cpp:443:reapply_after_crash] Reapply: Removing key 3348
[07/21/26 21:47:02-07:00] [D] [245408] [test_index_crash_recovery.cpp:443:reapply_after_crash] Reapply: Removing key 3176
[07/21/26 21:47:02-07:00] [D] [245408] [test_index_crash_recovery.cpp:443:reapply_after_crash] Reapply: Removing key 1510
[07/21/26 21:47:02-07:00] [D] [245408] [test_index_crash_recovery.cpp:443:reapply_after_crash] Reapply: Removing key 2070
[07/21/26 21:47:02-07:00] [D] [245408] [test_index_crash_recovery.cpp:443:reapply_after_crash] Reapply: Removing key 1615
[07/21/26 21:47:02-07:00] [D] [245408] [test_index_crash_recovery.cpp:443:reapply_after_crash] Reapply: Removing key 15
[07/21/26 21:47:02-07:00] [D] [245408] [test_index_crash_recovery.cpp:443:reapply_after_crash] Reapply: Removing key 2609
[07/21/26 21:47:02-07:00] [D] [245408] [test_index_crash_recovery.cpp:443:reapply_after_crash] Reapply: Removing key 1360
[07/21/26 21:47:02-07:00] [D] [245408] [test_index_crash_recovery.cpp:443:reapply_after_crash] Reapply: Removing key 425
[07/21/26 21:47:02-07:00] [D] [245408] [test_index_crash_recovery.cpp:443:reapply_after_crash] Reapply: Removing key 1915
[07/21/26 21:47:02-07:00] [C] [245408] [wb_cache.cpp:117] ******************** Assertion failure: =====> Expected 'false' to be == to 'true' Unable to add alloc'd node to cache, low memory or duplicate inserts?
test_index_crash_recovery: /home/shosseinimotlagh/homestore/src/lib/index/wb_cache.cpp:117: virtual homestore::BtreeNodePtr homestore::IndexWBCache::alloc_buf(uint32_t, homestore::node_initializer_t&&): Assertion `0' failed.
Aborted

shosseinimotlagh (approved):
LG now although there is some bugs (may or may not related to this pr) . Later triaging this bug.

JacksonYao287:
thanks @shosseinimotlagh for the review , I tried and found it is an existing issue in the current code, not newly brought by this PR.  so I will merger this PR first and then try to fix this issue.

PR #894, opened 2026-06-18
SDSTOR-22293: handle logdev recovery corner case

Besroy:


Besroy:
@JacksonYao287  Since there isn’t a clean, fully compatible way to handle this case right now (for example, the two sbs in logstore are not atomic), I only changed logstream recovery in this PR. PTAL.

xiaoxichen (commented):
TL;DR

Instead of fighting with the corner cases, it seems like we dont need the dynamic feature at all, we knows the journal vdev size and we know our #PG    per SM,  any possibility of static allocation / pre-fill works with minimal modification ?

JacksonYao287:
it`s a big change, will review it later this week.

Besroy:
> TL;DR
> 
> Instead of fighting with the corner cases, it seems like we dont need the dynamic feature at all, we knows the journal vdev size and we know our #PG per SM, any possibility of static allocation / pre-fill works with minimal modification ?

 I don't think static allocation/pre-fill would be a minimal modification. It would fundamentally change the chunk chain structure for each PG into a pre-allocated ring, requiring significant redesign across several areas:                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                           
  - Truncation logic: instead of releasing chunks back to the pool, it becomes a ring-pointer advance
  - Wrap-around handling: offset_to_chunk(), tail_offset(), and bounds checks all need circular semantics                                                                                                                                                                                                                                                                  
  - New edge cases to evaluate: what happens when data space is exhausted but the log hasn't caught up (data-full vs. log-full mismatch), and how to handle uneven request distribution across PGs (one PG's ring fills up while another sits idle)
  - Superblock changes: data_start_offset would need to become a relative offset within the ring, and we'd need to decide between a fixed ring head vs. a dynamic head marker — either way, the on-disk format changes                                                                                                                                                     

If we want to handle this issue, this PR is a quick fix. If we want to revisit the chunk allocation model, that should be treated as a standalone architectural initiative — worth looping in all affected members to evaluate and plan for the right release.

src/lib/logstore/log_dev.cpp
```diff
@@ -233,6 +233,46 @@ void LogDev::do_load(off_t device_cursor) {
...
+
+            // Scope and known limitations of this recovery:
+            // - This fix is scoped to logstream positioning only. It does NOT replay the full set
+            //   of meta changes that the aborted LogDev::truncate() had in-flight, specifically:
+            //   logstore superblock updates (start LSN), unreserve-store, and rollback-info cleanup
+            //   (log_dev.cpp LogDev::truncate()). Those changes may not have been persisted.
+            m_log_idx.store(hint->head_start_idx, std::memory_order_release);
+            m_vdev_jd->update_data_start_offset(hint->head_start_offset);
+            device_cursor = hint->head_start_offset;
+            do_load(device_cursor);
```
JacksonYao287, line 278: not sure do we need to update m_logdev_meta and persist with the correct offset in private data?

Besroy: I’m not entirely sure either. I’m a bit concerned that it could introduce some unknown intermediate states.
But if we don’t persist it, the recovery becomes re-entrant, and it can be automatically recovered during the next truncate.
One thing I’m still unsure about is whether we should also recover the logstore start_lsn (for example,by passing persist_now=true in truncate’s update_store_superblk), which would ensure that after recovery the NuRaft layer won’t mistakenly think it still has stale logs and trigger unexpected behavior.
But it feels like forcing an extra logstore persistence on every truncate just for this corner case might be unnecessary...

src/lib/logstore/log_dev.cpp
```diff
@@ -664,8 +710,29 @@ uint64_t LogDev::truncate() {
...
+    // Persist per-store start LSN before vdev truncation to avoid the following:
+    // If store superblocks are not persisted here and we crash after vdev truncate but before the
+    // final persist below, corner-case recovery still fixes logstream positioning via the head-chunk
+    // hint, but Raft start_index() loads a stale (too-low) m_first_seq_num from disk, making this
+    // node appear to retain more logs than it actually does and potentially triggering baseline resync.
+    m_logdev_meta.persist();
+
     // Truncate them in vdev
-    m_vdev_jd->truncate(min_safe_ld_key.dev_offset);
+    m_vdev_jd->truncate(min_safe_ld_key.dev_offset, min_safe_ld_key.idx);
```
JacksonYao287, line 721: when we call m_vdev_jd->truncate, there are some cases that some chunks will[ be released to journal dev chunk pool](https://github.com/eBay/HomeStore/blob/28b69210feac1272c665f0e4fcef5e97aa2cf140/src/lib/device/journal_vdev.cpp#L658),  and thus they might be allocated for new logs and then be overwritten.

so, I think even if we try to correct the start offset and start_idx here when crash recovery, the overwritten chunk data can not be get back.

JacksonYao287 (commented):
the root cause is that we can not guarantee the atomicity for the two metablks(logdev metablk and journal descriptor metablk) in a logdev.

if we want to thoroughly solve this issue, we need to think through. 

since this is not a blocker , suggest to put it into next sprint.

Besroy:
> the root cause is that we can not guarantee the atomicity for the two metablks(logdev metablk and journal descriptor metablk) in a logdev.
> 
> if we want to thoroughly solve this issue, we need to think through.
> 
> since this is not a blocker , suggest to put it into next sprint.

agree, let's put it into 0.5.7

src/lib/logstore/log_dev.cpp
Besroy: Can you explain more? The start offset and start_idx are mapped to the new chunk list, IIUC the new chunks are not overwritten.

PR #895, opened 2026-06-22
Run fetch_pending_data on a dedicated reactor

xiaoxichen:
Previously fetch_pending_data shared the raft_repl_svc_timer fiber with flush_durable_commit_lsn, gc_repl_*, and monitor_replace_member_*. When flush_durable_commit_lsn blocked on synchronous metablk I/O (under m_rd_map_mtx), queued fetch batches sat past consensus.data_receive_timeout_ms (10s), tripping the 'Data fetch timeout' assertion / TIMEOUT path.

Spawn a dedicated raft_repl_fetcher reactor (1 fiber) that owns only the fetch timer. Queue and locking stay global; only the consumer thread changes.

### DIFF
--- a/src/lib/replication/service/raft_repl_service.cpp
+++ b/src/lib/replication/service/raft_repl_service.cpp
@@ -645,19 +645,6 @@ void RaftReplService::start_repl_service_timers() {
                     gc_repl_devs();
                 });
 
-            // Check for queued fetches at the minimum every second
-            uint64_t interval_ns = std::min(
-                HS_DYNAMIC_CONFIG(consensus.wait_data_write_timer_ms) * 1000 * 1000, 1ul * 1000 * 1000 * 1000);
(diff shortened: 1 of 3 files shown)


src/lib/replication/service/raft_repl_service.cpp
JacksonYao287, line 727: NIT: we`d better add check to see the state of rdev here. only for the repl_dev with a state of  repl_dev_stage_t::ACTIVE, we do check and fetch data.

for pending fetch of repl_dev with other state, we can probably drop it.

JacksonYao287 (approved):
LGTM

src/lib/replication/service/raft_repl_service.cpp
xiaoxichen: yeah I dont want to change logic as of now. But it is a good point.

JacksonYao287: feel free to merge it , or you can do it if you want in this PR .

Besroy (approved):
LGTM

PR #896, opened 2026-07-13
[SDSTOR-23023] Add option to commit_blk for post-crash resync

koujl:
https://jirap.corp.ebay.com/browse/SDSTOR-23023

Normally commit_blk is called either during recovery (which advances both watermarks) or after a fresh alloc. However, in some scenarios, a blk_id may be recorded to be live by the caller but the allocator watermarks have not been successfully persisted.

The new `force_cache` parameter allows the caller to unconditionally advance both m_last_append_offset and m_commit_offset, covering such cases without requiring the system to be in initialising mode.

Related PR: https://github.com/eBay/HomeObject/pull/452

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "7.5.11"
+    version = "7.5.12"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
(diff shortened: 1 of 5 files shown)


Besroy (dismissed):
Thanks for the fix

JacksonYao287 (dismissed):

xiaoxichen (commented):
the `force_cache` is a bit wired, maybe `force` is enough or pick a better name (the cache expose internal implementation and wired to api user)

xiaoxichen:
other part LGTM

koujl:
> the force_cache is a bit wired, maybe force is enough or pick a better name (the cache expose internal implementation and wired to api user)

Good idea, changed to `recommit` which better captures our intent (re-commit a allocated blkid). Updated the argument description accordingly.

Besroy (approved):

PR #898, opened 2026-07-17
SDSTOR-24056: hold lock when updating superblk of raft_repl_dev

JacksonYao287:
- Add mutex protection around superblock write in leave() to prevent data races
- Fix typos across comments and log strings


src/lib/replication/repl_dev/raft_repl_dev.h
```diff
@@ -454,7 +452,10 @@ class RaftReplDev : public ReplDev,
      * \param lsn The LSN to be checked.
      * \return true if the LSN is within the last snapshot LSN, false otherwise.
      */
-    bool need_skip_processing(const repl_lsn_t lsn) { return lsn <= m_rd_sb->last_snapshot_lsn; }
+    bool need_skip_processing(const repl_lsn_t lsn) {
+        // skip processing if the lsn is less than or equal to the last snapshot lsn or checkpoint lsn
+        return lsn <= m_rd_sb->last_snapshot_lsn || lsn <= m_rd_sb->checkpoint_lsn;
```
Besroy: what's the context for moving skip_cp_lsn to here? I think replay_log_path has already covered correctly and this function is just used for the BR-replay case (https://github.com/eBay/HomeStore/commit/924204611d29b8aea679e59a4da767fffeb51ad0)

xiaoxichen (commented):
I dont get the context of this fix in both jira and commit message...

src/lib/replication/repl_dev/raft_repl_dev.h
JacksonYao287: yes, you are right , I misunderstand need_skip_processing , I have reverted it

JacksonYao287:
>I dont get the context of this fix in both jira and commit message...

in this PR : 
1 I first update a lot of typos, no need to pay too much attention since they are only log and comment text changes.

2 the key logic change in this PR is in RaftReplDev::leave(), where a lock is added before updating the superblk of raft_repl_dev, this is the part the reviewer should pay more attention:

```
    {
        std::unique_lock lg{m_sb_mtx};
        m_rd_sb->destroy_pending = 0x1;
        m_rd_sb.write();
    }
```

the reason we need to add this lock is that many threads will try to update m_rd_sb concurrently, for example, RaftReplDev::cp_flush will update it in cp flush thread.


if we update it without a lock, when coming to [`MetaBlkService::update_sub_sb`](https://github.com/eBay/HomeStore/blob/7eafb3a5f271f835e5fe1f86e20133191b42d8ca/src/lib/meta/meta_blk_service.cpp#L805), it will first [calculated the crc of the input buffer](https://github.com/eBay/HomeStore/blob/7eafb3a5f271f835e5fe1f86e20133191b42d8ca/src/lib/meta/meta_blk_service.cpp#L824), and then write the input data buffer into disk. before the data buffer is written into disk, [the crc will be calculated again](https://github.com/eBay/HomeStore/blob/7eafb3a5f271f835e5fe1f86e20133191b42d8ca/src/lib/meta/meta_blk_service.cpp#L750)

so, if the input data buffer is changed  by another thread after the first crc calculation , then crc mismatch will occur.

the lock here is used to protect to data buffer not to be changed until update_sub_sb is completed. also, we can see the lock is hold before m_rd_sb.write() in other places in homestore code, but missed in RaftReplDev::leave(), so just add it here.

Besroy (approved):

PR #899, opened 2026-07-17
Fix BlkReadTracker firing waiter callback under hashmap write lock.

Besroy:
Move waiters out of the record before delete so m_cb() runs after the lock is released, avoiding deadlock when the callback requires other resources. Add a regression test to verify the fix.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "7.5.17"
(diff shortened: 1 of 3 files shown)


JacksonYao287 (dismissed):
Root cause analysis LGTM. Moving m_waiters out before delete n ensures ~blk_track_waiter / m_cb() runs after the bucket unique_lock is released。

JacksonYao287 (approved):
LGTM

Besroy:
This PR uses folly::xxx; let's put it on hold for now and decide whether to merge it after the folly migration done.

PR #902, opened 2026-08-04
SDSTOR-24093: Duplicate inserts in wb_cache

nnastonen:
Fix for "Duplicate inserts in wb_cache" bug.

In async_cp_flush(), m_vdev->cp_flush() is called only for cp_ctx->id() == 0.

That design misses the case where recovery repaired allocator state but left no dirty index buffers.

### DIFF
--- a/src/lib/index/wb_cache.cpp
+++ b/src/lib/index/wb_cache.cpp
@@ -565,6 +565,7 @@ void IndexWBCache::recover(sisl::byte_view sb) {
     auto cpg = cp_mgr().cp_guard();
     auto icp_ctx = r_cast< IndexCPContext* >(cpg.context(cp_consumer_t::INDEX_SVC));
     std::map< BlkId, IndexBufferPtr > bufs = icp_ctx->recover(std::move(sb));
+    bool allocator_state_changed{false};
 
     LOGINFOMOD(wbcache, "Detected unclean shutdown, prior cp={} had to flush {} nodes, recovering... ", icp_ctx->id(),
                bufs.size());
@@ -657,7 +658,8 @@ void IndexWBCache::recover(sisl::byte_view sb) {
                             was_node_committed(buf->m_up_buffer));
                 buf->m_node_freed = false;
                 r_cast< persistent_hdr_t* >(buf->m_bytes)->node_deleted = false;
-                m_vdev->commit_blk(buf->m_blkid);
+                auto alloc_status = m_vdev->commit_blk(buf->m_blkid);
+                allocator_state_changed |= (alloc_status == BlkAllocStatus::SUCCESS);
                 if (buf->m_node_level) { potential_parent_recovered_bufs.insert(buf); }
                 prune_from_up_buffer(buf);
             }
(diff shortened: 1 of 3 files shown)


src/lib/index/wb_cache.cpp
```diff
@@ -657,7 +658,8 @@ void IndexWBCache::recover(sisl::byte_view sb) {
                             was_node_committed(buf->m_up_buffer));
                 buf->m_node_freed = false;
                 r_cast< persistent_hdr_t* >(buf->m_bytes)->node_deleted = false;
-                m_vdev->commit_blk(buf->m_blkid);
+                auto alloc_status = m_vdev->commit_blk(buf->m_blkid);
```
JacksonYao287, line 660: I would suggest to add a HS_REL_ASSERT(alloc_status == BlkAllocStatus::SUCCESS) to not ignore any failure when committing blk, since it will be dangerous to go ahead with an uncommitted blk here

JacksonYao287 (commented):
thanks @nnastonen  for this PR, the RCA looks good to me. I have several comments.

1 pls bump conan version.

2 this PR only handles the blk_commit case,  but not handle the blk_free case, which also need to be persisted by vdev->cp_flush.  so , my suggeesion is if there is any blk_commit or blk_free in a cp, we need to call vdev->cp_flush in index cp_flush.  actually , if there is any blk_commit or free in a cp, we need to trigger vdev->cp_flush, no matter the dirty list is empty

3 it is better to have a UT for this fix.

src/lib/index/wb_cache.cpp
nnastonen: Before this PR the result of m_vdev->commit_blk() was ignored and with the flow covered by this PR, unsuccessful status will not trigger the flush. I think fixing the error-path is a different concern and should be documented and implemented separately.

JacksonYao287: sure。 I am still thinking that since vdev->cp_flush is very light-weight, and for those non-dirtied blk_allocator ，[ it`s no-op](https://github.com/eBay/HomeStore/blob/0bdc47f4e17f22abd9abcdb2b5d69fb7c669d81f/src/lib/blkalloc/bitmap_blk_allocator.cpp#L60-L61 ).

so, I think it will be simpler for this PR to always call m_vdev->cp_flush() in async_cp_flush（no need to check allocator_state_changed）(even if cp_ctx->id() is not zero when there is not any dirty buffer). it will be just several line changes and can cover all cases

nnastonen: Addressed.

JacksonYao287:
@nnastonen  thank you for the update, the changes looks good to me. 
before merging, pls also update conan version in conan.py from 7.5.15 to 7.5.16 ， otherwise the check version work follow will fail.

@shosseinimotlagh do you have any input?

JacksonYao287 (approved):
LGTM

shosseinimotlagh:
This PR looks good to me! Thank you Niko for addressing it and thank you Jie to review it

PR #904, opened 2026-08-28
SDSTOR-20656: Added wbc epochs and UT

nnastonen:
0. Create homestore with chunk selector
1. Create btree1 and populate it
2. Nodes get into wbc cache
3. Destroy btree1 (wbc purge is not done, also volume's chunk bitmap is not cleared)
4. Create btree2 and populate it
5. Nodes get into wbc cache and we get cache corruption, duplicate inserts

This change adds a generation counter per chunk to invalidate stale WBC entries without scanning or deleting them one by  one when custom chunk selector is in use.

When a btree is destroyed with chunk selector enabled, its chunks can be reset and reused by a later btree. The problem is that old WBC entries may still exist for the same BlkIds. When a new btree reuses those BlkIds, WBC can see the old cached entry and treat the new allocation as a duplicate insert, or return stale data.

Instead, each chunk now has a monotonically increasing epoch. Every IndexBuffer records the epoch of the chunk it belongs to. On chunk reset, the chunk epoch is bumped. After that, any cached node whose stored epoch no longer matches the chunk's current epoch is treated as stale and evicted lazily on the next cache access or allocation.

This gives us O(1) invalidation at chunk-reset time, avoids walking millions of WBC entries during destroy, and makes BlkId reuse safe after chunk reset.

If chunk selector is not in use and chunk epochs are never bumped, the stale-entry checks are effectively no-ops.

### DIFF
--- a/src/lib/device/chunk.h
+++ b/src/lib/device/chunk.h
@@ -25,9 +25,10 @@ private:
     PhysicalDev* const m_pdev;
     const uint32_t m_chunk_slot;
     const uint32_t m_stream_id;
-    uint32_t m_vdev_ordinal{0};
+    uint32_t m_vdev_ordinal{};
     shared< BlkAllocator > m_blk_allocator;
     float blk_usage_report_threshold{0.9};
+    std::atomic<uint64_t> m_wbc_epoch{};
 
 public:
     static constexpr auto MAX_CHUNK_SIZE = std::numeric_limits< uint32_t >::max();
@@ -42,7 +43,7 @@ public:
     Chunk& operator=(Chunk&&) noexcept = delete;
     virtual ~Chunk() = default;
 
-    /////////////// Pointer Getters ////////////////////
+    /////////////// Pointer Getters /////////////////////
     const PhysicalDev* physical_dev() const { return m_pdev; }
     PhysicalDev* physical_dev_mutable() { return m_pdev; };
     const chunk_info& info() const { return m_chunk_info; }
@@ -84,6 +85,9 @@ public:
 
     ////////////// Misc ////////////////////////
     void reset_block_allocator();
+
+    uint64_t wbc_epoch() const { return m_wbc_epoch.load(std::memory_order_acquire); }
+    void bump_wbc_epoch() { m_wbc_epoch.fetch_add(1, std::memory_order_acq_rel); }
 
 private:
     void write_chunk_info();
--- a/src/lib/index/wb_cache.cpp
+++ b/src/lib/index/wb_cache.cpp
@@ -94,33 +94,53 @@ void IndexWBCache::start_flush_threads() {
 
 BtreeNodePtr IndexWBCache::alloc_buf(uint32_t ordinal, node_initializer_t&& node_initializer) {
     auto cpg = cp_mgr().cp_guard();
-    auto cp_ctx = r_cast< IndexCPContext* >(cpg.context(cp_consumer_t::INDEX_SVC));
 
     // Alloc a block of data from underlying vdev
     MultiBlkId blkid;
     // Ordinal used as a hint in the case of custom chunk selector exists
     blk_alloc_hints hints;
     hints.application_hint = ordinal;
-    auto ret = m_vdev->alloc_contiguous_blks(1, hints, blkid);
-    if (ret != BlkAllocStatus::SUCCESS) { return nullptr; }
+
+    if (m_vdev->alloc_contiguous_blks(1, hints, blkid) != BlkAllocStatus::SUCCESS) {
+        return nullptr;
+    }
 
     // Alloc buffer and initialize the node
     auto idx_buf = std::make_shared< IndexBuffer >(blkid, m_node_size, m_vdev->align_size());
     idx_buf->m_created_cp_id = cpg->id();
     idx_buf->m_dirtied_cp_id = cpg->id();
+    idx_buf->m_chunk_epoch = current_chunk_epoch(idx_buf->blkid());
     auto node = node_initializer(idx_buf);
     idx_buf->m_node_level = node->level();
 
+    // Add the node to the cache. Skip if we are in recovery mode.
     if (!m_in_recovery) {
-        // Add the node to the cache. Skip if we are in recovery mode.
-        bool done = m_cache.insert(node);
-        HS_REL_ASSERT_EQ(done, true, "Unable to add alloc'd node to cache, low memory or duplicate inserts?");
+retry_insert:
+        BtreeNodePtr existing;
+        if (m_cache.get(idx_buf->blkid(), existing)) {
+            if (!is_stale_node(existing)) {
+                HS_REL_ASSERT(false, "alloc_buf found live duplicate cache entry for blkid {}",
+                              idx_buf->blkid().to_string());
+            }
+            LOGTRACEMOD(wbcache, "alloc_buf evicting stale node for blkid {}", idx_buf->blkid().to_string());
+            m_cache.remove(idx_buf->blkid(), existing);
+        }
+
+        if (!m_cache.insert(node)) {
+            if (m_cache.get(idx_buf->blkid(), existing) && is_stale_node(existing)) {
+                m_cache.remove(idx_buf->blkid(), existing);
+                goto retry_insert;
+            }
+            HS_REL_ASSERT(false, "Unable to add alloc'd node to cache, low memory or duplicate inserts?");
+        }
     }
 
-    // The entire index is updated in the commit path, so we alloc the blk and commit them right away
-    auto alloc_status = m_vdev->commit_blk(blkid);
-    // if any error happens when committing the blk to index service, we should assert and crash
-    if (alloc_status != BlkAllocStatus::SUCCESS) HS_REL_ASSERT(0, "Failed to commit blk: {}", blkid.to_string());
+    // The entire index is updated in the commit path, so we alloc the blk and commit them right away.
+    // If any error happens when committing the blk to index service, we should assert and crash
+    if (m_vdev->commit_blk(blkid) != BlkAllocStatus::SUCCESS) {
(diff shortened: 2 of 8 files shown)


shosseinimotlagh:
what is the memory increase due to the introduction the notion of epoch?

src/lib/index/wb_cache.cpp
```diff
@@ -160,8 +200,15 @@ void IndexWBCache::read_buf(bnodeid_t id, BtreeNodePtr& node, node_initializer_t
...
-            goto retry;
+            BtreeNodePtr cached_node;
+            if (m_cache.get(blkid, cached_node) && is_stale_node(cached_node)) {
+                LOGTRACEMOD(wbcache, "stale cache race for blkid {}, removing and retrying", blkid.to_string());
+                BtreeNodePtr removed;
+                m_cache.remove(blkid, removed);
+                goto retry;
+            }
+
+            HS_REL_ASSERT(false, "Failed to insert read buf {} into cache", blkid.to_string());
```
JacksonYao287: assuming two threads try read_buf for the same blkid simultaneously. if the blkid does not exist in wb_cache,  the two threads will read this blk from vdev and then try to insert it into wb_cache(line 201, m_cache.insert(node);). eventually, only one thread will succeed, and the other will retry and will succeed get the blk from wb_cache. this is an expected scenario.

according to the changes here(especially line 204), if the thread , which failed at line 201, succeeds to get the blkid(since the other thread has already inserted it into cache), and this node is not a stale_node(probably happens), it will hit line 211 and lead to assert failure.

src/lib/index/wb_cache.hpp
```diff
@@ -93,5 +93,11 @@ class IndexWBCache : public IndexWBCacheBase {
     void load_buf(IndexBufferPtr const& buf);
     void update_up_buffer_counters(IndexBufferPtr const& buf);
     void prune_up_buffers(IndexBufferPtr const& buf, std::vector< IndexBufferPtr >& bufs_to_repair);
+
+    // Helpers for maintaining cache validity
+    uint64_t current_chunk_epoch(const BlkId& blkid) const;
+    // These two always return false is chunk selector is not in use
```
JacksonYao287: typo: is -> if

src/lib/device/chunk.h
```diff
@@ -25,9 +25,10 @@ class Chunk {
     PhysicalDev* const m_pdev;
     const uint32_t m_chunk_slot;
     const uint32_t m_stream_id;
-    uint32_t m_vdev_ordinal{0};
+    uint32_t m_vdev_ordinal{};
     shared< BlkAllocator > m_blk_allocator;
     float blk_usage_report_threshold{0.9};
+    std::atomic<uint64_t> m_wbc_epoch{};
```
JacksonYao287, line 31: suggest rename m_wbc_epoch to m_epoch, since chunk in not only used by index_service, it is also other services in homestore and these services have no idea of what wbc is.

JacksonYao287 (commented):
according to this PR, if we call chunk->reser_blk_allocator, the epoch will be bumped.  this is a behavior change. chunk is not only used by index service, it is also used by other services in homestore. other service or upper layer(like homeobject) will call chunk->reset() in there logic directly. Although epoch brings no impact as of now, not sure will this happen in the future.

IMO, it`s better to not offload checking stale blk to wb_cache itself. it is the responsibility of upper layer to explicitly mark the blk stale by free_node(destory_btree will call free_node). 

I don`t have a very strong opinion of this PR. I am not familiar with nublox code. if we can manage to do this in volume destroy phase, it will be better. do we have any workaround in nublox?

src/lib/index/wb_cache.cpp
nnastonen: Thank you, very good observation. Fixed.

src/lib/index/wb_cache.hpp
nnastonen: Fixed.

src/lib/device/chunk.h
nnastonen: Since epochs are related only to wbc, I'd keep the name.

src/lib/index/wb_cache.cpp
```diff
@@ -94,33 +94,53 @@ void IndexWBCache::start_flush_threads() {
...
+            if (!is_stale_node(existing)) {
+                HS_REL_ASSERT(false, "alloc_buf found live duplicate cache entry for blkid {}",
+                              idx_buf->blkid().to_string());
+            }
+            LOGTRACEMOD(wbcache, "alloc_buf evicting stale node for blkid {}", idx_buf->blkid().to_string());
+            m_cache.remove(idx_buf->blkid(), existing);
+        }
+
+        if (!m_cache.insert(node)) {
+            if (m_cache.get(idx_buf->blkid(), existing) && is_stale_node(existing)) {
```
raakella1: Can you explain the rational for a blanket retry here? Is there a case where someone can reinsert the same blkid between the previous remove on line 126 and the insert on  129?

src/lib/index/wb_cache.cpp
```diff
@@ -94,33 +94,53 @@ void IndexWBCache::start_flush_threads() {
...
+                m_cache.remove(idx_buf->blkid(), existing);
+                goto retry_insert;
```
raakella1: The goto statement is not very readable. We can achieve the same code by a while(true),

src/lib/index/wb_cache.cpp
```diff
@@ -134,35 +154,54 @@ void IndexWBCache::write_buf(const BtreeNodePtr& node, const IndexBufferPtr& buf
...
+        buf->m_chunk_epoch = current_chunk_epoch(buf->blkid());
+        m_cache.upsert(node);
+    }
+
+    LOGTRACEMOD(wbcache, "add to dirty list cp {} {}", cp_ctx->id(), buf->to_string());
+    r_cast<IndexCPContext*>(cp_ctx)->add_to_dirty_list(buf);
+    resource_mgr().inc_dirty_buf_size(m_node_size);
 }
 
 void IndexWBCache::read_buf(bnodeid_t id, BtreeNodePtr& node, node_initializer_t&& node_initializer) {
```
raakella1: similar comments as alloc_buf

raakella1:
I have no additional comments to the ones already posted here. You can merge after resolving the existing ones

src/lib/device/chunk.h
shosseinimotlagh: I agree with Jie too. this should be generic.

src/lib/index/wb_cache.cpp
shosseinimotlagh: @nnastonen would you please check if this happened. IIRC after acquiring the lock the other one will get the refresh lock and this scenario shouldnt happen. Please validate before change.

src/lib/index/wb_cache.cpp
```diff
@@ -94,33 +94,53 @@ void IndexWBCache::start_flush_threads() {
...
 
+    // Add the node to the cache. Skip if we are in recovery mode.
     if (!m_in_recovery) {
-        // Add the node to the cache. Skip if we are in recovery mode.
-        bool done = m_cache.insert(node);
-        HS_REL_ASSERT_EQ(done, true, "Unable to add alloc'd node to cache, low memory or duplicate inserts?");
+retry_insert:
+        BtreeNodePtr existing;
+        if (m_cache.get(idx_buf->blkid(), existing)) {
+            if (!is_stale_node(existing)) {
```
shosseinimotlagh, line 121: for each stale buffer of the current chunk, does this check cause performance degradation ?

JacksonYao287:
I have no more comments for this PR, but I would suggest to think through again if we have any workaround in nublox side to destroy btree gracefully and not touch the wbcache/chunk logic for this specific case.  theoretically, the destroy_btree should be called for a btree if we do want to destroy it. offloading this to wbcache/chunk seems a little wired.

if you think this change is indeed necessary for nublox case after thinking through, you can merge it.

src/lib/index/wb_cache.cpp
nnastonen: In my microbenchmark, creating and re-creating a btree with ~200k nodes and ~100k stale cache hits took about the same time. No measurable degradation.

shosseinimotlagh (approved):
LGTM, although still against `bump_wbc_epoch` in the chunk class . Leave it to author to make decision.

xiaoxichen (commented):
I think I need more context on the intent of this change...

1. it doesnt reduce works,  those entries in WBC still needed to be cleanup , it just spread the time cost by cleanup to future then makes the index destroy faster,   how much improvement can we get and why it is matters?

2. addition to #1,  an addition cost of check on each request had to be introduced due to lazy cleanup .  Yes it should be minor as it  just an atomic read (with barrier) however it needs to be reasoned together with #1,  what it the gain from this change?

3.  There is still a racing in flushing WBC vs chunk reset. The epoch checks currently protect cache lookup and allocation, but not dirty/in-flight WBC writes. A buffer can be selected for flush, then its chunk can be reset and reused, after which the old async write may overwrite the new owner’s data.

xiaoxichen:
A corrupting interleaving is:

1. Old buffer for  (chunk X, block Y)  is selected for flush.
2. Chunk X is reset and its epoch increments.
3. Another btree allocates and writes block Y.
4. The old WBC write is submitted or completes afterward, overwriting the new btree’s node.

nnastonen:
> A corrupting interleaving is:
> 
> 1. Old buffer for  (chunk X, block Y)  is selected for flush.
> 2. Chunk X is reset and its epoch increments.
> 3. Another btree allocates and writes block Y.
> 4. The old WBC write is submitted or completes afterward, overwriting the new btree’s node.

If the real concern is stale dirty buffers being written after chunk reset/reuse, then the epoch check should be enforced in the flush path as well, not only on cache access/allocation. If a buffer’s recorded epoch no longer matches the chunk epoch at write submission time, that write should be skipped/failed as stale.

PR #905, opened 2026-09-02
SDSTOR-25404: Cache eviction for destroyed volume fails

nnastonen:
If a volume has been already destroyed and eviction later is called and choose a node from this volume, the eviction will hit the SIGSEGV.


raakella1 (approved):
LG!

PR #912, opened 2026-09-10
SDSTOR-20656: clear wbc chunk selector

nnastonen:
0. Create homestore with chunk selector
1. Create btree1 and populate it
2. Nodes get into wbc cache
3. Destroy btree1 (wbc purge is not done, also volume's chunk bitmap is not cleared)
4. Create btree2 and populate it
5. Nodes get into wbc cache and we get cache corruption, duplicate inserts

This change makes chunk release asynchronous and defers returning the chunk to the free pool until cleanup is complete.

New flow:
- detach the volume's chunk list from m_volume_chunks
- run asynchronous cleanup for all the chunks
- in cleanup, evict all allocated blkids for the chunks from WBC
- [reset the chunks' block allocator state (bitmap)]
- reinsert the chunks into m_per_dev_chunks

This PR also addresses this ticket [SDSTOR-25404](https://jirap.corp.ebay.com/browse/SDSTOR-25404)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -66,3 +66,5 @@ add_library(homestore STATIC
     ${HOMESTORE_OBJECTS} 
 )
 target_link_libraries(homestore ${COMMON_DEPS})
+target_compile_definitions(homestore PRIVATE _PRERELEASE)
+
--- a/src/include/homestore/index/wb_cache_base.hpp
+++ b/src/include/homestore/index/wb_cache_base.hpp
@@ -25,11 +25,12 @@
 namespace homestore {
 
 class BtreeNode;
+class Chunk;
+struct CPContext;
+
 using BtreeNodePtr = boost::intrusive_ptr< BtreeNode >;
 using BtreeNodeList = boost::container::small_vector< BtreeNodePtr, 3 >;
 using node_initializer_t = std::function< BtreeNodePtr(const IndexBufferPtr&) >;
-
-struct CPContext;
 
 class IndexWBCacheBase {
 public:
@@ -67,6 +68,9 @@ public:
     /// @return
     // virtual IndexBufferPtr copy_buffer(const IndexBufferPtr& cur_buf, const CPContext* context) const = 0;
     virtual void recover(sisl::byte_view sb) = 0;
+
+    /// Remove entries for a specific chunk from wbc
+    virtual void evict_chunk_blkids(const Chunk& chunk) = 0;
 };
 
 } // namespace homestore
--- a/src/lib/device/chunk.cpp
+++ b/src/lib/device/chunk.cpp
@@ -84,4 +84,17 @@ void Chunk::reset_block_allocator() {
     vdev_ptr->reset_chunk_blk_allocator(this);
 }
 
+void Chunk::foreach_allocated_blk(std::function< void(blk_id const&) >&& cb) const {
+    auto* ba = blk_allocator();
+    RELEASE_ASSERT(ba, "Blk allocator is null for chunk_id={}", chunk_id());
+
+    const auto total_blks = ba->get_total_blks();
+    const auto cid = chunk_id();
+
+    for (blk_num_t blk_num{}; blk_num < total_blks; ++blk_num) {
+        blk_id blkid{blk_num, 1, cid};
+        if (ba->is_blk_alloced(blkid, false)) { cb(blkid); }
+    }
+}
+
 } // namespace homestore
--- a/conanfile.py
+++ b/conanfile.py
@@ -9,7 +9,7 @@ required_conan_version = ">=1.60.0"
 
 class HomestoreConan(ConanFile):
     name = "homestore"
-    version = "8.2.0"
+    version = "8.2.1"
 
     homepage = "https://github.com/eBay/Homestore"
     description = "HomeStore Storage Engine"
--- a/src/lib/device/chunk.h
+++ b/src/lib/device/chunk.h
@@ -16,8 +16,11 @@
 #include "device/physical_dev.hpp"
 
 namespace homestore {
+
 class BlkAllocator;
 class CP;
+struct blk_id;
+
 class Chunk {
 private:
     std::shared_mutex m_mgmt_mutex;
@@ -73,6 +76,8 @@ public:
     }
     float get_blk_usage_report_threshold() const { return blk_usage_report_threshold; }
     float get_blk_usage() const;
+
+    void foreach_allocated_blk(std::function< void(blk_id const&) >&& cb) const;
 
     ////////////// Setters /////////////////////
     void set_user_private(const sisl::blob& data);
--- a/src/lib/index/wb_cache.cpp
+++ b/src/lib/index/wb_cache.cpp
@@ -52,8 +52,13 @@ IndexWBCache::IndexWBCache(const std::shared_ptr< VirtualDev >& vdev, std::pair<
                     return static_cast< IndexBtreeNode* >(node.get())->m_idx_buf->m_blkid;
                 },
                 [](const sisl::CacheRecord& rec) -> bool {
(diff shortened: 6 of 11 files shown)


src/CMakeLists.txt
```diff
@@ -66,3 +66,5 @@ add_library(homestore STATIC
     ${HOMESTORE_OBJECTS} 
 )
 target_link_libraries(homestore ${COMMON_DEPS})
+target_compile_definitions(homestore PRIVATE _PRERELEASE)
```
JacksonYao287, line 69: why we need this _PRERELEASE here?  I think `target_compile_definitions(test_index_chunk_selector PRIVATE _PRERELEASE)` in `tests/CmakeLists.txt` is enough for the new test case

src/include/homestore/index/wb_cache_base.hpp
```diff
@@ -67,6 +68,9 @@ class IndexWBCacheBase {
     /// @return
     // virtual IndexBufferPtr copy_buffer(const IndexBufferPtr& cur_buf, const CPContext* context) const = 0;
     virtual void recover(sisl::byte_view sb) = 0;
+
+    /// Remove entries for a specific chunk from wbc
+    virtual void evict_chunk_blkids(const Chunk& chunk) = 0;
```
JacksonYao287, line 73: NIT: if this a time-consuming task, I would suggest make it return a sisl::async::task < void >, so that we can use co_wait to wait for the completion asyncronously?

src/lib/device/chunk.cpp
```diff
@@ -84,4 +84,17 @@ void Chunk::reset_block_allocator() {
...
 
+void Chunk::foreach_allocated_blk(std::function< void(blk_id const&) >&& cb) const {
+    auto* ba = blk_allocator();
+    RELEASE_ASSERT(ba, "Blk allocator is null for chunk_id={}", chunk_id());
+
+    const auto total_blks = ba->get_total_blks();
+    const auto cid = chunk_id();
+
+    for (blk_num_t blk_num{}; blk_num < total_blks; ++blk_num) {
+        blk_id blkid{blk_num, 1, cid};
```
JacksonYao287, line 95: I am not quite sure if we need a lock to guarantee the thread safety of this function

src/CMakeLists.txt
nnastonen: Because otherwise the code won't compile.

src/include/homestore/index/wb_cache_base.hpp
nnastonen: It is a time-consuming task, we call it from HomeBlocks within iomanager.run_on_forget(), check the other half of this PR: https://github.com/eBay/HomeBlocks/pull/179
We don't need to wait for it as it does all the job itself.

xiaoxichen:
@szmyd @shosseinimotlagh  Please if you can ensure the upgrade path and compatibility are properly discussed.

NuObject is moving to V7 mostly to intake the folly and co-routine removal, which is still a persistent-compatible version (meaning , no data migration is needed upgrading from v6 to v7). 

With the potential path forward of this feature,   it is unlikely to be compatible with v7 as of now.   I would suggest it should be in V8 with proper feature bit.  Also decisions regarding how V6 can upgrade to V8 should be discussed.

src/lib/device/chunk.cpp
nnastonen: Can you please elaborate? What exactly needs to be protected?
The chunk is not visible to chunk selector yet and SimpleCache seems to be thread safe.

src/CMakeLists.txt
JacksonYao287: adding _PRERELEASE unconditionally to the homestore library itself means every build (including production Release builds) will now permanently carry _PRERELEASE: extra fault-injection checkpoints, extra assert branches, and extra debug fields get baked into the production library. That's a much bigger behavioral change than "just make the new test compile." pls correct me if I misunderstand something.

src/include/homestore/index/wb_cache_base.hpp
JacksonYao287: IMO, if we do nott  need wait for it and it is a time-consuming task, then it would be better to return a sisl::async::task < void >, so that we can use `detach` to run this time-consuming task asyncronously and not block the calling thread

src/lib/device/chunk.cpp
JacksonYao287: honestly this is just my intuition . if a thread is calling free_blk or reset blk_allocator,  another thread is calling `foreach_allocated_blk`, not sure this will bring some thread safety issue.

src/CMakeLists.txt
nnastonen: You are right. I added _PRERELEASE just to make this compile and run. Please suggest a better place/way to enable _PRERELEASE as I didn't really find how it is enabled in 7.x. I presume it is not the same as Debug build.

src/include/homestore/index/wb_cache_base.hpp
nnastonen: iomanager.run_on_forget() is also async, it does not block the caller and I am not a big fan of detach() personally. So this is basically a stylistic matter, unless there is something I don't know about sisl::async.
Please elaborate how this would be better for **this particular case**.

src/lib/device/chunk.cpp
nnastonen: Nothing should touch (or even see) this chunk while chunk selector is in charge.

szmyd:
> Also decisions regarding how V6 can upgrade to V8 should be discussed.

It's my understanding that we only use v3 (NuBlox 1.x/2.0) and v7 (NuObject) at the moment. We're currently working on v8 for NuBlox 2.1; so where does v6 come into play?

xiaoxichen:
@szmyd sorry you are right, I mess up v7 vs v6.. Yes I am talking about v7 ->v8 compatibility

src/CMakeLists.txt
JacksonYao287: @szmyd  since we disable prerelease in https://github.com/eBay/sisl/pull/278, do you have any sugestion on this ?

src/lib/device/chunk.cpp
JacksonYao287: ok, we can keep it as this for now, and if there is any issue, we can revisit here. pls add a comment here 

"Nothing should touch (or even see) this chunk while chunk selector is in charge." for later revisit

src/include/homestore/index/wb_cache_base.hpp
JacksonYao287: when you submit a job by run_on_forget, it means you don`t care when the job is done and the result.

If I understand correctly, the chunk should be reused (selected by chunk selector again) only after all the chunk blk are evicted from wb_cache. so we should care when the job is done, right?  

if the signature  of this function is `virtual sisl::async::task<void> evict_chunk_blkids(const Chunk& chunk) = 0;`, then the code can be sisl::async::detach_then("evict_chunk_blkids", "return it back to chunk selector"), which means return the chunk back to selector after evict_chunk_blkids is done.

pls correct me if I misunderstand something

JacksonYao287 (commented):
thanks @nnastonen  for this change, I have no more in put. @szmyd @shosseinimotlagh can you give some inputs?

src/include/homestore/index/wb_cache_base.hpp
nnastonen: Yes, the lambda ran with run_on_forget() is put on a worker thread pool and runs asynchronously as soon as a worker can pick it up. There is no need to create unneeded overhead on scheduler by creating a new thread. The lambda does all the required work of releasing the chunks by itself, we **do not** wait for this to happen.
I still don't understand how sisl::async::task will help here.

PR #914, opened 2026-09-11
logstore: reschedule flush_if_necessary() on lost try_lock instead of dropping it

shosseinimotlagh:
## Scenario

`LogDev::flush_if_necessary()` decides whether pending data needs flushing, then tries to grab the flush mutex with `std::try_to_lock`:

```cpp
if (flush_by_size || flush_by_time) {
    std::unique_lock lck(m_flush_mtx, std::try_to_lock);
    if (lck.owns_lock()) {
        decr_pending_request_num();
        return flush();
    }
    // lost the race -> falls through, does nothing
}
```

If the `try_lock` loses the race — because some *other* flush is holding the mutex right at that instant (another write's own `flush_if_necessary()` call, or `HomeLogStore::truncate()`'s internal `flush()`) — this call just gives up. No retry, no rescheduling. That's fine as long as *something else* is guaranteed to call `flush_if_necessary()` again later for the same logdev — normally the periodic flush timer (`flush_timer_frequency_us`) fills that role.

The gap: if the periodic timer is disabled (`flush_timer_frequency_us = 0`) and this happens to be the *last* write issued for that logdev in a given run, nothing ever calls `flush_if_necessary()` again. The data behind that lost race sits unflushed indefinitely, and anything waiting on its completion callback blocks forever.

Concretely, this reproduces in `LogStoreTest.VarRateInsertThenTruncate` (`test_log_store.cpp`), which disables the timer globally for the whole binary and deliberately runs writes concurrently with repeated `truncate_validate()` calls in its Step 4 — exactly the write-vs-truncate mutex race described above, on the very last batch issued.

## Fix v1

When the `try_lock` fails, reschedule a follow-up `flush_if_necessary()` attempt via `iomanager.run_on_forget()`, mirroring the existing reschedule already used a few lines above for the `!can_flush_in_this_thread()` case. The concurrent flush holding the mutex will release it shortly, so the rescheduled attempt gets a real chance to succeed instead of the data being silently abandoned.

## Follow-up 1: the reschedule can still be silently dropped by a stale clock

Review caught that the reschedule just re-invokes `flush_if_necessary()`, which re-derives `flush_by_size`/`flush_by_time` from scratch. `LogDev::flush()` unconditionally resets `m_last_flush_time` at its very start, so the concurrent flush that just won the race may have already reset that clock by the time the retry runs — `flush_by_time` can read false again, and if this write's own size never crosses `threshold_size` on its own, `flush_by_size` stays false too, silently dropping the retry with nothing left to trigger it again.

Fix: `flush_if_necessary()` gains a `force` parameter. The reschedule now passes `force=true`, which skips the size/time re-derivation entirely and goes straight to `try_lock` — we already decided this data needs flushing, so the retry's only job is to keep trying to acquire the lock, not re-litigate whether to. Still a non-blocking `try_to_lock`, so no new deadlock surface.

## Follow-up 2: the reschedule itself could stack-overflow the process

Building a deterministic repro for follow-up 1 surfaced a separate, more severe, **pre-existing** bug in the original reschedule (present with or without `force`, i.e. it predates this PR's own fix): `IOReactor::deliver_msg` takes a same-thread shortcut, calling the target handler inline instead of queuing it whenever sender and receiver are the same reactor. The reschedule always targets `flush_thread()`, and by construction we're always already on `flush_thread()` when we reach it (that's how we got past `can_flush_in_this_thread()`), so every failed retry was a real recursive call on the same stack, not a queued task. Under sustained lock contention (confirmed with the lock held for ~200ms in testing) that recursion runs tens of thousands of frames deep and stack-overflows the process.

Fix: route the retry through `iomgr::reactor_regex::random_worker` first (the same pattern already used a few lines below in this file for an analogous reentrancy concern) instead of `flush_thread()` directly. This forces a genuine cross-thread hop — the message is queued and the stack unwinds before the retry runs, no matter how many times `try_lock` fails.

## Regression test

`LogStoreTest.FlushIfNecessaryRetrySurvivesStaleClockReset` deterministically reproduces both follow-up issues using two small `_PRERELEASE`-only `LogDev` test hooks (`test_acquire_flush_mtx()`, `test_touch_last_flush_time()`) instead of a real racer `flush()` call, which can't cleanly isolate the bug (releasing it would always flush the target write anyway, masking the result regardless of the fix). Verified against all four combinations of the two fixes, same test unchanged:

| `force` | `random_worker` | Result |
|---|---|---|
| ❌ | ❌ (original) | crash (SIGSEGV), 5/5 |
| ❌ | ✅ | fails (assertion), 9/10 |
| ✅ | ❌ | crash (SIGSEGV), 5/5 |
| ✅ | ✅ (this PR) | passes, 20/20 |

## Why this doesn't reintroduce anything during shutdown

If the rescheduled attempt runs after `is_stopping()` has become true, it just returns immediately — same as today, no new reschedule. That's fine because shutdown already has its own independent, *guaranteed* flush: `LogDev::stop()` calls `flush_under_guard()`, which takes `flush_guard()` with a real **blocking** `unique_lock` (not `try_to_lock`) and then calls `flush()` unconditionally. That flush covers everything up to `m_log_idx - 1`, and `m_log_idx` is already incremented synchronously at append time — well before any later, asynchronous `flush_if_necessary()` attempt — so any data a dropped reschedule would have covered is still guaranteed to be flushed by `stop()`'s own path before teardown proceeds. The reschedule and the shutdown-guaranteed flush are two independent safety nets for two different situations (normal operation vs. shutdown), not two paths racing over the same data.

### DIFF
--- a/src/lib/logstore/log_dev.cpp
+++ b/src/lib/logstore/log_dev.cpp
@@ -462,6 +462,17 @@ bool LogDev::flush_if_necessary(int64_t threshold_size) {
             decr_pending_request_num();
             return flush();
         }
+        // Lost the race to a concurrent flush() (e.g. another write's own flush_if_necessary() call, or
+        // HomeLogStore::truncate()'s internal flush()). That concurrent flush's own snapshot of m_log_idx
+        // may not include the data that made this call decide to flush, so giving up here silently can
+        // leave that data unflushed indefinitely if nothing else ever calls flush_if_necessary() again
+        // for this logdev -- normally masked by the periodic flush timer eventually retrying, but with
+        // it disabled (flush_timer_frequency_us=0, e.g. in tests) this is a real, reproducible hang: the
+        // very last write issued in a run has no later trigger to fall back on. Reschedule a follow-up
+        // attempt on the flush thread instead of dropping it, mirroring the !can_flush_in_this_thread()
+        // reschedule above.
+        iomanager.run_on_forget(logstore_service().flush_thread(),
+                                [this, threshold_size]() { flush_if_necessary(threshold_size); });
     }
     decr_pending_request_num();
     return false;
--- a/src/lib/logstore/log_dev.cpp
+++ b/src/lib/logstore/log_dev.cpp
@@ -482,7 +482,7 @@ bool LogDev::flush_if_necessary(int64_t threshold_size, bool force) {
         // here, and IOReactor::deliver_msg runs same-reactor targets inline instead of queuing them --
         // so posting straight back to flush_thread would recurse synchronously on every failed try_lock.
         // Under sustained contention this stack-overflows the process (confirmed with the lock held for
-        // just ~200ms; this predates force -- it's already present in the plain reschedule above).
+        // ~200ms in testing; this predates force -- it's already present in the plain reschedule above).
         // Routing through random_worker first forces a real queued hop, so the stack unwinds between
         // attempts no matter how many times try_lock fails.
         iomanager.run_on_forget(iomgr::reactor_regex::random_worker,
--- a/src/lib/logstore/log_dev.cpp
+++ b/src/lib/logstore/log_dev.cpp
@@ -513,6 +513,12 @@ bool LogDev::flush_under_guard() {
     return flush();
 }
 
+#ifdef _PRERELEASE
+void LogDev::test_touch_last_flush_time() {
+    if (iomgr_flip::instance()->test_flip("test_touch_last_flush_time")) { m_last_flush_time = Clock::now(); }
+}
+#endif
+
 bool LogDev::flush() {
     if (!is_ready()) {
         THIS_LOGDEV_LOG(INFO, "LogDev is not ready to flush, log_dev={}", m_logdev_id);
--- a/src/lib/logstore/log_dev.cpp
+++ b/src/lib/logstore/log_dev.cpp
@@ -470,28 +470,21 @@ bool LogDev::flush_if_necessary(int64_t threshold_size, bool force) {
         // it disabled (flush_timer_frequency_us=0, e.g. in tests) this is a real, reproducible hang: the
         // very last write issued in a run has no later trigger to fall back on.
         //
-        // Reschedule a follow-up attempt -- with force=true, unlike the initial call. LogDev::flush()
-        // unconditionally resets m_last_flush_time at its very start, so the concurrent flush that just
-        // stole this race will, by the time the retry runs, likely have already reset that clock to "now"
-        // (making flush_by_time re-evaluate false) while this call's own pending size may still be under
-        // threshold_size (making flush_by_size false too) -- silently abandoning the retry with nothing
-        // left to trigger it again. force=true bypasses that re-derivation on the retry: we already
-        // decided this data needs flushing, so the retry's only job is to keep trying to actually acquire
-        // the lock, not re-litigate whether to.
+        // Reschedule with force=true. LogDev::flush() unconditionally resets m_last_flush_time at its
+        // start, so the flush that just won this race may have already reset that clock by the time the
+        // retry runs -- re-deriving flush_by_size/flush_by_time here could then read false again (this
+        // write's own size may still be under threshold_size) and silently abandon the retry for good.
+        // force=true skips that re-derivation: we already decided to flush, so the retry only needs to
+        // keep trying the lock, not re-litigate whether to.
         //
-        // Reschedule onto a random *worker* reactor, not flush_thread() directly, even though flush_thread
-        // is where the retry ultimately needs to run (can_flush_in_this_thread() will bounce it back
-        // there). We're already executing on flush_thread at this point (that's how we got past the
-        // can_flush_in_this_thread() check above) -- IOReactor::deliver_msg takes a same-thread shortcut
-        // that calls the target inline instead of queuing it when sender and receiver are the same
-        // reactor, so posting straight back to flush_thread here would recurse synchronously on this same
-        // call stack for every failed try_lock, not queue a new task. Under sustained lock contention
-        // (verified: a lock held for as little as ~200ms is enough) that recursion runs thousands of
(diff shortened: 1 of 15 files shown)


JacksonYao287:
thanks @shosseinimotlagh for this finding. 

well,  I noticed that in UT we set flush_timer_frequency_us to 0, but I am not quite clear why.
I think logdev timer should always be enabled(This is a fallback measure), so that  the case here will not happen?

shosseinimotlagh:
> thanks @shosseinimotlagh for this finding.
> 
> well, I noticed that in UT we set flush_timer_frequency_us to 0, but I am not quite clear why. I think logdev timer should always be enabled(This is a fallback measure), so that the case here will not happen?

@JacksonYao287   test_log_store.cpp:172 — the test itself calls m_log_store->get_logdev()->flush_if_necessary(1) directly after every insert. That's not incidental; it's the mechanism. Since these logdevs have neither INLINE nor TIMER bits, the only thing driving flush attempts is this per-write explicit call — so many concurrent writes each fire their own flush_if_necessary() concurrently, and whichever one loses the try_lock race on the batch's last write has nothing left to retry it. That's exactly the reproducer, and it's not a test artifact — it's a faithful simulation of a legitimate production pattern: a caller who deliberately chose flush_mode_t::EXPLICIT to self-drive flushing without paying for a background timer.

That's the key point for the fix's justification: "the timer should always be on" doesn't actually close this gap in general, because EXPLICIT-only logdevs (a deliberate, supported mode for callers who want to control flush timing themselves) never get a timer safety net in the first place — timer-based mitigation is orthogonal to this bug.

src/lib/logstore/log_dev.cpp
JacksonYao287, line 456: if we schedule the flush to the flush_thread, when it is executed, flush_by_size and flush_by_time are both false, we will still lose this flush right?

so for the case mentioned in this PR, do we need a way to guarantee the flush will be done?

JacksonYao287 (commented):
thanks , mehdi, the explanation makes sense to me. I have another concern, pls take a look

src/lib/logstore/log_dev.cpp
shosseinimotlagh: You're right, and this needed a real fix.
 The reschedule as originally written re-invokes flush_if_necessary(), which re-derives flush_by_size/flush_by_time from scratch. flush() resets m_last_flush_time unconditionally at its very start, so the concurrent flush that won the race can reset that clock right before the retry runs — if this write's own size never crosses threshold_size on its own, both conditions can read false and the retry silently drops, exactly the failure this PR set out to fix.

 Fix: flush_if_necessary() now takes a force parameter. The reschedule passes force=true, which skips the size/time re-derivation entirely and goes straight to try_lock. We already decided this data needs flushing when we entered this branch — the retry's only job is to keep trying to acquire the lock, not re-litigate whether to. Still a non-blocking try_to_lock, so no new deadlock surface.

 Building a deterministic repro to verify this (rather than relying on timing/luck) surfaced a second, separate, more severe issue in the original reschedule — present with or without force, so it's independent of that fix and in fact predates this PR entirely (it's already in the original reschedule-on-lost-race commit): IOReactor::deliver_msg takes a same-thread shortcut, calling the target inline instead of queuing it whenever sender and receiver are the same reactor. The reschedule always targets flush_thread(), and we're always already on flush_thread() by the time we reach it, so every failed retry was a real recursive call on the same stack, not a queued task. Under sustained lock contention (confirmed with the lock held for ~200ms in testing) that recursion runs tens of thousands of frames deep and stack-overflows the process. Fixed by routing the retry through iomgr::reactor_regex::random_worker first (the same pattern already used a few lines below in this file), forcing a genuine cross-thread hop so the stack always unwinds between attempts.

 Added LogStoreTest.FlushIfNecessaryRetrySurvivesStaleClockReset to cover both, deterministically (two small _PRERELEASE-only LogDev hooks instead of a real racer flush(), which can't cleanly isolate this — releasing it would flush the target write regardless of the fix). Verified all four combinations, same test unchanged: neither fix → crash (5/5), force alone → crash (5/5), random_worker alone → fails (9/10), both → passes (20/20). Pushed.

src/lib/logstore/log_dev.cpp
```diff
@@ -437,12 +437,12 @@ bool LogDev::can_flush_in_this_thread() {
     return (!HS_DYNAMIC_CONFIG(logstore.flush_only_in_dedicated_thread) && iomanager.am_i_worker_reactor());
 }
 
-bool LogDev::flush_if_necessary(int64_t threshold_size) {
+bool LogDev::flush_if_necessary(int64_t threshold_size, bool force) {
     if (is_stopping()) return false;
```
JacksonYao287, line 441: generally ，LGTM。a small concern is that if is_stopping here returns false, then even if force is true, we will still lose that flush?

shosseinimotlagh: @JacksonYao287 force fixes the steady-state hang (no periodic timer to retry); shutdown durability is a separate, already-correct guarantee (append_async's own is_stopping() gate + stop()'s blocking flush_under_guard()), and the two don't need to cooperate for correctness — force racing against is_stopping() was never load-bearing.


You're right that is_stopping() short-circuits before force is even checked — so a retry that lands after is_stopping() flips is dropped regardless of force. That's not new to this change though (every method here gates on is_stopping() first), and it's harmless in practice: LogDev::stop() has its own independent safety net for exactly this window — after draining pending_request_num it calls flush_under_guard(), which takes a plain blocking lock (not try_lock), so it can't lose a race. Any write that reached flush_if_necessary() already holds a pending_request_num slot until that call returns, so its data is already in the pending buffer by the time stop()'s drain completes — flush_under_guard()'s own flush picks it up regardless of what happened to that write's retry chain. force is specifically about the steady-state hang (no periodic timer to retry later); shutdown durability comes from stop()'s blocking flush, which doesn't depend on this retry at all.

JacksonYao287 (approved):
LGTM

PR #917, opened 2026-09-16
SDSTOR-24993, SDSTOR-25623: propagate journal flush I/O errors and fail-stop on unrecoverable flush failure

shosseinimotlagh:
## Summary
- `log_write_comp_cb_t`/`log_req_comp_cb_t` now carry a `std::error_condition status`. `LogDev::on_flush_completion()` always fires the completion callback -- success or failure -- instead of returning early on a `sync_pwritev` failure and never calling back at all.
- `HomeRaftLogStore::append`/`write_at` and `SoloReplDev::write_journal` now check that status: on a flush I/O error they escalate via `fault_cmt_service` if the embedder registered one, otherwise fail-stop (`HS_REL_ASSERT`), instead of silently treating a non-durable write as committed.

## Test plan
- New regression tests: `FlushIOErrorPropagatesToCallback`, `FlushIOErrorPropagatesToEveryRequestInGroup` in `test_log_store.cpp`, using a `_PRERELEASE` flip to force a flush failure.
- Existing `test_log_store`/`test_log_store_long_run` suites updated mechanically for the new callback signature and pass.


src/lib/replication/log_store/home_raft_log_store.cpp
JacksonYao287, line 220: here， we ignore the return value of flush. this means even if the log flush failed, we will still call this and update m_last_durable_lsn?

https://github.com/eBay/NuRaft/blob/0b01b185b46b8e482f6c384b636c35605121a41e/src/handle_append_entries.cxx#L1027

this function is critical for nuraft, and thus for raft_repl_dev. so we need think through if log flush failed, how to handle these logs for raft

src/lib/logstore/log_dev.cpp
JacksonYao287, line 526: here it returns false, but the log is actually flushed. should we consider this as failed?

src/lib/logstore/log_dev.cpp
```diff
@@ -561,11 +562,16 @@ bool LogDev::flush() {
         HISTOGRAM_OBSERVE(logstore_service().m_metrics, logdev_flush_records_distribution, lg->nrecords());
         HISTOGRAM_OBSERVE(logstore_service().m_metrics, logdev_flush_size_distribution, lg->actual_data_size());
 
-        // TODO:: add logic to handle this error in upper layer
         auto error = m_vdev_jd->sync_pwritev(lg->iovecs().data(), int_cast(lg->iovecs().size()), lg->m_log_dev_offset);
+#ifdef _PRERELEASE
+        if (iomgr_flip::instance()->test_flip("simulate_log_flush_error")) {
+            error = std::make_error_code(std::errc::io_error);
```
JacksonYao287, line 568: NIT：since we have return the error，should we return the specific error from sync_pwritev, not always return an io_error

src/lib/logstore/log_store.cpp
JacksonYao287, line 111: do we need care about the returned value of flush_under_guard  to see if it is successful?

JacksonYao287 (commented):
thanks @shosseinimotlagh , I thinks this is a little big change for the behavior of log store. let`s wait for more inputs from other folks.

We need to carefully evaluate the semantic and behavioral impact that the change may introduce, especially checking whether it would alter any of NuRaft’s behavioral expectations for log_stor

